PR #477, opened 2021-09-17
Add init-db stage to pre-populate database

connorgorman:
- Requires change to Central helm chart, separates concerns
- Will follow up and make pg_hba.conf a configmap realistically


RTann:
https://github.com/stackrox/scanner/pull/774

PR #604, opened 2022-01-26
ROX-9497: Make use of the injected CA bundle

0x656b694d:
Update the `import-additional-cas` to copy the injected CA bundle to the common location before calling `update-ca-trust`.

> Cluster administrators can configure additional trusted CAs in OpenShift. The combined set of trusted CAs (from a cluster perspective) can be obtained by creating an empty ConfigMap with a special label attached to it. An operator will then inject the trusted CAs into a well-defined field of the configmap.

Related PR to stackrox: https://github.com/stackrox/stackrox/pull/174

### DIFF
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 copy_existing () {
-    src=$1
+    src="$1"
     if [ -d "$src" ] && [ "$(ls -A -I "..*" "$src")" ]; then
         cp -v -L "$src"/* /etc/pki/ca-trust/source/anchors
     else
@@ -14,4 +14,4 @@ copy_existing () {
 copy_existing /usr/local/share/ca-certificates
 copy_existing /etc/pki/injected-ca-trust
 
-update-ca-trust
+update-ca-trust extract
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -5,8 +5,8 @@ set -euo pipefail
 lines=$(for d in /usr/local/share/ca-certificates /etc/pki/injected-ca-trust
 do
     if [ -d "$d" ]; then
-        find "$d" -maxdepth 1 -type f \
-         -print -exec cp -L {} /etc/pki/ca-trust/source/anchors \;
+        find "$d" -maxdepth 1 -type f -print0 | \
+            xargs --no-run-if-empty -0 cp -t /etc/pki/ca-trust/source/anchors
     fi
 done | wc -l)
 
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -2,7 +2,10 @@
 
 set -euo pipefail
 
-[ -d /usr/local/share/ca-certificates ] || exit 0
-[ "$(find /usr/local/share/ca-certificates -maxdepth 1 -name '*.crt' | wc -l)" -gt 0 ] || exit 0
-cp -L /usr/local/share/ca-certificates/* /etc/pki/ca-trust/source/anchors
-update-ca-trust
+lines=$(for d in /usr/local/share/ca-certificates /etc/pki/injected-ca-trust
+do
+    [ -d "$d" ] && find "$d" -maxdepth 1 -name '*.crt' -o -name '*.pem' \
+         -print -exec cp -L {} /etc/pki/ca-trust/source/anchors \;
+done | wc -l)
+
+[ "$lines" -gt 0 ] && update-ca-trust
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -2,15 +2,5 @@
 
 set -euo pipefail
 
-lines=$(for d in /usr/local/share/ca-certificates /etc/pki/injected-ca-trust
-do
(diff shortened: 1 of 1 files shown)


image/scanner/scripts/import-additional-cas
```diff
@@ -1,8 +1,17 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -euo pipefail
 
-[ -d /usr/local/share/ca-certificates ] || exit 0
-[ "$(find /usr/local/share/ca-certificates -maxdepth 1 -name '*.crt' | wc -l)" -gt 0 ] || exit 0
-cp -L /usr/local/share/ca-certificates/* /etc/pki/ca-trust/source/anchors
+copy_existing () {
+    src=$1
```
msugakov, line 6: [nit]
```suggestion
    src="$1"
```

image/scanner/scripts/import-additional-cas
```diff
@@ -1,8 +1,17 @@
...
+    if [ -d "$src" ] && [ "$(ls -A -I "..*" "$src")" ]; then
+        cp -v -L "$src"/* /etc/pki/ca-trust/source/anchors
+    else
+        echo "No certificates found in $src"
```
msugakov, line 10: Maybe just skip it quietly?
```suggestion
```

image/scanner/scripts/import-additional-cas
```diff
@@ -1,8 +1,17 @@
...
+    fi
+}
+
+copy_existing /usr/local/share/ca-certificates
+copy_existing /etc/pki/injected-ca-trust
+
 update-ca-trust
```
msugakov, line 17:
```suggestion
update-ca-trust extract
```

image/scanner/scripts/import-additional-cas
```diff
@@ -1,8 +1,17 @@
...
 
 set -euo pipefail
 
-[ -d /usr/local/share/ca-certificates ] || exit 0
-[ "$(find /usr/local/share/ca-certificates -maxdepth 1 -name '*.crt' | wc -l)" -gt 0 ] || exit 0
-cp -L /usr/local/share/ca-certificates/* /etc/pki/ca-trust/source/anchors
+copy_existing () {
+    src=$1
+    if [ -d "$src" ] && [ "$(ls -A -I "..*" "$src")" ]; then
+        cp -v -L "$src"/* /etc/pki/ca-trust/source/anchors
```
msugakov, line 8: On one hand I like `-v`, on the other hand, this code is called each time the image is started and will produce some logging that may be noisy. WDYT?

image/scanner/scripts/import-additional-cas
```diff
@@ -1,8 +1,17 @@
...
+#!/usr/bin/env bash
 
 set -euo pipefail
 
-[ -d /usr/local/share/ca-certificates ] || exit 0
-[ "$(find /usr/local/share/ca-certificates -maxdepth 1 -name '*.crt' | wc -l)" -gt 0 ] || exit 0
-cp -L /usr/local/share/ca-certificates/* /etc/pki/ca-trust/source/anchors
+copy_existing () {
+    src=$1
+    if [ -d "$src" ] && [ "$(ls -A -I "..*" "$src")" ]; then
```
msugakov, line 7: 1. Why is `-I "..*"` needed?
2. You can merge predicates
```suggestion
if [[ -d "$src" && -n "$(ls -A -I "..*" "$src")" ]]; then
```

image/scanner/scripts/import-additional-cas
msugakov: Feel free to ignore for consistency with https://github.com/stackrox/stackrox/pull/1088

image/scanner/scripts/import-additional-cas
msugakov: Feel free to ignore for consistency with https://github.com/stackrox/stackrox/pull/1088

0x656b694d: That `-v` was added specifically for generating the little noise for logging purposes to simplify debugging.

image/scanner/scripts/import-additional-cas
0x656b694d: 1. `-I "..*"` is there to ignore files like "..data" created by kubernetes volume mounting.
2. Yes, but why?

msugakov: 2: Clarity. Each `[` spawns a command. In `[ something ] && [ something ]` there are at least two `[` commands.
Unlike `[`, `[[` is, I believe, a bash built-in that doesn't span extra command on its own. Plus `&&` inside of `[` or `[[` doesn't cause another command execution.
By _command_ I mean starting a process.

0x656b694d: Hm. I'm not sure how `[[a && -n b]]` is more clear than `[a] && [b]`.

msugakov: You can try `[[ a && b ]]`. Most likely it will work the same. Will it still be not clear enough?

msugakov: Regarding
> `-I "..*"` is there to ignore files like "..data" created by kubernetes volume mounting.

It is really not obvious from just looking at the code. I would recommend you adding this info as a comment in the code.

0x656b694d:
Merged with #692

PR #683, opened 2022-04-01
Fix cron time for nightly

c-du:


.circleci/config.yml
```diff
@@ -1191,7 +1191,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "* 18 * * *"
+          cron: "18 11 * * *"
```
RTann, line 1194: Should these be flipped for "nightly"? https://www.unix.com/man-page/POSIX/1posix/crontab/

RTann (approved):
Approving because it's the right idea, but I think we should flip the two numbers

.circleci/config.yml
c-du: This is early morning in UTC. So I want result ready every morning.

RTann: I forgot it's UTC :)

PR #687, opened 2022-04-07
Use stackrox/stackrox instead of stackrox/rox

connorgorman:


jvdm (approved):

RTann (approved):
Keeping the version the same was a good idea. I ran into some fun when I bumped it

PR #688, opened 2022-04-08
ROX-10021: Ignore RHCOS nodes

RTann:
We now explicitly ignore RHCOS nodes to prevent an influx of false-positives. We now, also, exit early upon detecting explicitly unsupported node OSes. This means that we will no longer look for kubelet, kube-proxy, nor container runtime vulns in Google COS and Windows nodes as we did before.

This is what it looks like from the UI:

Screen Shot 2022-04-08 at 10 47 12 AM


ext/kernelparser/rhel/rhel.go
```diff
@@ -15,7 +15,11 @@ const (
 )
 
 var (
-	regex = regexp.MustCompile(`el[0-9]+`)
+	// rhcosOSImagePattern is the pattern by which versions of Red Hat CoreOS follow.
+	// Example: Red Hat Enterprise Linux CoreOS 47.84.202203141332-0 (Ootpa)
+	// The parser expects osImage to be lowercase, so the pattern is lowercase.
+	rhcosOSImagePattern = regexp.MustCompile(`^red hat enterprise linux coreos `)
```
RTann, line 21: I'm tempted to just call these `rhcosOSImage` and `rhelKernel`, respectively

ext/kernelparser/register.go
```diff
@@ -20,7 +20,12 @@ type ParseMatch struct {
 }
 
 // Parser is a kernel parser.
-type Parser func(db database.Datastore, kernelVersion, osImage string) (*ParseMatch, bool, error)
+// osImage is expected to be lowercase (for example by calling `strings.ToLower(osImage)`).
+// Returns a non-nil *ParseMatch, match, if the kernel is supported.
+// Returns ok = true if the kernel was recognized by the parser.
+// Returns a non-nil error when there is an error parsing the kernel or if the node represented by the
+// osImage and kernel is recognized and completely unsupported.
+type Parser func(db database.Datastore, kernelVersion, osImage string) (match *ParseMatch, ok bool, err error)
```
jvdm, line 28: This comment points out that `Parse` return value is getting difficult. Can we simplify the different combinations of return values and encode the different match scenarios in the match object itself? Then we can use `(match *ParseMatch, err error)` to mean:

- `match == nil && err == nil` - The parser did not recognize this os/version, but it might be recognized by another parser (same as current `ok = false`).
- `match == nil && err != nil` - The parser recognized the os/version string, but it is malformed. `Parse()` could not extract meaningful information, hence `match` is `nil`., but no other parser will recognize this os/version.
- `match != nil && err == nil` - The parser recognized and successfully parsed the os/version. 

Then `match` can have `match.isSupported` to signal the os/version is not supported. Or maybe others, so we don't have to use `ok`.

api/v1/nodescan/service.go
```diff
@@ -227,6 +227,12 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
 
 	resp.OperatingSystem, resp.KernelVulnerabilities, resp.KernelComponent, err = s.evaluateLinuxKernelVulns(req)
 	if err != nil {
+		// If the node is unsupported, exit early.
+		if err == kernelparser.ErrNodeUnsupported {
+			resp.Notes = append(resp.Notes, v1.NodeNote_NODE_OS_UNSUPPORTED)
+			return resp, nil
+		}
```
jvdm, line 234: Why a "note" and not a plain empty response, or an error? Does the UI consume empty responses with notes in a meaningful way?

I noticed that we return empty responses in some scenarios (like "windows"). I would instead use the same pattern unless there is a reasonable tradeoff because adding a specialized error increases complexity, and I want to avoid it.

RTann: An error implies something went wrong when scanning. Putting a note here tells us that scanning went exactly as expected, but the results are incomplete do to the reasons shown in the note. We have the same idea for images: https://github.com/stackrox/scanner/blob/master/proto/scanner/api/v1/note.proto#L9. The UI displays the scanning notes in the Vuln Mgmt pages. See https://github.com/stackrox/stackrox/pull/1275 for the related PR.

Good idea adding a note for something like "windows". Will add that

ext/kernelparser/register.go
RTann: Yeah this is getting complicated. I'm going to make specialized errors we will handle in a `switch`

ext/kernelparser/rhel/rhel.go
RTann: good idea. no need for a regex when it's just an exact prefix

ext/kernelparser/ubuntu/ubuntu.go
```diff
@@ -30,10 +31,10 @@ func parser(db database.Datastore, kernelVersion, osImage string) (*kernelparser
 
 	matches := regex.FindStringSubmatch(osImage)
 	if len(matches) == 0 {
-		return nil, false, fmt.Errorf("could not find Ubuntu version in OS string: %q", osImage)
+		return nil, false, errors.Errorf("could not find Ubuntu version in OS string: %q", osImage)
```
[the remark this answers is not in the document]
RTann: I don't think anything under `ext/kernelparser` should be affected by what the gRPC server (caller) does with the error. That being said, I see your point, and it was already `fmt`, so I'll just switch it back

proto/scanner/api/v1/note.proto
```diff
@@ -12,3 +12,8 @@ enum Note {
   LANGUAGE_CVES_UNAVAILABLE       = 2;
   CERTIFIED_RHEL_SCAN_UNAVAILABLE = 3;
 }
+
+enum NodeNote {
+  NODE_OS_UNSUPPORTED     = 0;
```
RTann, line 17: OS unsupported and Kernel unsupported may be confusing. Going to rename `NODE_OS_UNSUPPORTED` to just `NODE_UNSUPPORTED`

ext/kernelparser/register.go
jvdm: OK, will follow up there.

api/v1/nodescan/service.go
jvdm: It makes sense, considering the UI handles notes in the way you clarified.

About incomplete results and how best we can report them (as errors or not), that's debatable. It is better explained as a convention rather than a principle, and I am unsure if you are pointing out a convention. But, considering the notes are handled the way they are, I am OK with your approach anyway.

ext/kernelparser/debian/debian.go
```diff
@@ -24,32 +24,32 @@ func init() {
...
 
 	if !strings.Contains(osImage, "debian") {
-		return nil, false, nil
+		return nil, kernelparser.ErrKernelUnrecognized
 	}
 
 	matches := regex.FindStringSubmatch(osImage)
 	if len(matches) == 0 {
-		return nil, false, fmt.Errorf("could not find Debian version in OS string: %q", osImage)
+		return nil, fmt.Errorf("could not find Debian version in OS string: %q", osImage)
```
jvdm, line 43: It means we will panic, right? Let's not panic when we cannot parse the OS, better handle it gracefully. Can we signal `ErrKernelUnrecognized`, for example? Or another status code?

ext/kernelparser/rhel/rhel.go
```diff
@@ -15,30 +15,39 @@ const (
...
-	matches := regex.FindStringSubmatch(kernelVersion)
+	if strings.HasPrefix(osImage, rhcosOSImagePrefix) {
+		// Explicitly ignore RHEL CoreOS nodes completely.
+		return nil, kernelparser.ErrNodeUnsupported
+	}
+
+	matches := rhelKernelPattern.FindStringSubmatch(kernelVersion)
 	if len(matches) == 0 {
-		return nil, false, fmt.Errorf("could not find RHEL version in kernel version string: %q", osImage)
+		return nil, fmt.Errorf("could not find RHEL version in kernel version string: %q", osImage)
```
jvdm, line 41: See comment on debian.go, it applies here.

ext/kernelparser/ubuntu/ubuntu.go
```diff
@@ -23,17 +23,17 @@ func init() {
...
+func parser(db database.Datastore, kernelVersion, osImage string) (*kernelparser.ParseMatch, error) {
 	if !strings.Contains(osImage, "ubuntu") {
-		return nil, false, nil
+		return nil, kernelparser.ErrKernelUnrecognized
 	}
 
 	matches := regex.FindStringSubmatch(osImage)
 	if len(matches) == 0 {
-		return nil, false, fmt.Errorf("could not find Ubuntu version in OS string: %q", osImage)
+		return nil, fmt.Errorf("could not find Ubuntu version in OS string: %q", osImage)
```
jvdm, line 33: See comment in debian.go, it applies here.

api/v1/nodescan/service.go
```diff
@@ -95,32 +96,38 @@ func featureVersionToKernelComponent(fv database.FeatureVersion) *v1.GetNodeVuln
...
+			// This parser did not recognize the kernel. Try another one.
+			continue Parsers
+		case kernelparser.ErrKernelUnsupported:
+			log.Debugf("%s parser found unsupported kernel: %s %s", name, req.GetOsImage(), kernelVersion)
+		case kernelparser.ErrNodeUnsupported:
+			log.Debugf("%s parser found unsupported node: %s %s", name, req.GetOsImage(), kernelVersion)
+		default:
+			// Should never happen.
+			utils.Must(errors.Errorf("Found unexpected error in kernel parser: %v", err))
 		}
```
jvdm, line 120: Wouldn't it make more sense to use a Status Enum rather than error codes? It would give us a compile-time assurance instead of a runtime failure. It would also make the function interface less ambiguous, and prevent inadvertent panicking.

proto/scanner/api/v1/note.proto
```diff
@@ -12,3 +12,8 @@ enum Note {
   LANGUAGE_CVES_UNAVAILABLE       = 2;
   CERTIFIED_RHEL_SCAN_UNAVAILABLE = 3;
 }
+
+enum NodeNote {
+  NODE_UNSUPPORTED        = 0;
+  NODE_KERNEL_UNSUPPORTED = 1;
+}
```
jvdm, line 19: Nice.

ext/kernelparser/debian/debian.go
RTann: great catch i completely missed this!

RTann: I removed the panic, and I added tests to the service to ensure we support these "non-standard" errors

jvdm (approved):

PR #691, opened 2022-04-14
deps: Correct replace path and sync bleve versions

crozzy:
The replace path for github.com/stackrox/rox was incorrect
and the version of bleve being replaced not being the same
as stackrox/stackrox was causing an import error.

Signed-off-by: crozzy <joseph.crosland@gmail.com>


connorgorman (approved):
thx

RTann:
Replacing with https://github.com/stackrox/scanner/pull/694

PR #693, opened 2022-04-19
Use pure go xz implementation

RTann:
https://github.com/ulikunitz/xz


connorgorman (commented):
how did you test this?

pkg/tarutil/tarutil.go
```diff
@@ -277,7 +237,7 @@ func (r *TarReadCloser) Close() error {
 // Gzip/Bzip2/XZ detection is done by using the magic numbers:
 // Gzip: the first two bytes should be 0x1f and 0x8b. Defined in the RFC1952.
 // Bzip2: the first three bytes should be 0x42, 0x5a and 0x68. No RFC.
-// XZ: the first three bytes should be 0xfd, 0x37, 0x7a, 0x58, 0x5a, 0x00. No RFC.
+// XZ: the first six bytes should be 0xfd, 0x37, 0x7a, 0x58, 0x5a, 0x00. No RFC.
```
janisz, line 240: Why this has changed?

janisz:
Indeed it's pure go without any deps https://github.com/ulikunitz/xz/blob/master/go.mod

msugakov:
What worries me is that the author [acknowledges](https://github.com/ulikunitz/xz#package-xz) instability of the library

> The package is currently under development. There might be bugs and APIs are not considered stable. At this time the package cannot compete with the xz tool regarding compression speed and size. The algorithms there have been developed over a long time and are highly optimized.

Not sure if trading one os-level dependency for potential bugs is worth it.

pkg/tarutil/tarutil.go
RTann: I think "three" was a typo. It is six bytes

RTann:
> What worries me is that the author [acknowledges](https://github.com/ulikunitz/xz#package-xz) instability of the library

I didn't notice that originally. It mostly acknowledges the compression (write) portion, and we only care about the read ability. That being said, read may have issues, too. When testing I didn't notice, but there is always a chance there is something wrong.

> Not sure if trading one os-level dependency for potential bugs is worth it, at this point.

I was a bit curious, too, especially since there is no native golang support for this. I saw over 9k users of this with some notable names in there, but that being said, that could also mean at least 9k users are vulnerable to potential bugs/vulnerabilities.

I will close this, as the native `xz` in the OS has been fine so far. Or at least, nobody has complained about it, so why fix what ain't broken

PR #708, opened 2022-04-26
Protect from potential "zip slip"

RTann:
We control what goes into these files, and we know they do not have ".." in the name, but let's protect ourselves from it just in case.

### DIFF
--- a/cmd/updater/diffdumps/cmd.go
+++ b/cmd/updater/diffdumps/cmd.go
@@ -78,6 +78,11 @@ func generateK8sDiffs(outputDir string, baseZipR *zip.ReadCloser, headZipR *zip.
 	baseFiles := make(map[string]*zip.File)
 	for _, baseF := range baseZipR.File {
 		name := baseF.Name
+
+		if strings.Contains(name, "..") {
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


cmd/updater/diffdumps/cmd.go
```diff
@@ -78,13 +78,24 @@ func generateK8sDiffs(outputDir string, baseZipR *zip.ReadCloser, headZipR *zip.
 	baseFiles := make(map[string]*zip.File)
 	for _, baseF := range baseZipR.File {
 		name := baseF.Name
+
+		if strings.Contains(name, "..") {
```
viswajithiii, line 82: Shouldn't we only check for a specific component of the path being `..`? Like, a file name like "README..txt" is valid and does not cause zip slip.

RTann: Good idea. I copied what codeql said to do, but https://snyk.io/research/zip-slip-vulnerability has a better check

RTann: actually, going to just switch it to `../`. Idea being `../` covers all cases except when the path ends with `..`, but we check the extension so we already implicitly disallow `..` at the end

viswajithiii (approved):

PR #712, opened 2022-04-27
ROX-9497: Copy custom trusted CA bundles injected by OpenShift Network Operator

0x656b694d:
## Description

Update the system trusted CA bundle with the ones injected by the Openshift Network Operator and mounted as a kubernetes secret volume.

The script is copied from the stackrox repo and enriched with some comments and the `extract` parameter for verbosity. Once merged, the script will be copied back to stackrox for consistency.

This PR is a resurrection of #604.

## Testing Done

Tests performed previously when merging to the stackrox repository.

### DIFF
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -19,4 +19,4 @@ copy_existing /usr/local/share/ca-certificates
 # Copy the custom trusted CA bundles injected by the Openshift Network Operator.
 copy_existing /etc/pki/injected-ca-trust
 
-update-ca-trust
+update-ca-trust extract
--- a/image/scanner/scripts/import-additional-cas
+++ b/image/scanner/scripts/import-additional-cas
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+# The function copies everything that could be found under the path provided in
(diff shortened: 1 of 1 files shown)


image/scanner/scripts/import-additional-cas
```diff
@@ -13,4 +16,7 @@ copy_existing () {
 
 copy_existing /usr/local/share/ca-certificates
 
-update-ca-trust
+# Copy the custom trusted CA bundles injected by the Openshift Network Operator.
+copy_existing /etc/pki/injected-ca-trust
+
+update-ca-trust extract
```
RTann, line 22: what does `extract` do? I don't see it used here https://github.com/stackrox/stackrox/blob/master/image/static-bin/import-additional-cas

RTann: Ah ok https://www.unix.com/man-page/centos/8/update-ca-trust/ tells me it's the same but with more warnings. What encouraged the switch?

RTann: I see you answered all of this in the description I clearly didn't read

RTann (approved):

msugakov (approved):
Thanks for following through.

PR #714, opened 2022-04-29
ROX-9614: Add slim updater

jvdm:
## Description

Add an updater for Scanner Slim, called "Slim Updater", which pipelines request to retrieve vulnerability definitions from Sensor rather than Central. Also, refactor parts of the `updater` package to properly decouple the code between the two updaters by factoring out shared code between them two.

The slim updater is necessary to retrieve updates to the `repoToCPE` mapping, which comes from retrieving `rhvel2/repository-to-cpe.json` from the Scanner's definition bundle. See https://github.com/stackrox/stackrox/pull/1420 for details on how Sensor supports this operation.

## Testing

This was tested manually.

1. Locally in a `minikube` single cluster setup, where `-slim` was forced to be deployed and the API pipeline verified.
   - `curl` command for testing Central's API: `curl -v -sk -G -u "admin:$(oc -n stackrox get secrets central-htpasswd -o json | jq .data.password -r | base64 -d)" "https://localhost:8000/api/extensions/scannerdefinitions?file=rhelv2%2Frepository-to-cpe.json&uuid=90ed7d2d-7839-4f65-93fb-e56f37a62224"`
2. At Openshift, 2 clusters setup, single namespace.
   - https://infra.rox.systems/cluster/jvdm-ekloom-secured
   - https://infra.rox.systems/cluster/jvdm-ekloom-central
   - Tests performed:
     a. Verify the slim-updater loop was executed successfully by inspecting logs in the secured cluster;
     b. Inspect Scanner images for `/var/lib/stackrox/slim-updater-artifacts.d/rhelv2/repository-to-cpe.json`.


pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,140 @@
...
+		}, "/"),
+		url.Values{
+			"uuid": []string{uuid},
+			"file": []string{repoToCPEFilename},
+		})
+	utils.CrashOnError(err)
+
+	// Create sensor's HTTP client.
+	//
+	// FIXME Adopt clientconn.NewHTTPClient() when available in the repository.
```
jvdm, line 64: Will be provided by https://github.com/stackrox/stackrox/pull/1420

go.sum
```diff
@@ -318,6 +318,7 @@ github.com/ckaznocha/protoc-gen-lint v0.2.4 h1:Vp7LuGORwOdAFEqgBEfLcZyKqdtzrN8lc
 github.com/ckaznocha/protoc-gen-lint v0.2.4/go.mod h1:x9chIbwX2yctbChCqkuASKiEMT2XzxNr8KMpvrwcIGE=
```
RTann, line 318: hmm how come only the go.sum was updated?

pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,140 @@
...
+		repoToCPE:              repoToCPE,
+		repoToCPELocalFilename: path.Join(slimUpdaterDir, repoToCPEFilename),
+		repoToCPEURL:           repoToCPEURL,
+	}
+
+	return slimUpdater, nil
+}
+
+// RunForever starts the updater loop.
+func (u SlimUpdater) RunForever() {
```
RTann, line 91: nit: `u *SlimUpdater`

pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,140 @@
...
+			continue
+		case <-u.stopSig.Done():
+			return
+		}
+	}
+
+}
+
+// Stop stops the updater loop.
+func (u SlimUpdater) Stop() {
```
RTann, line 109: nit: `u *SlimUpdater`

cmd/clair/config.go
```diff
@@ -64,6 +67,8 @@ func DefaultConfig() Config {
 		MaxExtractableFileSizeMB:       tarutil.DefaultMaxExtractableFileSizeMB,
 		MaxELFExecutableFileSizeMB:     tarutil.DefaultMaxELFExecutableFileSizeMB,
 		MaxImageFileReaderBufferSizeMB: tarutil.DefaultMaxLazyReaderBufferSizeMB,
+		CentralEndpoint:                "https://central.stackrox.svc",
+		SensorEndpoint:                 "https://sensor.stackrox.svc",
```
RTann, line 71: Can you add a small test for this in `config_test.go`?

cmd/clair/main.go
```diff
@@ -144,7 +144,14 @@ func Boot(config *Config, slimMode bool) {
 	wg.Wait()
 	defer db.Close()
 
-	if !slimMode {
+	if slimMode {
+		u, err := updater.NewSlimUpdater(config.Updater, config.SensorEndpoint, repoToCPE)
+		if err != nil {
+			log.WithError(err).Fatal("failed to initialize slim updater")
```
RTann, line 150: nit: Uppercase `Failed` just to match the full-updater equivalent

pkg/updater/fetcher.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return false, errors.Wrap(err, "executing request")
+	}
+	defer utils.IgnoreError(resp.Body.Close)
+	if resp.StatusCode == http.StatusNotModified {
+		// Not modified
+		return false, nil
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		logrus.Warnf("vulnerability definition not found: %s", url)
```
RTann, line 45: nit: maybe just `definitions not found in %s` (mostly remove "vulnerability" since repo-to-cpe.json does not have vulns)

pkg/updater/genesis_manifests.go
```diff
@@ -51,16 +44,30 @@ func getRelevantDownloadURL(centralEndpoint string) (string, error) {
...
 	if err := validateUUID(uuid); err != nil {
+		return "", err
+	}
+	return uuid, err
+}
+
+// getRelevantDownloadURL gets the genesis manifests from the dump, finds the one
+// with the highest timestamp, and returns the location for the diff dump from that location.
+// This ensures that we get the smallest diff dump that works for this version of scanner.
+func getRelevantDownloadURL(centralEndpoint string) (string, error) {
```
RTann, line 63: is this solely used for Central communication? If so, let's try to specify that in the name

go.sum
jvdm: Due to this: https://app.circleci.com/pipelines/github/stackrox/scanner/26474/workflows/07119e8c-449c-4ace-b6fb-042d0c9e255f/jobs/262755/parallel-runs/0/steps/0-102

RTann: That seems odd to me... Nothing has changed and the master branch is fine.

RTann: ah perhaps it's because of the new direct dependency on `httputil` and other related packages

jvdm:
Turns out, we need to include the `genesis_manifest.json` in the scanner-slim images for this to work. It's [currently only copied to the regular scanner's image](https://github.com/stackrox/scanner/blob/master/image/scanner/rhel/Dockerfile#L61).

cmd/clair/main.go
```diff
@@ -132,19 +132,28 @@ func Boot(config *Config, slimMode bool) {
...
 		}()
-
-		wg.Add(1)
-		go func() {
-			defer wg.Add(-1)
-			repoToCPE = repo2cpe.Singleton()
-		}()
 	}
 
+	var repoToCPE *repo2cpe.Mapping
```
RTann, line 137: good catch

pkg/updater/genesis_manifests.go
RTann: In case you missed this ^

pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,142 @@
...
+
+const (
+	repoToCPEFilename = "rhelv2/repository-to-cpe.json"
+)
+
+var (
+	slimUpdaterDir = filepath.Join(wellknowndirnames.WriteableDir, "slim-updater-artifacts.d")
+)
+
+// SlimUpdater updates the Scanner's vulnerability definitions, contacting
```
RTann, line 31: nit: remove `vulnerability` since the slim scanner doesn't have vulns

pkg/updater/slim_updater.go
RTann: You can update the `go.mod` with the following:

1. In the `github.com/stackrox/rox => github.com/stackrox/stackrox` replace line, change the version to `github.com/stackrox/stackrox jvdm/ROX-9614/pull`
2. Run `go mod tidy`
3. You can use the `clientconn.NewHTTPClient()` form this point. You'd just have to be sure to update the version in the `go.mod` with `master` once your PR is merged and then `go mod tidy` again

pkg/updater/fetcher.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func newHTTPClient(transport http.RoundTripper) *http.Client {
+	return &http.Client{
+		Timeout:   defaultTimeout,
+		Transport: transport,
+	}
+}
+
+func fetchDumpFromURL(ctx concurrency.Waitable, client *http.Client, url string, lastUpdatedTime time.Time, outputPath string) (bool, error) {
```
RTann, line 28: assuming this is completely copy/pasted, right? I don't think there are any changes here?

pkg/updater/updater.go
```diff
@@ -175,16 +131,14 @@ func New(config Config, centralEndpoint string, db database.Datastore, repoToCPE
...
-	}
+	client := newHTTPClient(proxy.RoundTripper())
 	clientConfig, err := mtls.TLSClientConfigForCentral()
 	if err != nil {
 		return nil, errors.Wrap(err, "generating TLS client config for Central")
 	}
 	client.Transport = &http.Transport{
-		TLSClientConfig: clientConfig,
+		TLSClientConfig:    clientConfig,
+		DisableCompression: true,
```
RTann, line 141: 👍

cmd/clair/config.go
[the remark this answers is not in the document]
jvdm: Yes, they are going to be updated here as well.

pkg/updater/fetcher.go
jvdm: Correct.

pkg/updater/genesis_manifests.go
jvdm: That was also copied and pasted. When moving things like, I think it's better for everyone to not modify the code and make it easier on future readers that this was a C'n'P. We can refactor the name in a different PR.

go.mod
```diff
@@ -75,7 +68,10 @@ replace (
 	github.com/heroku/docker-registry-client => github.com/stackrox/docker-registry-client v0.0.0-20220204234128-07f109db0819
 	github.com/mattn/goveralls => github.com/viswajithiii/goveralls v0.0.3-0.20190917224517-4dd02c532775
 	github.com/nilslice/protolock => github.com/viswajithiii/protolock v0.10.1-0.20190117180626-43bb8a9ba4e8
-	github.com/stackrox/rox => github.com/stackrox/stackrox v0.0.0-20210914215712-04ef1c1cfc62
+
+	github.com/operator-framework/helm-operator-plugins => github.com/stackrox/helm-operator v0.0.8-0.20220506091602-3764c49abfb3
+	github.com/stackrox/rox => github.com/stackrox/stackrox v0.0.0-20220512161225-64d3c8bd40a8
```
jvdm, line 73: This was updated to bring in `clientconn.NewHTTPClient()` from `stackrox/stackrox`, but I am now 100% confident that's the best way to do it, and my knowledge of go's module dependency is very limited.

pkg/updater/updater.go
RTann: Can you leave a comment here why we do this? It may not be clear why we explicitly disable compression

pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,137 @@
...
+			"uuid": []string{uuid},
+			"file": []string{repoToCPEFilename},
+		})
+	if err != nil {
+		return nil, errors.Wrapf(err, "setting up sensor URL at %s", sensorEndpoint)
+	}
+
+	// Create sensor's HTTP client.
+	sensorClient, err := clientconn.NewHTTPClient(
+		mtls.SensorSubject, urlfmt.FormatURL(sensorEndpoint, urlfmt.NONE, urlfmt.NoTrailingSlash), defaultTimeout)
```
RTann, line 65: didn't know `urlfmt.NONE` existed. Makes life easier

pkg/updater/slim_updater.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "creating sensor client")
+	}
+
+	// Set up the repo2cpe local filename and its directory.
+	repoToCPELocalFilename := path.Join(slimUpdaterDir, repoToCPEFilename)
```
RTann, line 71: nit: `filepath` should be used for filesystem ops

pkg/repo2cpe/mapping.go
```diff
@@ -50,7 +50,11 @@ func NewMapping() *Mapping {
 
 // Load loads the contents of the RHELv2CPERepoName file in the given directory into the Mapping.
 func (m *Mapping) Load(dir string) error {
-	path := filepath.Join(dir, RHELv2CPERepoName)
+	return m.LoadFile(filepath.Join(dir, RHELv2CPERepoName))
+}
+
+// LoadFile loads the contents of the specified file into the Mapping.
+func (m *Mapping) LoadFile(path string) error {
```
RTann, line 57: Why not just use the `Load` function?

RTann (approved):
LGTM. Just a few tiny comments, then 🚢

pkg/updater/updater.go
c-du: +1

c-du (approved):
LGTM.

pkg/updater/slim_updater.go
jvdm: Is there a need for us to be concerned about being dir-separator-agnostic?

pkg/repo2cpe/mapping.go
jvdm: `Load() assumes the file's path within the directory. `LoadFile()` doesn't have that assumption and frees callers from having the setup of the file in a specific way, decoupling them.

pkg/updater/slim_updater.go
RTann: [filepath](https://pkg.go.dev/path/filepath) ensures the operations are compatible with the OS. Since we are opensource now, it's possible (for example) to someone to decide to build Scanner on a Windows machine, which would be incompatible with [path](https://pkg.go.dev/path)

jvdm: I mean, Scanner is a containerized app, and inside the container, we don't have to worry about the directory separator, right? This would also be true for building Scanner since it also happens in a container.

PR #719, opened 2022-05-02
Update Debian name mappings

RTann:
Debian 11 is "stable". See https://www.debian.org/releases/, https://www.debian.org/releases/testing/releasenotes, and https://deb.debian.org/debian/dists/

### DIFF
--- a/database/namespace_mapping.go
+++ b/database/namespace_mapping.go
@@ -25,13 +25,14 @@ var DebianReleasesMapping = map[string]string{
 	"stretch":  "9",
 	"buster":   "10",
 	"bullseye": "11",
+	"bookworm": "12",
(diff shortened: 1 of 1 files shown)


connorgorman:
How is stable used in our code?

RTann:
> How is stable used in our code?

The `DebianReleasesMapping` is used in both the `aptsources` package list handler and the debian vulnsrc fetcher. It is possible `stable`, `testing`, etc are used in the `aptsources`-related file. I don't think the vuln list tends to have it. From what I can tell, Claircore does not reference these names. Only the release's name. That being said, it doesn't hurt to account for it, as we have been doing just in case we miss something.

connorgorman (approved):

PR #727, opened 2022-05-11
ROX-10924: Add automated tests for slim scanner

RTann:
## Description

Currently, we only test the full Scanner/ScannerDB. We should add automated tests for Scanner-slim/ScannerSB-slim as well to catch slim-specific errors.

Scanner-slim only allows callers to use `/scannerV1.ImageScanService/GetImageComponents`, so that is the only endpoint tested. Non-vulnerability-related database queries are also tested, directly.


RTann:
The failed test for commit https://github.com/stackrox/scanner/pull/727/commits/d5dae014a4d03e02dba27d28c2428e05e77df0d1 shows ScannerDB-slim failed to start due to missing initdb directory. The next run for https://github.com/stackrox/scanner/pull/727/commits/acc76a0993063bd1f2a842be5e02db1c77b5a26a fixes it and everything passes :)

.circleci/config.yml
```diff
@@ -895,6 +916,14 @@ jobs:
       - provision-gke-cluster:
           cluster-id: rhel
 
+  provision-slim-cluster:
```
jvdm, line 919:
```suggestion
  provision-rhel-slim-cluster:
```

Makefile
```diff
@@ -219,6 +219,13 @@ deploy: clean-helm-rendered
 	helm template scanner chart/ --set tag=$(TAG),logLevel=$(LOGLEVEL),updateInterval=2m --output-dir rendered-chart
 	kubectl apply -R -f rendered-chart
 
+.PHONY: deploy-slim
+deploy-slim: clean-helm-rendered
+	@echo "+ $@"
+	kubectl create namespace stackrox || true
```
jvdm, line 225: Do we ignore failures here because the namespace might already exist, or is it something else?

e2etests/testcase_test.go
```diff
@@ -2484,19 +2484,20 @@ var testCases = []testCase{
...
-		namespace: "", // Fedora 28
-		unexpectedFeatures: []apiV1.Feature{
-			{
-				Name:          "p11-kit",
-				VersionFormat: "rpm",
-				Version:       "0.23.5-3.el7",
-			},
-		},
-	},
+	// TODO(ROX-10926): The image below no longer exists. Replace this image as well as the one above with other images which accomplish the same task.
```
jvdm, line 2487: We are handling that in a different [PR](https://github.com/stackrox/scanner/pull/728). You might need to handle conflicts or bring the follow-up being discussed there to this patch.

e2etests/testcase_test.go
```diff
@@ -2808,10 +2809,10 @@ var testCases = []testCase{
...
 					{
-						Name:          "CVE-2020-9488",
-						Description:   "Improper validation of certificate with host mismatch in Apache Log4j SMTP appender. This could allow an SMTPS connection to be intercepted by a man-in-the-middle attack which could leak any log messages sent through that appender. Fixed in Apache Log4j 2.12.3 and 2.13.1",
-						Link:          "https://nvd.nist.gov/vuln/detail/CVE-2020-9488",
-						Severity:      "Low",
+						Name:        "CVE-2020-9488",
+						Description: "Improper validation of certificate with host mismatch in Apache Log4j SMTP appender. This could allow an SMTPS connection to be intercepted by a man-in-the-middle attack which could leak any log messages sent through that appender. Fixed in Apache Log4j 2.12.3 and 2.13.1",
+						Link:        "https://nvd.nist.gov/vuln/detail/CVE-2020-9488",
+						Severity:    "Low",
 						Metadata: map[string]interface{}{
```
jvdm, line 2816: Nit: since you are making an effort to align all columns properly, shouldn't we align `Metadata:` as well? Or is it because it's a different type we don't align?

e2etests/utils_test.go
```diff
@@ -0,0 +1,25 @@
+// +build e2e slim_e2e
+
+package e2etests
+
+import (
+	"testing"
+
+	v1 "github.com/stackrox/scanner/api/v1"
```
jvdm, line 8: We don't need to translate the module's name here.

e2etests/utils_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func getMatchingFeature(t *testing.T, featureList []v1.Feature, featureToFind v1.Feature, allowNotFound bool) *v1.Feature {
+	candidateIdx := -1
```
jvdm, line 13: Use a `*v1.Feature` variable and check for `nil` rather than comparing to -1, this will make the code way simpler to understand, and would be OK with a named return.

jvdm (changes requested):
So far, so good, definitely rebase (or merge master) once you deal with the comments in https://github.com/stackrox/scanner/pull/728, then re-request a review. Nice work!

Makefile
RTann: Yeah, in case it already exists

e2etests/testcase_test.go
RTann: Yep, just put this here for now until that one is merged. Needed to ignore this image so I can confirm the tests all pass (and they fail when expected)

e2etests/testcase_test.go
RTann: This was formatted by `go fmt ./e2etests/...`, so this is how go, itself, formats structs. I've noticed it doesn't really align it when you specify a type like this

e2etests/utils_test.go
RTann: this is done automatically, idk why. it's like this everywhere and this was just copied over from elsewhere, so i'm going to keep it

e2etests/utils_test.go
RTann: this was also just copied over from its original file, so I may just keep it as-is for now

.circleci/config.yml
RTann: how come we want to add "rhel" here? I don't think it's needed since the other provision steps don't specify rhel in the name (just in the cluster-id)

e2etests/testcase_test.go
jvdm: Good to know.

.circleci/config.yml
jvdm: Makes sense.

jvdm (approved):
Tests for Scanner Slim are good at the time of comment. LGTM.

PR #728, opened 2022-05-11
Remove non-existent image for now

RTann:
## Description

Simply don't test with this image at this time. Finding a replacement image to test the same concept may be challenging, so I created a [Jira ticket for it.](https://issues.redhat.com/browse/ROX-10926).

### DIFF
--- a/e2etests/testcase_test.go
+++ b/e2etests/testcase_test.go
@@ -2484,19 +2484,20 @@ var testCases = []testCase{
 			},
 		},
 	},
-	{
-		image:     "quay.io/cgwalters/coreos-assembler@sha256:6ed6cd0006b6331d8cfd4a794afe7d2a87dc9019b80658a21b28d9941a97356d",
-		registry:  "https://quay.io",
-		source:    "NVD",
-		namespace: "", // Fedora 28
-		unexpectedFeatures: []apiV1.Feature{
-			{
-				Name:          "p11-kit",
-				VersionFormat: "rpm",
-				Version:       "0.23.5-3.el7",
-			},
-		},
-	},
+	// TODO: The image below no longer exists. Replace this image as well as the one above with other images which accomplish the same task.
+	//{
+	//	image:     "quay.io/cgwalters/coreos-assembler@sha256:6ed6cd0006b6331d8cfd4a794afe7d2a87dc9019b80658a21b28d9941a97356d",
+	//	registry:  "https://quay.io",
+	//	source:    "NVD",
+	//	namespace: "", // Fedora 28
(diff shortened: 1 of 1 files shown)


e2etests/testcase_test.go
```diff
@@ -2484,19 +2484,20 @@ var testCases = []testCase{
...
-		namespace: "", // Fedora 28
-		unexpectedFeatures: []apiV1.Feature{
-			{
-				Name:          "p11-kit",
-				VersionFormat: "rpm",
-				Version:       "0.23.5-3.el7",
-			},
-		},
-	},
+	// TODO(ROX-10926): The image below no longer exists. Replace this image as well as the one above with other images which accomplish the same task.
```
jvdm, line 2487: You don't need to comment the entry. Just remove it, and you can revert the commit when you want to enable it back.

RTann: I'll remove the image entry and move the TODO

e2etests/testcase_test.go
```diff
@@ -2468,6 +2468,8 @@ var testCases = []testCase{
 	// The next two images have the same layer and thus verify lineage checks between different images
 	// The first is a centos:7 image that has the package p11-kit. The second image is from fedora, and we
 	// can't identify the OS, so it should not have p11-kit.
+	// TODO(ROX-10926): quay.io/cgwalters/coreos-assembler@sha256:6ed6cd0006b6331d8cfd4a794afe7d2a87dc9019b80658a21b28d9941a97356d no longer exists.
+	// That image as well as this one should be replaced with two images which accomplish the same task.
```
jvdm, line 2472: You also don't need this TODO or comments. You have a JIRA ticket tracking the task to add the entry back, so these comments belong to the JIRA, not the code. This PR will lead to the JIRA if someone wants to follow up.

jvdm (changes requested):

e2etests/testcase_test.go
jvdm: That's what I meant. Remove the comment about the two images. After your comment, though, I am now asking myself, is there any value in testing `open-adventure`? Should we remove both tests, and the comments altogether?

RTann: There is still value, as it proves we can scan this centos 7 image to find this specific package. I'll remove the comments and update the ticket

RTann: Removed comments and updated the description of https://issues.redhat.com/browse/ROX-10926 so it's clear why this is happening and how to resolve it later

jvdm (approved):
LGTM.

PR #731, opened 2022-05-13
Update builder image

RTann:
Use `stackrox-build-<tag>` instead of `centos:8`.


BUILD_IMAGE_VERSION
```diff
@@ -0,0 +1 @@
+scanner-build-0.3.35-5-ge540580ff7
```
RTann, line 1: This version is based on this commit: https://github.com/stackrox/rox-ci-image/commit/e540580ff70f1dfb61b176dd09d1a29ca090a37e

gavin-stackrox: will you tag a 0.3.36?

PR #732, opened 2022-05-13
fix: Remove `path` and use `filepath`

jvdm:
## Description

Remove usage of `path` when interacting with the file system, replace it the `filepath`.

It allows unit tests to run in any development environment since they can create temporary directories. Also, it keeps the codebase ready for additional testing and running containerized environments in the future, if any.

## Tests

Relying on CI.


RTann (approved):
Thanks!

PR #740, opened 2022-05-17
ROX-10991: Fix block alignment for disk overflow

RTann:
`to := mathutil.MinInt64(((till-1)/r.maxBufferSize+1)*overflowBlockSize, r.size)` was meant to block-align the bytes written to the overflow file on disk; however, `r.maxBufferSize` should be `overflowBlockSize` for this to work properly.

Added a test to ensure a file can be read in-full, which fails with the original implementation. This test mimics a file of ~200MB by using a 201 byte string with `overflowBlockSize = 64 bytes` and `r.maxBufferSize = 100 bytes` (ie reduce the real-world values by a factor of 1021*1024 bytes)


pkg/ioutils/lazy_reader_disk.go
```diff
@@ -2,7 +2,6 @@ package ioutils
 
 import (
 	"io"
-	"io/ioutil"
```
RTann, line 5: `io/ioutil` is deprecated, so might as well remove it here while I'm touching this file

pkg/ioutils/lazy_reader_disk.go
```diff
@@ -163,36 +164,38 @@ func (r *diskBackedLazyReaderAt) ensureOverflowToDisk(till int64) {
 		r.pos = r.maxBufferSize
 	}
 
-	if r.overFlowFile == nil {
+	if r.overflowFile == nil {
 		var err error
+		// "" indicates we want to use os.TempDir().
```
RTann, line 169: making it clear why we do `_ = os.RemoveAll(filepath.Join(os.TempDir(), d.Name()))` in `CleanUpTempFiles`

pkg/ioutils/lazy_reader_disk.go
```diff
@@ -163,36 +164,38 @@ func (r *diskBackedLazyReaderAt) ensureOverflowToDisk(till int64) {
...
 			r.err = errors.Wrapf(err, "create overFlowFile %s", filePath)
 			return
 		}
 	}
 
-	// Copy until next block align with size r.maxBufferSize or the size of file.
+	// Copy until next block align with size overflowBlockSize or the size of file.
 	// Request to copy an extra byte to ensure EOF is recorded.
-	to := mathutil.MinInt64(((till-1)/r.maxBufferSize+1)*overflowBlockSize, r.size)
+	to := mathutil.MinInt64(((till-1)/overflowBlockSize+1)*overflowBlockSize, r.size)
```
RTann, line 193: `r.maxBufferSize` was the original block size, but it was switched over to `overflowBlockSize`; however, this division was missed.

pkg/ioutils/lazy_reader_disk_test.go
```diff
@@ -7,9 +7,33 @@ import (
...
+	// Read entire contents.
+	p := make([]byte, reader.Size())
+	n, err := lazyReaderAt.ReadAt(p, 0)
+	assert.NoError(t, err)
+	assert.Equal(t, int(reader.Size()), n)
+	assert.Equal(t, []byte(testString), p)
+}
+
 func TestDiskLazyReader(t *testing.T) {
+	overflowBlockSize = 30 // Set the overflow block size to something smaller than the test string.
```
RTann, line 35: After adding this and running the test with the original implementation, the `dataSize - 20` and `dataSize - 1` cases failed

jvdm (approved):

c-du (approved):

PR #742, opened 2022-05-18
Elf check shall not hide io error

c-du:
This is the other part why we did not find the reader issue earlier.


RTann (approved):

c-du:
That is covered by the TestIsELFExecutable.

PR #765, opened 2022-06-07
update ubuntu:22.04 test

RTann:
AddedBy changed as the base image changed and this is only a single layer image. Append the image's specific digest so we don't need to update the AddedBy in the future.


jvdm (commented):
@RTann, could you please update the description or share the details on what triggered the test failures, especially if it was external or internal? Should we capture a long-term improvement that can avoid future issues like this one, or your `AddedBy` trick should be sufficient?

I acknowledge that these are probably unnecessary if you have the context around these e2e tests, how they work, and what they do. But sharing in the PR can benefit people coming to touch the code base.

RTann:
> @RTann, could you please update the description or share the details on what triggered the test failures, especially if it was external or internal? Should we capture a long-term improvement that can avoid future issues like this one, or your `AddedBy` trick should be sufficient?
> 
> I acknowledge that these are probably unnecessary if you have the context around these e2e tests, how they work, and what they do. But sharing in the PR can benefit people coming to touch the code base.

@jvdm tests started failing because the `ubuntu:22.04` image updated. It is a single-layer image, so the `AddedBy` layer changed due to the image update. This PR pins the `ubuntu:22.04` we use for testing to a specific version of `ubuntu:22.04` by appending the digest to the image name. Adding the digest like this prevents this from happening again for this image (caveat it's possible this digest is deleted, but highly unlikely)

connorgorman (approved):

PR #774, opened 2022-06-09
ROX-11335: Move Scanner DB secret out of main Scanner DB container

RTann:
This PR makes a few PostgreSQL-related updates:

* Move Scanner DB's secret out of the main Scanner DB container
    * This moves the secret into the init container, as it is only needed there
* Some Dockerfile cleanup

We also start looking at initContainer logs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -650,7 +650,10 @@ jobs:
             - updater-bin
 
   create-postgres-dump-from-genesis-dump:
-    <<: *defaults
+    docker:
+      - *defaultImage
+      - image: postgres:12-alpine
+    working_directory: *defaultWorkingDirectory
     steps:
       - checkout
       - check-label-to-run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -652,9 +652,7 @@ jobs:
   create-postgres-dump-from-genesis-dump:
     docker:
       - *defaultImage
-      - image: postgres:12-alpine
-        environment:
-          POSTGRES_PASSWORD: password
+      - image: postgres:12.0-alpine
     working_directory: *defaultWorkingDirectory
     steps:
       - checkout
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -652,7 +652,9 @@ jobs:
   create-postgres-dump-from-genesis-dump:
     docker:
       - *defaultImage
-      - image: postgres:12.0-alpine
+      - image: postgres:12-alpine
+        environment:
+          POSTGRES_PASSWORD: password
```
RTann, line 657: Took this from the stackrox repo, but we should probably not expose this nor make it obvious as maybe it's possible for someone who has access to the CI container to mess with the data. Any thoughts @gavin-stackrox ?

RTann: Note: we don't actually use this password for our needs. It was just required to set this environment variable in the latest postgres 12 version

RTann:
Removing reviewers for now, as it is not as ready as i originally thought

connorgorman (approved):
I would prefer we unify the Central and Scanner scripts and YAMLs. Does the way that Central-db do initialization work for scanner db?

connorgorman (changes requested):
sorry meant to request changes and not approve

RTann:
> I would prefer we unify the Central and Scanner scripts and YAMLs. Does the way that Central-db do initialization work for scanner db?

Central does not initialize the database with data like Scanner does/needs: https://github.com/stackrox/stackrox/blob/3.70.0/image/templates/helm/stackrox-central/templates/01-central-12-central-db.yaml#L43. Central solely calls `docker-entrypoint.sh initdb ...` which just does `initdb` https://github.com/stackrox/stackrox/blob/3.70.0/image/postgres/scripts/docker-entrypoint.sh#L341. This just initializes the directory. We need to ensure the initial data is populated https://github.com/stackrox/scanner/blob/master/image/db/rhel/scripts/docker-entrypoint.sh#L336

connorgorman (approved):

PR #775, opened 2022-06-09
ROX-9309: Require a password for any local connections to postgres

RTann:
Require a password to access pgsql locally. We opt for `md5` cryptographic hashing.

Verified this works by deploying StackRox locally with this version of Scanner/ScannerDB, execing into the ScannerDB pod and trying to access `psql`:

```
[postgres@scanner-db-566cc8c545-ffjkv /]$ psql
Password for user postgres: 
psql: error: FATAL:  password authentication failed for user "postgres"

...

[postgres@scanner-db-566cc8c545-ffjkv /]$ psql
Password for user postgres: 
psql (12.11)
Type "help" for help.

postgres=#
```


RTann:
This is based on https://github.com/stackrox/scanner/pull/776, which separates the db integration tests from the e2e tests, so we can use a local postgres for testing purposes without needing to retrieve and use the k8s secret. That PR should be reviewed first

database/pgsql/pgsql.go
```diff
@@ -114,7 +114,9 @@ func openDatabase(registrableComponentConfig database.RegistrableComponentConfig
 		}
 		src = fmt.Sprintf("%s password=%s", pg.config.Source, password)
 	} else if !os.IsNotExist(err) {
-		log.Errorf("error stating password file %q: %v", passwordFile, err)
+		log.Errorf("pgsql: error stating password file %q; continuing without one: %v", passwordFile, err)
```
connorgorman, line 117: Why do we continue without a password?

image/db/pg_hba.conf
```diff
@@ -81,16 +81,16 @@
...
+host    all             all             127.0.0.1/32            md5
 # IPv6 local connections:
-host    all             all             ::1/128                 trust
+host    all             all             ::1/128                 md5
 # Allow replication connections from localhost, by a user with the
 # replication privilege.
-local   replication     all                                     trust
-host    replication     all             127.0.0.1/32            trust
-host    replication     all             ::1/128                 trust
+local   replication     all                                     md5
```
connorgorman, line 91: Just change the replication ones to reject

database/pgsql/pgsql.go
RTann: DB integration tests do not require a password. We allow Scanner to continue without crashing or anything, but if the database requires it, then it'll fail all connections/queries

connorgorman: Yeah, I think this is suboptimal if people mess up and break the scanner password because it'll just spew errors instead of being straightforward in terms of what's wrong. Can we only allow this in a test-focused function?

RTann: I am ok with us erroring out here for this `else if !os.IsNotExist(err)` since there was some kind of error with the underlying fielsystem/OS; however, for the `else`, I think it's ok to continue because we know no password file was not used.

Also keep in mind this is nothing new. This is just updating the logs for something we already do.

RTann: I enforce a password in non-tests now

database/database.go
```diff
@@ -50,7 +50,7 @@ var drivers = make(map[string]Driver)
 
 // Driver is a function that opens a Datastore specified by its database driver type and specific
 // configuration.
-type Driver func(RegistrableComponentConfig) (Datastore, error)
+type Driver func(cfg RegistrableComponentConfig, passwordRequired bool) (Datastore, error)
```
RTann, line 53: I added `passwordRequired bool` instead of making it a possible field in `RegistrableComponentConfig` as I do not want users to adjust this setting (this is part of Scanner's ConfigMap). I wanted to hardcode where password should be required vs not

database/pgsql/pgsql.go
```diff
@@ -106,22 +106,26 @@ func openDatabase(registrableComponentConfig database.RegistrableComponentConfig
...
-	} else if !os.IsNotExist(err) {
-		log.Errorf("error stating password file %q: %v", passwordFile, err)
-	}
-
 	dbName, pgSourceURL, err := parseConnectionString(pg.config.Source)
 	if err != nil {
 		return nil, err
 	}
 
+	src := pg.config.Source
```
RTann, line 114: Moved this down as `dbName, pgSourceURL, err := parseConnectionString(pg.config.Source)` uses `pg.config.Source` instead of `src`

database/pgsql/pgsql.go
```diff
@@ -106,22 +106,26 @@ func openDatabase(registrableComponentConfig database.RegistrableComponentConfig
...
+		if _, err := os.Stat(passwordFile); err == nil {
+			password, err := os.ReadFile(passwordFile)
+			if err != nil {
+				return nil, errors.Wrapf(err, "pgsql: could not load password file %q", passwordFile)
+			}
+			src = fmt.Sprintf("%s password=%s", pg.config.Source, password)
+		} else if !os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "pgsql: could not stat password file %q", passwordFile)
+		} else {
+			return nil, errors.Errorf("pgsql: no password file at expected location %q", passwordFile)
```
connorgorman, line 125: can unindent this

connorgorman (approved):

database/pgsql/pgsql.go
RTann: true

RTann: lol oops forgot it is needed because the if {} sets `src = `

PR #776, opened 2022-06-10
Separate DB integration tests to a separate job

RTann:
There is no need to run the DB integration tests using a live Scanner DB, so might as well move the tests elsewhere so they do not depend on a pre-existing deployment.

Also use the postgres:12-bullseye image instead of postgres:12.0-alpine for two reasons:
1. Let's us use the latest version of PostgreSQL 12
2. We took the docker-entrypoint.sh from the [Debian version](https://github.com/docker-library/postgres/blob/master/12/bullseye/docker-entrypoint.sh), so this gives us a sense of consistency (kind of)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -582,6 +582,8 @@ jobs:
     docker:
       - *defaultImage
       - image: postgres:12-bullseye
+        environment:
+          POSTGRES_PASSWORD: password
     working_directory: *defaultWorkingDirectory
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -582,7 +582,7 @@ jobs:
     docker:
       - *defaultImage
       - image: postgres:12-bullseye
-      working_directory: *defaultWorkingDirectory
+    working_directory: *defaultWorkingDirectory
     steps:
       - checkout
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -406,27 +406,7 @@ commands:
             .circleci/collect-service-logs.sh kube-system
           when: always
 
-      - run:
-          name: Verify the scanner did not restart
-          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -406,28 +406,8 @@ commands:
             .circleci/collect-service-logs.sh kube-system
           when: always
 
-      - run:
```
RTann, line 409: We do this again lower, so I removed this one, as there is no need to collect the logs twice

.circleci/config.yml
```diff
@@ -406,28 +406,8 @@ commands:
...
-          name: Verify the scanner did not restart
```
RTann, line 410: same with this

.circleci/config.yml
```diff
@@ -406,28 +406,8 @@ commands:
...
-            fi
-            cat nohup.out || true
-          when: always
-
       - fetch-and-upload-scanner-metrics
 
-      - run-db-integration-tests:
-          is-slim: << parameters.is-slim >>
-
-      - run:
```
RTann, line 424: logs are collected above, so no need to do this twice

.circleci/config.yml
```diff
@@ -613,6 +578,22 @@ jobs:
           name: Run style checks
           command: make style
 
+  db-integration-tests:
+    docker:
+      - *defaultImage
+      - image: postgres:12-bullseye
+        environment:
+          POSTGRES_PASSWORD: password
```
RTann, line 586: I am not sure how to best proceed with this

.circleci/config.yml
```diff
@@ -652,7 +633,9 @@ jobs:
   create-postgres-dump-from-genesis-dump:
     docker:
       - *defaultImage
-      - image: postgres:12.0-alpine
+      - image: postgres:12-bullseye
+        environment:
+          POSTGRES_PASSWORD: password
```
RTann, line 638: same with this

gavin-stackrox (approved):

PR #784, opened 2022-06-15
ROX-11519: Onboard E2E tests to OSCI

RTann:
Tested with https://github.com/openshift/release/pull/29534


gavin-stackrox:
I'll come back with a full review when you are feature complete.

But in the meantime: I would keep `scripts/ci/jobs` for files that are a 1:1 match with test names (AKA jobs) coming from openshift release. That is what we are striving for on the stackrox/stackrox side, so if we want to maintain *maximum symmetry* I'd follow that. (With the python wrapper **** this is not followed but I hope to remedy that).

RTann:
@gavin-stackrox moved deploy.sh from ci/jobs to ci/

RTann:
@gavin-stackrox @jvdm https://github.com/openshift/release/pull/29534 passes now, so this is ready for review. https://github.com/openshift/release/pull/29807 is the PR for the master branch

gavin-stackrox:
@RTann looking through the test rehearse log I see: ```scripts/ci/deploy.sh: line 5: /go/src/github.com/stackrox/scripts/lib.sh: No such file or directory```. Probably a 'ROOT' source path setup wrong after the file move (though the failure is silently ignore)

gavin-stackrox:
@RTann I think you will need this: https://github.com/stackrox/stackrox/blob/4d3fb539547d1935a35040e4a4e8c258a53a92e4/scripts/ci/lib.sh#L773 once the e2e test starts running in regular PRs. At least this is what was required for stackrox/e2e tests to get them to stop interacting with the OpenShift cluster.

scripts/ci/e2e-tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
RTann, line 1: @gavin-stackrox I didn't know where else to put this because I had to put `e2e_tests.py` in `scripts/ci/jobs`. Perhaps I should put this under something like `scripts/ci/jobs/e2etests`? I think it's just a little awkward because this is essntially the same script as `scripts/ci/jobs/unit-tests.sh`

RTann: I put it in `scripts/ci/jobs/e2etests` but let me know if you'd prefer something else

gavin-stackrox: @RTann I think it is fine there.

This python layer makes it a bit awkward. On the stackrox side the intent was not to repeat the shared behavior among the e2e tests but given that scanner has 1 e2e test, going `dispatch.sh -> e2e_tests.py -> ci_tests.py -> E2eTest() -> e2e-tests.sh` is confusing.

RTann: I'm planning on 1-2 follow up PR(s) to add support for `slim-e2e-tests` and `scale-tests` which will require the same infrastructure as the `e2e-tests`. They will probably also need very similar `.sh` files as well

gavin-stackrox (approved):
👍 I'm going to approve on the basis that we'll refactor the duplication across the repos at some later date

PR #793, opened 2022-06-27
Add runAsNonRoot to pod's security context

RTann:
Let the kubelet validate the image does not run as UID 0 (root). Putting this in the pod's security context means it applies to all containers in the pod (anything in a specific container's security context will take precedence).

By adding this, we cannot use named user/group for the `USER` field in the Dockerfiles, so this also reverts `USER postgres:postgres` back to `USER 70:70`

Also move the `runAsGroup` and `runAsUser` to the pod-level, as the two containers have the same security context, otherwise.

### DIFF
--- a/chart/templates/db-deployment.yaml
+++ b/chart/templates/db-deployment.yaml
@@ -32,9 +32,6 @@ spec:
         - name: scanner-db-password
           mountPath: /run/secrets/stackrox.io/secrets
           readOnly: true
-        securityContext:
-          runAsUser: 70
-          runAsGroup: 70
       containers:
       - name: db
         image: {{.Values.scannerDBImage}}:{{.Values.tag}}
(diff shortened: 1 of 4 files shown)


RTann:
@msugakov the CI run for the first commit (https://github.com/stackrox/scanner/pull/793/commits/9c1ba86f2a70505e2e2525184f0d8645715c7dcb) seems to have been able to deploy Scanner and Scanner DB just fine without changing the Dockerfile?

msugakov (approved):

msugakov:
@RTann I don't know why that run worked. Unfortunately there are no yaml dumps of resources to see if things were correctly applied.

msugakov:
The problem is that describe output is not the same as `kubectl get pod ... -o yaml` output. The latter is more rich.
I suspect the issue did not occur here since you provided numeric user and group ids right in `securityContext`. Maybe in my experiment I did not have that and k8s relied on user/group from the Dockerfile.
I will take a look but most likely tomorrow because I'm in the final meeting for today.

msugakov:
Yes, it was most likely that what I just wrote
https://github.com/stackrox/stackrox/blob/372b7a04e3efc8cd64386055f79d61d4c96a5a68/operator/bundle/manifests/rhacs-operator.clusterserviceversion.yaml#L815-L816

I guess it's up to you if you want to go back to `USER postgres:postgres` in the `Dockerfile`.

PR #795, opened 2022-06-28
ROX-11538, ROX-11539: Migrate slim and scale tests to OSCI

RTann:
Tested with https://github.com/openshift/release/pull/29887


gavin-stackrox (approved):
LGTM

PR #802, opened 2022-07-01
use https instead of ssh for git

RTann:
OSCI does not forward `SSH_AUTH_SOCK`, so I don't think we can use ssh creds anymore for this. Good news: these repos are open sourced, so we don't really need creds! So, let's switch to HTTPS auth instead


viswajithiii (approved):
Sweet!

PR #813, opened 2022-07-08
ROX-11082, ROX-11083: Migrate diff dumps to OSCI

RTann:
This step combines both the `create-diff-dumps` CircleCI step and the `upload-diff-dumps-and-offline-dumps` step into a single OSCI step, `diff-dumps`.

This is tested here: https://github.com/openshift/release/pull/30301

This PR will be followed up with a PR to the openshift/release repo to add the step and then a PR to this repo to remove the diff-dumps steps from CircleCI from PR/merge builds, as there is no need to upload diff dumps twice. We will keep it in the hourly builds for now, as hourly builds have not migrated to OSCI yet


scripts/ci/jobs/diff-dumps.sh
```diff
@@ -0,0 +1,130 @@
...
+        expected_zip_file_loc="/tmp/diff-dumps/dump${idx}/diff.zip"
+        [[ -f "${expected_zip_file_loc}" ]]
+        if [[ -z "${diffUUID}" ]]; then
+            continue
+        fi
+        echo "Found file at ${expected_zip_file_loc}"
+        du -skh "${expected_zip_file_loc}"
+        cmd=()
+        # Note: CIRCLE_TAG is defined via openshift_ci_mods in scripts/ci/lib.sh.
+        if [[ "$branch" != "master" && -z "${CIRCLE_TAG}" ]]; then
```
RTann, line 51: @gavin-stackrox I pretty blindly took this from circleci expect I set branch to `branch=$(get_base_ref)` instead of `CRICLE_BRANCH`. Do you think this is an ok way to check if this is master or tag or do we have a better way that I missed?

gavin-stackrox (approved):

PR #846, opened 2022-07-25
Bump Scanner to UBI9

jvdm:
## Description

1. Bump Scanner and Scanner Slim image to UBI9.
1. Re-build the rpm database when analyzing rpm packages.

The goal of (1.) is to get `rpm >= 4.16` to support RHEL9 scanning. The goal of (2.) is to ensure the rpm database format in the image is migrated to the latest to avoid issues with compatibility between the Scanner's rpm and the images'. In particular, Scanner's rpm (UBI9) only has `bdb_ro` mode enabled, [which doesn't support querying indexes in BDB-based databases](https://github.com/rpm-software-management/rpm/pull/1578).

## Tests

- Local deployment (`make deploy`) followed by E2E tests.
- OCI Rehearsal: https://github.com/openshift/release/pull/31088
- CI

### DIFF
--- a/pkg/rpm/database.go
+++ b/pkg/rpm/database.go
@@ -143,6 +143,19 @@ func CreateDatabaseFromLayer(imageFiles tarutil.LayerFiles) (*rpmDatabase, error
 			return nil, commonerr.ErrFilesystem
 		}
 	}
+	// Rebuild the rpm database, it will recreate indexes and convert old formats
+	// to the latest supported by the current rpm version.
+	dbCmd := exec.Command(
+		"rpmdb",
+		"--dbpath", dbDir,
+		"--rebuilddb",
+	)
+	var errBuffer bytes.Buffer
+	dbCmd.Stderr = &errBuffer
+	if err := dbCmd.Run(); err != nil {
+		logrus.Warnf("failed to rebuild the rpm database: %s", errBuffer.String())
+		return nil, errors.Wrap(err, "failed to rebuild rpm database")
+	}
 	return &rpmDatabase{
 		dbPath: dbDir,
 	}, nil
--- a/.openshift-ci/Dockerfile.build_root
+++ b/.openshift-ci/Dockerfile.build_root
@@ -10,4 +10,4 @@
 # - `make update` and commit the results
 # - run `/test pj-rehearse-max` on the openshift/release PR to validate the change
 
-FROM quay.io/stackrox-io/apollo-ci:scanner-test-0.3.43
+FROM quay.io/stackrox-io/apollo-ci:scanner-test-0.3.45
--- a/BUILD_IMAGE_VERSION
+++ b/BUILD_IMAGE_VERSION
@@ -1,1 +1,1 @@
-scanner-build-0.3.43
+scanner-build-0.3.45
--- a/image/scanner/rhel/Dockerfile
+++ b/image/scanner/rhel/Dockerfile
@@ -1,12 +1,12 @@
 ARG BASE_REGISTRY=registry.access.redhat.com
-ARG BASE_IMAGE=ubi8-minimal
-ARG BASE_TAG=8.6
+ARG BASE_IMAGE=ubi9-minimal
+ARG BASE_TAG=9.0.0
 
 FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG} AS extracted_bundle
 
 COPY bundle.tar.gz /
 WORKDIR /bundle
-RUN microdnf install tar gzip && tar -zxf /bundle.tar.gz
+RUN microdnf -y install tar gzip && tar -zxf /bundle.tar.gz
 
 FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG} AS base
 
@@ -24,9 +24,9 @@ COPY --from=extracted_bundle /bundle/scanner ./
 
 COPY --from=extracted_bundle /bundle/THIRD_PARTY_NOTICES/ /THIRD_PARTY_NOTICES/
 
-RUN microdnf upgrade && \
-    microdnf install xz && \
-    microdnf clean all && \
+RUN microdnf -y upgrade && \
+    microdnf -y install xz && \
+    microdnf -y clean all && \
     # (Optional) Remove line below to keep package management utilities
     # We don't uninstall rpm because scanner uses it to get packages installed in scanned images.
     rpm -e --nodeps $(rpm -qa curl '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
--- a/image/scanner/rhel/Dockerfile.slim
+++ b/image/scanner/rhel/Dockerfile.slim
@@ -1,12 +1,12 @@
 ARG BASE_REGISTRY=registry.access.redhat.com
-ARG BASE_IMAGE=ubi8-minimal
-ARG BASE_TAG=8.6
+ARG BASE_IMAGE=ubi9-minimal
+ARG BASE_TAG=9.0.0
 
 FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG} AS extracted_bundle
 
 COPY bundle.tar.gz /
 WORKDIR /bundle
-RUN microdnf install tar gzip && tar -zxf /bundle.tar.gz
+RUN microdnf -y install tar gzip && tar -zxf /bundle.tar.gz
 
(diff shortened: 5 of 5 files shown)


pkg/rpm/database.go
```diff
@@ -143,6 +143,19 @@ func CreateDatabaseFromLayer(imageFiles tarutil.LayerFiles) (*rpmDatabase, error
 			return nil, commonerr.ErrFilesystem
 		}
 	}
+	// Rebuild the rpm database, it will recreate indexes and convert old formats
+	// to the latest supported by the current rpm version.
+	dbCmd := exec.Command(
```
RTann, line 148: hmm this should not be needed when just trying to read the Berkeley DB rpm database. I tried this locally on an image based on UBI 9 and was able to query the packages database I copied over from a UBI 8-based image without needing this step. Simply just `rpm --dbpath /tmp/rpm --query --all` (I put `Packages` inside of `/tmp/rpm`)

RTann: This was needed for the failing E2E test?

jvdm: @RTann:

> I tried this locally on an image based on UBI 9

Were you using `bdb_ro` database backend?

I would say probably not, but need to check first.

UBI9's rpm uses `bdb_ro` and not `bdb`, the linked RPM PR should give you the context of why this is a problem for us.

Translating the DB puts us on the "safe" side, and I am Ok with the extra overhead of rebuilding the DB even when not strictly necessary. But let me know.

RTann: I did the following:

1. `docker save -o whatever.tar registry.access.redhat.com/ubi8/ubi:8.6`
2. Extract `var/lib/rpm/Packages` from the base layer
3. Build docker image*, `docker build -t test:1 Dockerfile .`
4. Run `docker run -it test:1 bash`
5. Run `rpm --dbpath /tmp/rpm --query --all --queryformat "%{name}\n%{evr}\n%{ARCH}\n%{RPMTAG_MODULARITYLABEL}\n[%{FILENAMES}\n].\n"`
6. Success

* Dockerfile contents:
```
FROM registry.access.redhat.com/ubi9-minimal:9.0.0

RUN mkdir -p /tmp/rpm
COPY ./Packages /tmp/rpm/Packages
```

Note: I extracted the `Packages` file from the UBI 8 image to my current directory

RTann: `bdb_ro` tells me we should have no problems reading from the Berkeley DB, but we cannot write to it, which is fine for us, as we simply just want to read

RTann: Looked at the linked PR https://github.com/rpm-software-management/rpm/pull/1578, and it sounds like just simply looking at `Packages` is fine, which is all we care about

jvdm: @RTann:

> I did the following:

(Your `docker build` needs an `-f`.)

You queried for all packages, which is OK with `bdb_ro`. However, the `-qa` query doesn't use indexes. The problem arises when we need secondary indexes. As its name states, it is a "read-only driver" and won't create or modify the DB. So, when doing indexed queries, for example, using `--whatprovides`, it fails:

```
% docker run  test:1 rpm --dbpath /tmp/rpm -qal | grep /usr/bin/bash
warning: Found bdb_ro Packages database while attempting sqlite backend: using bdb_ro backend.
warning: could not open /tmp/rpm/Name: No such file or directory
warning: could not open /tmp/rpm/Basenames: No such file or directory
warning: could not open /tmp/rpm/Group: No such file or directory
warning: could not open /tmp/rpm/Requirename: No such file or directory
warning: could not open /tmp/rpm/Providename: No such file or directory
warning: could not open /tmp/rpm/Conflictname: No such file or directory
warning: could not open /tmp/rpm/Obsoletename: No such file or directory
warning: could not open /tmp/rpm/Triggername: No such file or directory
warning: could not open /tmp/rpm/Dirnames: No such file or directory
warning: could not open /tmp/rpm/Installtid: No such file or directory
warning: could not open /tmp/rpm/Sigmd5: No such file or directory
warning: could not open /tmp/rpm/Sha1header: No such file or directory
warning: could not open /tmp/rpm/Filetriggername: No such file or directory
warning: could not open /tmp/rpm/Transfiletriggername: No such file or directory
warning: could not open /tmp/rpm/Recommendname: No such file or directory
warning: could not open /tmp/rpm/Suggestname: No such file or directory
warning: could not open /tmp/rpm/Supplementname: No such file or directory
warning: could not open /tmp/rpm/Enhancename: No such file or directory
/usr/bin/bash
/usr/bin/bashbug
/usr/bin/bashbug-64
%
%
% docker run  test:1 rpm --dbpath /tmp/rpm -q --whatprovides /usr/bin/bash
warning: Found bdb_ro Packages database while attempting sqlite backend: using bdb_ro backend.
warning: could not open /tmp/rpm/Name: No such file or directory
warning: could not open /tmp/rpm/Basenames: No such file or directory
warning: could not open /tmp/rpm/Group: No such file or directory
warning: could not open /tmp/rpm/Requirename: No such file or directory
warning: could not open /tmp/rpm/Providename: No such file or directory
warning: could not open /tmp/rpm/Conflictname: No such file or directory
warning: could not open /tmp/rpm/Obsoletename: No such file or directory
warning: could not open /tmp/rpm/Triggername: No such file or directory
warning: could not open /tmp/rpm/Dirnames: No such file or directory
warning: could not open /tmp/rpm/Installtid: No such file or directory
warning: could not open /tmp/rpm/Sigmd5: No such file or directory
warning: could not open /tmp/rpm/Sha1header: No such file or directory
warning: could not open /tmp/rpm/Filetriggername: No such file or directory
warning: could not open /tmp/rpm/Transfiletriggername: No such file or directory
warning: could not open /tmp/rpm/Recommendname: No such file or directory
warning: could not open /tmp/rpm/Suggestname: No such file or directory
warning: could not open /tmp/rpm/Supplementname: No such file or directory
warning: could not open /tmp/rpm/Enhancename: No such file or directory
file /usr/bin/bash is not owned by any package
descending% 
```

> bdb_ro tells me we should have no problems reading from the Berkeley DB, but we cannot write to it, which is fine for us, as we simply just want to read

That is not what it means. It means it is OK to read the RPM DB (based on Berkely DB), but we won't modify anything, including creating any secondary indexes.

> Looked at the linked PR rpm-software-management/rpm#1578, and it sounds like just simply looking at Packages is fine, which is all we care about

If you read the PR, you will see that while `rpm` yields warnings [when it cannot find the secondary indexes](https://github.com/rpm-software-management/rpm/pull/1578/files#diff-c8cf1feb936ea210009d352febda1a727790927b519dcd0a525d24d69ee37167R612) before it did not. The author argued that moving these errors to warnings allows queries to the main DB even when the indexes are not available. But some queries are expected to fail. And that's fine since the driver's primary use case is migrating rather than backward compatible. Hence, why I linked the PR.

This discussion is why I think migrating the DB and rebuilding indexes is an effective way to avoid compatibility issues between RPM versions. But I am open to discussing other options.

RTann: > > I did the following:
> 
> (Your `docker build` needs an `-f`.)

Just wrote it incorrectly. I meant `docker build -t test:1 .`

> So, when doing indexed queries, for example, using --whatprovides, it fails

That is the part I was asking about originally. I didn't know `--whatprovides` used the secondary indexes, so I didn't know why this was added. It seemed more like a "just in case this becomes an issue in the future"

RTann: Do we know anything about the performance of this? Do we always have to do this or can we just limit it to RHEL < 9? I also wonder if it would be better to do this lazily (as in, only if there are any detected `languageComponents` in the same layer. Perhaps only when `rpm.CreateDatabaseFromLayer` is called from `AnnotateComponentsWithPackageManagerInfo`)?

jvdm: @RTann:

> Do we know anything about the performance of this?

Empirically it was roughly the same as an `rpm -qa`, in user time. This makes sense when we think it is rebuilding indexes.

Notice, for the previous implementation based on `rpm` with `bdb` enabled, the indexes were already re-generated automatically at the first `-q`. I am expecting we should not feel any additional impact by preemptively doing it once before querying all packages.

> Do we always have to do this or can we just limit it to RHEL < 9?

Yes, this and the other options are possible. But I think it doesn't worth the trade-off. At least from a performance standpoint vs. guaranteeing queries will likely work. Re-indexing every time is simpler.

RTann: Checked that `rpm` does, in fact, generate indexes in version 4.14.x

RTann (approved):

PR #847, opened 2022-07-26
Check for ELF executables and scripts that start with shebang

connorgorman:
gitlab EE image returned 6500 paths

Now it returns 1300 paths including .so's

TBD if we can remove .so from the path

### DIFF
--- a/ext/featurefmt/driver.go
+++ b/ext/featurefmt/driver.go
@@ -139,10 +139,14 @@ func TestLister(t *testing.T, l Lister, testData []TestData) {
 func AddToDependencyMap(filename string, fileData tarutil.FileData, execToDeps, libToDeps database.StringToStringsMap) {
 	if fileData.Executable {
 		deps := set.NewStringSet()
-		if fileData.ELFMetadata != nil {
-			deps.AddAll(fileData.ELFMetadata.ImportedLibraries...)
+		if elfMeta := fileData.ELFMetadata; elfMeta != nil {
+			deps.AddAll(fileData.elfMeta.ImportedLibraries...)
+			if !elfMeta.SharedObject {
+				execToDeps[filename] = deps
+			}
+		} else {
+			execToDeps[filename] = deps
 		}
-		execToDeps[filename] = deps
 	}
 	if fileData.ELFMetadata != nil {
 		for _, soname := range fileData.ELFMetadata.Sonames {
--- a/ext/featurefmt/driver.go
+++ b/ext/featurefmt/driver.go
@@ -140,7 +140,7 @@ func AddToDependencyMap(filename string, fileData tarutil.FileData, execToDeps, 
 	if fileData.Executable {
 		deps := set.NewStringSet()
 		if elfMeta := fileData.ELFMetadata; elfMeta != nil {
-			deps.AddAll(fileData.elfMeta.ImportedLibraries...)
+			deps.AddAll(elfMeta.ImportedLibraries...)
 			if !elfMeta.SharedObject {
 				execToDeps[filename] = deps
 			}
--- a/ext/featurefmt/driver.go
+++ b/ext/featurefmt/driver.go
@@ -141,12 +141,8 @@ func AddToDependencyMap(filename string, fileData tarutil.FileData, execToDeps, 
 		deps := set.NewStringSet()
 		if elfMeta := fileData.ELFMetadata; elfMeta != nil {
 			deps.AddAll(elfMeta.ImportedLibraries...)
-			if !elfMeta.SharedObject {
-				execToDeps[filename] = deps
-			}
-		} else {
-			execToDeps[filename] = deps
 		}
+		execToDeps[filename] = deps
 	}
 	if fileData.ELFMetadata != nil {
 		for _, soname := range fileData.ELFMetadata.Sonames {
--- a/pkg/elf/elf.go
+++ b/pkg/elf/elf.go
@@ -9,6 +9,8 @@ import (
 	"github.com/stackrox/rox/pkg/utils"
 )
 
+const df1PIE = 0x08000000
+
 var (
 	allowedELFTypeList = set.NewFrozenIntSet(int(elf.ET_EXEC), int(elf.ET_DYN))
 )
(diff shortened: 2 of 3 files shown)


pkg/tarutil/tarutil.go
```diff
@@ -164,17 +165,31 @@ func ExtractFiles(r io.Reader, filenameMatcher matcher.Matcher) (LayerFiles, err
...
+			executable, _ := executableMatcher.Match(filename, hdr.FileInfo(), contents)
 			if hdr.Size > maxELFExecutableFileSize {
 				log.Warnf("Skipping ELF executable check for file %q (%d bytes) because it is larger than the configured maxELFExecutableFileSizeMB of %d", filename, hdr.Size, maxELFExecutableFileSize/1024/1024)
-			} else if hdr.Size >= elfHeaderSize { // Only bother attempting to get ELF metadata if the file is large enough for the ELF header.
-				fileData.ELFMetadata, err = elf.GetExecutableMetadata(contents)
-				if err != nil {
-					log.Errorf("Failed to get dependencies for %s: %v", filename, err)
+			} else if executable {
+				if hdr.Size >= elfHeaderSize { // Only bother attempting to get ELF metadata if the file is large enough for the ELF header.
+					fileData.ELFMetadata, err = elf.GetExecutableMetadata(contents)
```
c-du, line 173: One issue is some library file are not marked as executable if it does not contain main() entry.

connorgorman: Ah gotcha, so we want to check all files for ELF

connorgorman: will make the change

pkg/tarutil/tarutil.go
```diff
@@ -164,17 +165,29 @@ func ExtractFiles(r io.Reader, filenameMatcher matcher.Matcher) (LayerFiles, err
...
-					log.Errorf("Failed to get dependencies for %s: %v", filename, err)
+			} else {
+				if hdr.Size >= elfHeaderSize { // Only bother attempting to get ELF metadata if the file is large enough for the ELF header.
+					fileData.ELFMetadata, err = elf.GetExecutableMetadata(contents)
+					if err != nil {
+						log.Errorf("Failed to get dependencies for %s: %v", filename, err)
+					}
+				}
+				if executable && fileData.ELFMetadata == nil {
+					shebangBytes := make([]byte, 2)
```
RTann, line 179: nit: instead of 2, do `len(shebangHeader)`

pkg/tarutil/tarutil.go
```diff
@@ -164,17 +165,29 @@ func ExtractFiles(r io.Reader, filenameMatcher matcher.Matcher) (LayerFiles, err
...
-				if err != nil {
-					log.Errorf("Failed to get dependencies for %s: %v", filename, err)
+			} else {
+				if hdr.Size >= elfHeaderSize { // Only bother attempting to get ELF metadata if the file is large enough for the ELF header.
+					fileData.ELFMetadata, err = elf.GetExecutableMetadata(contents)
+					if err != nil {
+						log.Errorf("Failed to get dependencies for %s: %v", filename, err)
+					}
+				}
+				if executable && fileData.ELFMetadata == nil {
```
RTann, line 178: so if this file is smaller than `maxELFExecutableFileSize`, was determined to not be an ELF file (due to failure or it's just not an ELF file), and has the executable bit set, then we declare this file is executable iff it is larger than 2 bytes and has a shebang header?

RTann: What if the script doesn't have the shebang header? Isn't that possible?

pkg/tarutil/tarutil.go
```diff
@@ -164,17 +165,29 @@ func ExtractFiles(r io.Reader, filenameMatcher matcher.Matcher) (LayerFiles, err
...
+					shebangBytes := make([]byte, 2)
+					if hdr.Size > 2 {
```
RTann, line 180: nit: instead of 2, do len(shebangHeader) or maybe make a global var that is set to `len(shebangHeader)`

pkg/tarutil/tarutil.go
connorgorman: @c-du Does this mean that we can't find the dependencies for this executable anyways if it is actually ELF?

connorgorman: > What if the script doesn't have the shebang header? Isn't that possible?
This also means that the process will come in like `bash -c script` so it's different scenario altogether

pkg/elf/elf.go
```diff
@@ -49,5 +50,6 @@ func GetExecutableMetadata(r io.ReaderAt) (*Metadata, error) {
 	return &Metadata{
 		Sonames:           sonames,
 		ImportedLibraries: libraries,
+		SharedObject:      elfFile.Type == elf.ET_DYN,
```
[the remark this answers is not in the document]
connorgorman: I will test this, but I believe I implemented logic that will pick up pie executables

ext/featurefmt/driver.go
```diff
@@ -139,8 +139,8 @@ func TestLister(t *testing.T, l Lister, testData []TestData) {
 func AddToDependencyMap(filename string, fileData tarutil.FileData, execToDeps, libToDeps database.StringToStringsMap) {
 	if fileData.Executable {
 		deps := set.NewStringSet()
-		if fileData.ELFMetadata != nil {
-			deps.AddAll(fileData.ELFMetadata.ImportedLibraries...)
+		if elfMeta := fileData.ELFMetadata; elfMeta != nil {
```
RTann, line 142: up to you, but this change can probably be removed

pkg/tarutil/tarutil.go
```diff
@@ -164,17 +166,28 @@ func ExtractFiles(r io.Reader, filenameMatcher matcher.Matcher) (LayerFiles, err
...
 
+			executable, _ := executableMatcher.Match(filename, hdr.FileInfo(), contents)
 			if hdr.Size > maxELFExecutableFileSize {
 				log.Warnf("Skipping ELF executable check for file %q (%d bytes) because it is larger than the configured maxELFExecutableFileSizeMB of %d", filename, hdr.Size, maxELFExecutableFileSize/1024/1024)
 			} else if hdr.Size >= elfHeaderSize { // Only bother attempting to get ELF metadata if the file is large enough for the ELF header.
 				fileData.ELFMetadata, err = elf.GetExecutableMetadata(contents)
 				if err != nil {
 					log.Errorf("Failed to get dependencies for %s: %v", filename, err)
 				}
+			} else if executable && hdr.Typeflag != tar.TypeLink && fileData.ELFMetadata == nil {
```
RTann, line 177: I think `fileData.ELFMetadata == nil` is always `true`?

c-du (approved):

RTann (commented):
So we now define executable as either:

* ELF file
* a file with a `#!`

This will filter out a file without the `#!`, though, right?

pkg/tarutil/tarutil.go
connorgorman: oh good catch, this block should not be here

PR #852, opened 2022-07-27
Set GKE zone upon teardown

RTann:
Ensure we teardown the correct zone, which was set in the `create_cluster` step


sbostick (approved):
LGTM

PR #867, opened 2022-08-05
proto v3.20.1

RTann:
Let's match the proto version here https://github.com/stackrox/stackrox/blob/master/make/protogen.mk#L37


RTann:
Closing in favor of #872

PR #872, opened 2022-08-07
Update protobuf tooling

misberner:
This ports over recent (and some future, namely https://github.com/stackrox/stackrox/pull/2650) changes to the `protogen.mk` protobuf build logic from stackrox/stackrox to scanner, with the aim of (a) staying up-to-date and (b) reducing divergence between the two.

### DIFF
--- a/make/protogen.mk
+++ b/make/protogen.mk
@@ -1,4 +1,6 @@
 BASE_PATH ?= $(CURDIR)
+# Set to empty string to echo some command lines which are hidden by default.
+SILENT ?= @
 
 # GENERATED_API_XXX and PROTO_API_XXX variables contain standard paths used to
 # generate gRPC proto messages, services, and gateways for the API.
@@ -24,7 +26,7 @@ GENERATED_API_SWAGGER_SPECS = $(API_SERVICE_PROTOS:%.proto=$(GENERATED_BASE_PATH
 ## Protobuf ##
 ##############
 # Set some platform variables for protoc.
-PROTOC_VERSION := 3.6.1
+PROTOC_VERSION := 21.4
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
 PROTOC_ARCH = linux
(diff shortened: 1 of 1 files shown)


make/protogen.mk
```diff
@@ -24,7 +26,7 @@ GENERATED_API_SWAGGER_SPECS = $(API_SERVICE_PROTOS:%.proto=$(GENERATED_BASE_PATH
 ## Protobuf ##
 ##############
 # Set some platform variables for protoc.
-PROTOC_VERSION := 3.6.1
+PROTOC_VERSION := 21.4
```
RTann, line 29: coincidentally, I was also going to bump this version up https://github.com/stackrox/scanner/pull/867

make/protogen.mk
```diff
@@ -1,4 +1,6 @@
 BASE_PATH ?= $(CURDIR)
+# Set to empty string to echo some command lines which are hidden by default.
+SILENT ?= @
```
RTann, line 3: I saw this in the stackrox repo, but I'm wondering why we have this

make/protogen.mk
```diff
@@ -118,7 +120,7 @@ GRPC_GATEWAY_DIR = $(shell go list -f '{{.Dir}}' -m github.com/grpc-ecosystem/gr
 .PHONY: proto-fmt
 proto-fmt: $(PROTOC_GEN_LINT)
 	@echo "Checking for proto style errors"
-	@PATH=$(PROTO_GOBIN) $(PROTOC) \
+	$(SILENT)PATH=$(PROTO_GOBIN) $(PROTOC) \
```
RTann, line 123: why `$(SILENT)` here but `@` above?

RTann: same for some others below this

RTann (approved):

make/protogen.mk
misberner: The idea is that you can easily debug a makefile (i.e., print what commands are being executed) by running it with `make SILENT= target`, not having to manually remove the `@` on all the lines that you are interested in.

misberner: The idea is that you can easily debug a makefile (i.e., print what commands are being executed) by running it with `make SILENT= target`, not having to manually remove the `@` on all the lines that you are interested in.

make/protogen.mk
misberner: Because the `echo` is never interesting from a debug perspective

PR #877, opened 2022-08-10
Fix vulnerability and package relationship for RHEL

RTann:
To reduce disk utilization, we separated vulnerabilities from the related packages, and we related them via the vulnerability's name (ex: [CVE-2022-1650](https://access.redhat.com/security/cve/cve-2022-1650)). However, when a vulnerability (for example [CVE-2022-1650](https://access.redhat.com/security/cve/cve-2022-1650)) has a different severity for different packages/environments, this becomes a problem, as multiple entries from the `vuln` table will be returned. We just simply took the first entry, which means for something like `dotnet`, the vulnerability entry we returned could have been attributed with severity `Important` instead of the correct severity, `Moderate`.

This PR fixes the relationship between `vuln` and `vuln_package` by using the vulnerability's hash as the "foreign key" instead of just the name. That way, we attribute the package to the correct vulnerability entry, which will give us the correct severity.

Creating a migration for this would have been a headache, as we need to correctly align the entries in `vuln_package` to `vuln`, as they are not correct now. This does not seem very easy to do with pure SQL. However, ScannerDB is currently *not* persisted, so we can get away with not performing a DB migration.

These changes will *not* affect live Scanners, as those Scanners come with their own already-made genesis dump and any diff we perform comparing the genesis dumps with the most up-to-date data do not involve the Postgres tables, but rather the Go struct `RHELv2Vulnerability`, which it not affected.

### DIFF
--- a/database/pgsql/rhelv2_vulnerability_test.go
+++ b/database/pgsql/rhelv2_vulnerability_test.go
@@ -3,12 +3,11 @@
 package pgsql
 
 import (
-	"testing"
-
-	"github.com/stackrox/scanner/pkg/archop"
 	"github.com/stackrox/scanner/database"
+	"github.com/stackrox/scanner/pkg/archop"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"testing"
 )
 
 func TestRHELv2Vulnerability(t *testing.T) {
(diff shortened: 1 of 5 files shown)


database/pgsql/rhelv2_vulnerability_test.go
```diff
@@ -188,6 +188,76 @@ func TestRHELv2Vulnerability(t *testing.T) {
...
+			CPE: "cpe:/o:redhat:enterprise_linux:8::baseos",
+		},
+	}
+
+	vulns, err := datastore.GetRHELv2Vulnerabilities(records)
+	assert.NoError(t, err)
+	assert.Len(t, vulns, 1)
+	assert.Len(t, vulns[0], 1)
+	vuln := vulns[0][0]
+	assert.Equal(t, "Moderate", vuln.Severity)
```
RTann, line 258: Without the fix, this would have been `Important`

c-du (approved):

RTann:
Need to check and test if instead of updating existing entries we just make new entries in the tables. It is preferred to update existing entries, so not continuously increase the DB size upon update

RTann:
Going to close this to rethink if there is another/better way to do this

RTann:
Replaced by https://github.com/stackrox/scanner/pull/895

PR #878, opened 2022-08-16
Update UID and PATH for generate-db-dump

RTann:
Received a warning about using a UID out of range: `adduser warning: pg's uid 71 outside of the UID_MIN 1000 and UID_MAX 60000 range.`

This step fails on `initdb`, though, because it cannot find it in the path. That's because it's no longer in `/usr/bin` as it used to be. So, let's update the `PATH` used by the `pg` user. The error message is:
```
/go/src/github.com/stackrox/scanner/scripts/ci/postgres.sh: line 10: initdb: command not found
```

Also, forgot to source lib.sh in postgres.sh, so I can see `info: No menu item 'Starting Postgres' in node '(dir)Top'`

Tested this via the following:

1. `docker run -it quay.io/stackrox-io/apollo-ci:scanner-test-0.3.45 bash`
2. `git clone https://github.com/stackrox/scanner && cd scanner`
3. Update the .openshift-ci/build/generate-db-dump.sh to comment out `openshift_ci_mods` and `gate_job generate-db-dump`, as they are not necessary and will prevent the test from working
4. Commented out anything under the `runuser` line as well inside the `generate_db_dump` function because they are not needed for the test
5. Run `./.openshift-ci/build/generate-db-dump.sh`

### DIFF
--- a/.openshift-ci/build/generate-db-dump.sh
+++ b/.openshift-ci/build/generate-db-dump.sh
@@ -16,7 +16,7 @@ generate_db_dump() {
     adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
 
     # The PATH is not completely preserved, so set the PATH here to ensure postgres-related commands can be found.
-    runuser -l pg -c "PATH=$PATH:/usr/pgsql-$PG_MAJOR/bin/ $ROOT/scripts/ci/postgres.sh start_postgres"
+    runuser -l pg -c "PATH=$PATH $ROOT/scripts/ci/postgres.sh start_postgres"
 
     "$ROOT/bin/updater" load-dump --postgres-host 127.0.0.1 --postgres-port 5432 --dump-file /tmp/genesis-dump/genesis-dump.zip
 
--- a/.openshift-ci/build/generate-db-dump.sh
+++ b/.openshift-ci/build/generate-db-dump.sh
@@ -12,10 +12,10 @@ gate_job generate-db-dump
 generate_db_dump() {
     info "Generating DB dump"
 
-    groupadd -g 71 pg
-    adduser pg -u 71 -g 71 -d /var/lib/postgresql -s /bin/sh
+    groupadd -g 1001 pg
+    adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
 
-    runuser -l pg -c "$ROOT/scripts/ci/postgres.sh start_postgres"
+    runuser -l pg -c "PATH=$PATH:/usr/pgsql-$PG_MAJOR/bin/ $ROOT/scripts/ci/postgres.sh start_postgres"
 
     "$ROOT/bin/updater" load-dump --postgres-host 127.0.0.1 --postgres-port 5432 --dump-file /tmp/genesis-dump/genesis-dump.zip
 
--- a/.openshift-ci/build/generate-db-dump.sh
+++ b/.openshift-ci/build/generate-db-dump.sh
@@ -15,6 +15,7 @@ generate_db_dump() {
     groupadd -g 1001 pg
     adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
 
+    # The PATH is not completely preserved, so set the PATH here to ensure postgres-related commands can be found.
     runuser -l pg -c "PATH=$PATH:/usr/pgsql-$PG_MAJOR/bin/ $ROOT/scripts/ci/postgres.sh start_postgres"
 
     "$ROOT/bin/updater" load-dump --postgres-host 127.0.0.1 --postgres-port 5432 --dump-file /tmp/genesis-dump/genesis-dump.zip
--- a/scripts/ci/postgres.sh
+++ b/scripts/ci/postgres.sh
@@ -2,6 +2,9 @@
 
 # Adapted from https://github.com/stackrox/stackrox/blob/master/scripts/ci/jobs/go-postgres-tests.sh
 
+SCRIPTS_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
+source "$SCRIPTS_ROOT/scripts/lib.sh"
+
 set -euo pipefail
 
 start_postgres() {


.openshift-ci/build/generate-db-dump.sh
```diff
@@ -12,10 +12,10 @@ gate_job generate-db-dump
 generate_db_dump() {
     info "Generating DB dump"
 
-    groupadd -g 71 pg
-    adduser pg -u 71 -g 71 -d /var/lib/postgresql -s /bin/sh
+    groupadd -g 1001 pg
+    adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
 
-    runuser -l pg -c "$ROOT/scripts/ci/postgres.sh start_postgres"
+    runuser -l pg -c "PATH=$PATH:/usr/pgsql-$PG_MAJOR/bin/ $ROOT/scripts/ci/postgres.sh start_postgres"
```
jvdm, line 18: I actually fixed the `PATH` [when updating the base test CI image](https://github.com/stackrox/rox-ci-image/commit/da336c78463a6597a938d2b781d7b82dfff42c33#diff-f849cde71446382ccc46e79ca6640ed1aec3c28388622eb99e61e339c7a70a91R38). I think what we need to do is to update the `FROM` from the `.openshift-ci/Dockerfile*`. Does that make sense?

The changes for higher group/user IDs is OK. Although, why don't we simply use the next available ID or use [`-r` (`--system`)](https://man7.org/linux/man-pages/man8/groupadd.8.html)?

jvdm (changes requested):

.openshift-ci/build/generate-db-dump.sh
```diff
@@ -12,10 +12,10 @@ gate_job generate-db-dump
 generate_db_dump() {
     info "Generating DB dump"
 
-    groupadd -g 71 pg
-    adduser pg -u 71 -g 71 -d /var/lib/postgresql -s /bin/sh
+    groupadd -g 1001 pg
```
jvdm, line 15: Although, why don't we simply use the next available ID or use [`-r` (`--system`)](https://man7.org/linux/man-pages/man8/groupadd.8.html)?

jvdm: Keeping it a system GID/UID might be slightly better because GN/Linux distros sometimes have configurations in place that only affects one of the types, like being more permissive for users. Not sure if this would be relevant in a container env, though.

.openshift-ci/build/generate-db-dump.sh
jvdm: Humm, I actually checked the build log one more time, and I noticed that [PATH was correctly set to include `/usr/pgsql-12/bin`](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-stackrox-scanner-master-merge-e2e-tests/1559333718210908160#1:build-log.txt%3A124).

jvdm: I think we should look into using `--preserve-environment` when using `runserver`, so that the environment is not lost in the subprocess (in this case, the shell running `postgres.sh`).

jvdm:
Could you please link the failures and warnings from OCI into the PR, to enhance the context around the change?

Could you also please clarify why `generate-db-dump` is not exercised pre-merge to master?

.openshift-ci/build/generate-db-dump.sh
jvdm: Argh, `--preserve-environment` is ignored when using `--login`, as per [the man page](https://manpages.debian.org/testing/util-linux/runuser.1.en.html).

jvdm: I guess then just passing `PATH=$PATH` would do the trick. Adding a comment on why would be nice too.

RTann: added a comment

.openshift-ci/build/generate-db-dump.sh
RTann: I didn't know that existed :) that being said, I think being explicit is ok, as it gives us full control of these values. Also, I don't want to rely on the value of `USERGROUPS_ENAB`, as I don't know what it is and if it could change if we update the base image, so I don't want something like `adduser` to create a new group for us https://linux.die.net/man/8/adduser

RTann:
> Could you please link the failures and warnings from OCI into the PR, to enhance the context around the change?

I copy and pasted the error messages since OSCI deletes artifacts rather quickly (I think after 12 hours)

> Could you also please clarify why `generate-db-dump` is not exercised pre-merge to master?

In `openshift/release`? Sometimes rehearsal jobs are skipped. You need to comment `/pj-rehearsal-max` (something like that) to try to get all your jobs to rehearse. As for this repo, `generate-db-dump` only runs on master, on tag, and when a PR is labelled with `generate-dumps-on-pr`. This is to decrease the amount of time needed for CI when new dumps are not necessary

.openshift-ci/build/generate-db-dump.sh
jvdm: Please, use `PATH=$PATH` instead of `PATH=$PATH:/usr/pgsql-$PG_MAJOR/bin/`. First, because the base image already set up the PATH correctly and it's easier to leave that knowledge in one place. Second, it's redundant.

scripts/ci/postgres.sh
```diff
@@ -2,6 +2,9 @@
 
 # Adapted from https://github.com/stackrox/stackrox/blob/master/scripts/ci/jobs/go-postgres-tests.sh
 
+SCRIPTS_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
+source "$SCRIPTS_ROOT/scripts/lib.sh"
+
```
jvdm, line 7: Why this was added?

jvdm (changes requested):

.openshift-ci/build/generate-db-dump.sh
jvdm: Interesting. How changes in the UID when rebuilding the base image could affect us?

scripts/ci/postgres.sh
RTann: Updated the description, but it's because we use `info` and `die` here, and those are defined in `scripts/lib.sh`. Got an error message from CI about it:

`info: No menu item 'Starting Postgres' in node '(dir)Top'`

jvdm:
> In `openshift/release`? Sometimes rehearsal jobs are skipped. You need to comment `/pj-rehearsal-max` (something like that) to try to get all your jobs to rehearse. As for this repo, `generate-db-dump` only runs on master, on tag, and when a PR is labelled with `generate-dumps-on-pr`. This is to decrease the amount of time needed for CI when new dumps are not necessary

Thanks for the explanation. I've gone ahead and added the `generate-db-dump` to the PR, so we don't have to merge to confirm the fix.

.openshift-ci/build/generate-db-dump.sh
RTann: I think it could be useful for debugging purposes. Also, I don't think there is a concise way to get the group ID given a group name https://askubuntu.com/questions/639990/what-is-the-group-id-of-this-group-name. Might as well just set one ourselves

scripts/ci/postgres.sh
jvdm: :+1:

jvdm:
Regarding why removing the label, I asked offline, the reply was:

> I don’t think adding the label will work here because OSCI will use the scripts as defined in the master branch

So the only option would be rehearsing at `openshift/release`. :-(

.openshift-ci/build/generate-db-dump.sh
jvdm: Humm, OK about debugging.

Anyway, you don't have to worry about `getent` unless on NSS, simply parsing `/etc/groups` would do it.

jvdm (approved):

PR #879, opened 2022-08-16
Bump CI images to latest scanner-test-0.3.45

jvdm:
## Purpose

Onboard UBI9 to all CI images by bumping the base image to the latest.

## Tests

Relying on CI

### DIFF
--- a/.openshift-ci/build/Dockerfile.build-bundle
+++ b/.openshift-ci/build/Dockerfile.build-bundle
@@ -1,4 +1,4 @@
-FROM quay.io/stackrox-io/apollo-ci:scanner-test-0.3.43
+FROM quay.io/stackrox-io/apollo-ci:scanner-test-0.3.45
 
 COPY . /go/src/github.com/stackrox/scanner
 WORKDIR /go/src/github.com/stackrox/scanner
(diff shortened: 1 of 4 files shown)


RTann (approved):
We can do this, but it actually doesn't matter, as these images are overridden in OpenShift CI. 

For example:
```
build_root:
  project_image:
    dockerfile_path: .openshift-ci/Dockerfile.build_root
images:
- dockerfile_path: .openshift-ci/build/Dockerfile.generate-genesis-dump
  from: src
  to: genesis-dump
- dockerfile_path: .openshift-ci/build/Dockerfile.generate-db-dump
  from: genesis-dump
  to: db-dump
- dockerfile_path: .openshift-ci/build/Dockerfile.build-bundle
  from: genesis-dump
  to: scanner-bundle
- dockerfile_path: .openshift-ci/build/Dockerfile.build-db-bundle
  from: db-dump
  to: scanner-db-bundle
- dockerfile_path: Dockerfile
  from: ubi-minimal
  inputs:
    scanner-bundle:
      paths:
      - destination_dir: .
        source_path: /go/src/github.com/stackrox/scanner/image/scanner/rhel/.
  to: scanner
```
The `build_root` specifies the `root` image to use in CI. On top of that, there is a `src` image which created based on the `root`, which is essentially `root` with the desired repo cloned. From there, we build the `genesis-dump` image and the `FROM` in the Dockerfile is replaced with the image represented by `src`. This same process happens for the other steps. For example, `scanner-bundle`'s Dockerfile's `FROM` is replaced with the image represented by `genesis-dump`. So the FROMs in these images aren't really used anyway

jvdm:
> So the FROMs in these images aren't really used anyway

I know. But unfortunately, they are here. So, unless we can replace them with a generic `FROM` I think it is worth aligning them with the actual `scanner-test` tag we use to avoid confusion. And I guess you can still build them locally for testing.

PR #885, opened 2022-08-19
ROX-12261: Add severity to node vulnerabilities

RTann:
We had not been setting the severity rating for node vulnerabilities, so this PR adds it.


md2119 (approved):
lgtm

api/v1/convert/convert.go
```diff
@@ -107,6 +107,7 @@ func NVDVulns(nvdVulns []*nvdtoolscache.NVDCVEItemWithFixedIn) ([]*v1.Vulnerabil
 			Link:        nvd.Link(vuln.CVE.CVEDataMeta.ID),
 			MetadataV2:  Metadata(m),
 			FixedBy:     vuln.FixedIn,
+			Severity:    string(DatabaseSeverityToSeverity(m.GetDatabaseSeverity())),
```
md2119, line 110: why is `Severity` a string field and not enum?

RTann: not sure I think the gRPC API predates my StackRox tenure

PR #894, opened 2022-08-25
Add go1.17 directives to db integration test files

RTann:
Add the go1.17 directives to all remaining files which were missing it. The older format is deprecated and will be removed in go1.18

### DIFF
--- a/database/pgsql/complex_test.go
+++ b/database/pgsql/complex_test.go
@@ -1,3 +1,4 @@
+//go:build db_integration
 // +build db_integration
 
 // Copyright 2017 clair authors
--- a/database/pgsql/feature_test.go
+++ b/database/pgsql/feature_test.go
@@ -1,3 +1,4 @@
+//go:build db_integration || slim_db_integration
 // +build db_integration slim_db_integration
 
 // Copyright 2016 clair authors
--- a/database/pgsql/keyvalue_test.go
+++ b/database/pgsql/keyvalue_test.go
@@ -1,3 +1,4 @@
+//go:build db_integration || slim_db_integration
 // +build db_integration slim_db_integration
 
 // Copyright 2016 clair authors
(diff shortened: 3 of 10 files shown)


jvdm (changes requested):
> The older format is deprecated and will be removed in go1.18

We should be able to remove the old format since it is already deprecated. If not, update the description with the reason. Thanks.

RTann:
> > The older format is deprecated and will be removed in go1.18
> 
> We should be able to remove the old format since it is already deprecated. If not, update the description with the reason. Thanks.

According to https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives, it looks like if `// +build` is completely removed, then older versions of Go which depend on parts of our code which used to have that directive will no longer work. That being said, I don't see this being a problem at all for this particular situation, but I figure if go1.17 allows them to coexist, then I don't see why we don't allow it, too.

If you think we should still just remove it, then that's fine, too

jvdm:
> That being said, I don't see this being a problem at all for this particular situation, but I figure if go1.17 allows them to coexist, then I don't see why we don't allow it, too.

It's a little bit confusing to have the two. It opens to questions like: Do we have a problem if I remove it? Is there something depending on these tags? Can I remove them? If there is no actual reason to keep them (as in "I don't see this being a problem at all for this particular situation"), then I think a cleaner approach is to remove them.

jvdm (commented):
> That being said, I don't see this being a problem at all for this particular situation, but I figure if go1.17 allows them to coexist, then I don't see why we don't allow it, too.

It's a little bit confusing to have the two. It opens to questions like: Do we have a problem if I remove it? Is there something depending on these tags? Is there an older version of Go somewhere depending on it?

> If you think we should still just remove it, then that's fine, too

If there is no actual reason to keep them (as in "I don't see this being a problem at all for this particular situation"), then I think a cleaner approach is to remove them.

PR #898, opened 2022-08-30
ROX-12238: Create interface for images files from tarutil

jvdm:
This PR slightly modifies the interface used to explore the image filesystem during analysis.  It replaces the `tarutil.LayerFiles` with an interface and introduces methods where its implementation detail leaks.  It also moves the interface and supporting type `tarutil.FileData` into the `analyzer` package. We are doing this to support node scanning.

In container image analysis, image content is retrieved as tarballs from registries, and streamed through the network.  Currently, `tarutil.LayerFiles` exists to encapsulate tarball manipulation and allow in-memory random access to intermediate image (layer) files[^1].  Random access to image content is necessary for O.S.-based package analysis because the details we want can only be identified by looking at multiple files in the intermediate images[^2].  For example, listing rpm packages require access to the rpm database, content manifests, and executable files.

This work paves the way for a different implementation of `analyzer.Files` based on an actual filesystem.  A hypothetical `nodescanning.FilesystemFiles`, or similar, will retrieve data from mounted volumes in the node scanning container rather than an image layer tar archive.

Now, the underlying file data source is already randomly accessible, so why do we need a `LayerFiles` in node scanning?  If we provide an `Layerfiles` we can reuse the same exact O.S. package detection and analysis code.  We don't necessarily need to store all contents in memory like intermediate images, but we need to find a way to handle errors gracefully since `FileData` hides them. A discussion to happen in a subsequent PR.

[^1]: This is not needed for language component analysis since each language component file is analyzed individually to identify components. But we still use the "image file map" to coordinate the information with O.S.-level components.

[^2]: With `LayerFiles`, the analyzer can randomly check for file existence or retrieve its content or metadata.  But not all the files added or removed from a layer are put into the `LayerFiles` map.  Instead, we filter them based on filename patterns and file attributes.  [A step delegated to "matchers"](https://github.com/stackrox/scanner/blob/15cccdf47932743b5aec81ef92e10510acd46d04/worker.go#L230).

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -166,10 +166,9 @@ func ProcessLayerFromReader(datastore database.Datastore, imageFormat, name, lin
 	return files, datastore.InsertLayerComponents(layer.Name, lineage, languageComponents, files.GetRemovedFiles(), opts)
 }
 
-func detectFromFiles(files tarutil.LayerFiles, name string, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool) (*database.Namespace, bool, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
+func detectFromFiles(files analyzer.Files, name string, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool) (*database.Namespace,
+	[]database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
 	namespace := DetectNamespace(name, files, parent, uncertifiedRHEL)
-
-	distroless := isDistroless(files) || (parent != nil && parent.Distroless)
 
 	var featureVersions []database.FeatureVersion
 	var rhelfeatures *database.RHELv2Components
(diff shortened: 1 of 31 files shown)


worker.go
```diff
@@ -247,7 +241,11 @@ func DetectContentFromReader(reader io.ReadCloser, format, name string, parent *
 		log.WithFields(log.Fields{logLayerName: name, "component count": len(m.components)}).Debug("detected components")
 	}
 
-	namespace, distroless, features, rhelv2Components, languageComponents, err := detectFromFiles(*files, name, parent, m.components, uncertifiedRHEL)
+	namespace, features, rhelv2Components, languageComponents, err := detectFromFiles(*files, name, parent, m.components, uncertifiedRHEL)
+	distroless := isDistroless(*files) || (parent != nil && parent.Distroless)
+	if !env.LanguageVulns.Enabled() {
```
connorgorman, line 246: Maybe not in this PR, but it seems like we're going to do a bunch of work to find language components and store a bunch of data just to toss them later. I think this check should probably be much earlier

jvdm: Agreed, I can actually do it in a different PR.

connorgorman (approved):
approving, but let's follow up as discussed to ignore language components earlier

PR #902, opened 2022-09-02
Update style checks

RTann:
Update style checks to more closely align with https://github.com/stackrox/stackrox.

Changes include:

* https://github.com/stackrox/stackrox/pull/2869
* https://github.com/stackrox/stackrox/pull/2887
* Aligning with https://github.com/stackrox/stackrox/blob/62a99a65b4f06d0c61779282dd4e845341d4b36d/.golangci.yml and adjusting for Scanner's needs
* Aligning with https://github.com/stackrox/stackrox/blob/62a99a65b4f06d0c61779282dd4e845341d4b36d/Makefile#L160 and adjusting for Scanner's needs


Makefile
```diff
@@ -92,29 +95,33 @@ build-updater: deps
 ## Style ##
 ###########
 .PHONY: style
-style: blanks golangci-lint staticcheck no-large-files
+style: golangci-lint blanks newlines staticcheck check-service-protos no-large-files
```
janisz, line 98:
```suggestion
style: golangci-lint blanks newlines check-service-protos no-large-files
```
`staticcheck` is included in golangci https://github.com/stackrox/scanner/pull/902/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R109

janisz (approved):

.golangci.yml
```diff
@@ -73,21 +76,22 @@ linters:
     # - dupl
     # - errcheck
     # - funlen
+    # - forbidigo
```
janisz, line 79: Looks weird to add commented line. But I understand the goal is to make it similar to stackrox

Makefile
RTann: good point

PR #904, opened 2022-09-03
ROX-12238: Isolate O.S. matchers to re-use by node scanning

jvdm:
Create singletons for two different types of "matcher" (O.S. features and active vulnerability), and add a new matcher interface called `PrefixMatcher` that offers a method to retrieve the list of directories to look for files.

Node scanning will exclusively use the O.S. matcher. And the list of directories will be used by the node scanner to determine which directories to walk and extract files from. This allows node scanning to be specific on the directories to walk and ignore the other files that are irrelevant during the extraction. Notice that, some analyzers (also known as detectors) do not have a deterministic set of files to scan, they look at prefixes, we still need to walk in directories. 

## Tests

UT and CI

### DIFF
--- a/pkg/matcher/matcher.go
+++ b/pkg/matcher/matcher.go
@@ -77,22 +77,22 @@ func (m *allowlistMatcher) GetCommonPrefixDirs() []string {
 // referenced. It does it by doing creating a trie-like structure with the
 // directory tree filtering paths with only single-children nodes.
 func findCommonDirPrefixes(prefixes []string) []string {
-	pre := make(map[string]set.StringSet)
+	prefixToSubdirs := make(map[string]set.StringSet)
 	for _, d := range prefixes {
 		for d != "" {
 			p, _ := path.Split(strings.TrimSuffix(d, "/"))
-			s := pre[p]
+			s := prefixToSubdirs[p]
 			s.Add(d)
-			pre[p] = s
+			prefixToSubdirs[p] = s
 			d = p
 		}
 	}
 	// Work on one step below root.
-	firstLevelDirs := pre[""].AsSlice()
+	firstLevelDirs := prefixToSubdirs[""].AsSlice()
 	ret := firstLevelDirs[:0]
 	for _, d := range firstLevelDirs {
-		for len(pre[d]) == 1 {
-			d = pre[d].GetArbitraryElem()
+		for len(prefixToSubdirs[d]) == 1 {
+			d = prefixToSubdirs[d].GetArbitraryElem()
 		}
 		d, _ := path.Split(d)
 		ret = append(ret, d)
--- a/pkg/matcher/matcher.go
+++ b/pkg/matcher/matcher.go
@@ -4,10 +4,12 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/stackrox/rox/pkg/set"
 	"github.com/stackrox/scanner/pkg/whiteout"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/matcher/matcher.go
```diff
@@ -39,6 +66,33 @@ func (w *allowlistMatcher) Match(fullPath string, _ os.FileInfo, _ io.ReaderAt)
 	return false, false
 }
 
+func (w *allowlistMatcher) GetCommonPrefixDirs() []string {
```
c-du, line 69: nit: w *allowlistMatcher, w should be a or m?

pkg/matcher/matcher.go
```diff
@@ -39,6 +66,33 @@ func (w *allowlistMatcher) Match(fullPath string, _ os.FileInfo, _ io.ReaderAt)
...
+		for d != "" {
+			p, _ := path.Split(strings.TrimSuffix(d, "/"))
+			s := pre[p]
+			s.Add(d)
+			pre[p] = s
+			d = p
+		}
+	}
+	dirs := pre[""].AsSlice()
+	ret := dirs[:0]
```
c-du, line 85: It looks like the common dir prefix just steps one level down from the root directory and it returns exactly one common prefix per first level dir referenced. And yes, that could be a good balance for how common the common prefix could be. Can you add more comments?
You can also rename the dirs as firstLevelDirs or dirsUnderRoot to make it more clear.

pkg/matcher/matcher.go
```diff
@@ -39,6 +66,33 @@ func (w *allowlistMatcher) Match(fullPath string, _ os.FileInfo, _ io.ReaderAt)
 	return false, false
 }
 
+func (w *allowlistMatcher) GetCommonPrefixDirs() []string {
+	return findCommonDirPrefixes(w.allowlist)
+}
+
+func findCommonDirPrefixes(prefixes []string) []string {
+	pre := make(map[string]set.StringSet)
```
c-du, line 74: How about prefixToSubdirsMap?

jvdm: Want to clarify that the method is doing a search in a trie-like structure finding paths with single-children nodes. Do you think replacing `find` with `search` would be better, or do you still think a "map" verb in the name would be clearer?

pkg/matcher/matcher.go
jvdm: :+1:

pkg/matcher/matcher.go
jvdm: :+1:

singletons/requiredfilenames/matcher.go
```diff
@@ -44,9 +50,27 @@ func SingletonMatcher() matcher.Matcher {
...
+			libDirSymlinkMatcher,
+			executableMatcher,
+		)
+	})
+	return activeVulnMatcher
+}
 
-		instance = matcher.NewOrMatcher(allMatchers...)
+// SingletonMatcher returns the singleton matcher instance to use for extracting
+// files for analyzing image container. It includes matching for O.S. features
```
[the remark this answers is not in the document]
jvdm: :+1:

pkg/matcher/matcher.go
```diff
@@ -20,13 +22,38 @@ type Matcher interface {
 	Match(fullPath string, fileInfo os.FileInfo, contents io.ReaderAt) (matches bool, extract bool)
 }
 
+// PrefixMatcher is a matcher that uses file prefixes.
+type PrefixMatcher interface {
+	Matcher
+
+	// GetCommonPrefixDirs returns a list with common directory prefixes from all
+	// file prefixes in this matcher, but excluding the root directory. The returned
```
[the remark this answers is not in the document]
jvdm: :+1:

pkg/matcher/matcher.go
c-du: I was proposing the alt name for the `pre`. I am ok with both names for the functions.

jvdm: Gotcha. Let me update it, but I will remove the `map` prefix, since golang is typed, and the decoration of the name to its type here is redundant.

c-du (approved):

PR #906, opened 2022-09-03
Fix vulnerability and package relationship for RHEL

c-du:
This is a continuous development for #895 with the following changes
1. move descriptions out of the vuln table and dedupe them
2. use hash to look for the vuln description


database/pgsql/migrations/00018_rhelv2_add_vuln_hash_vuln_package.go
```diff
@@ -0,0 +1,37 @@
+package migrations
```
c-du, line 1: Will look at updater side of changes.

PR #907, opened 2022-09-03
Update E2E testCases usage

RTann:
After doing https://github.com/stackrox/scanner/pull/905, I realized we can prevent anyone from accidentally running disabled tests by ensuring there is only one source for tests, `testCases` itself. This PR adds an `init()` which will overwrite `testCases` before anyone can use it.


e2etests/testcase_test.go
```diff
@@ -29,11 +28,19 @@ type testCase struct {
...
+			continue
+		}
+		enabledCases = append(enabledCases, tc)
+	}
+	testCases = enabledCases
+}
 
 // testCases defines all the E2E test cases.
-// testCases should not be used directly;
-// instead, users should call getEnabledTestCases().
```
jvdm, line 36: Could you please leave a comment on how this might be filtered by feature flags?

jvdm (changes requested):

jvdm (approved):
LGTM.

PR #924, opened 2022-09-08
Do not analyze language vulns unless the feature is enabled

jvdm:
Use an empty list of analyzers if language vulnerability feature is disabled.


RTann (approved):
Makes sense to me

PR #928, opened 2022-09-12
Deprecate RHELv2PackageInfo

RTann:
There has only been a single `RHELv2Package` per `RHELv2PackageInfo` since Scanner 2.19.0 (see https://github.com/stackrox/scanner/pull/456 for the related PR). Because of this, there is no real need for `RHELv2PackageInfo` any longer, so this PR deprecates it and removes all current usages of it. It is still required to exist to ensure we correctly diff older genesis dumps correctly. See the related changes in `cmd/updater/diffdumps/rhelv2_diff.go`.

Let's look at the two cases when we'd populate a `RHELv2Vulnerability`:

1. [Vuln dump generation](https://github.com/stackrox/scanner/blob/c2f1f75a976b6afc37babb22c2787053a69870f4/pkg/rhelv2/ovalutil/rpm.go#L118)
    * You can see here, for each criterion, we create a new `RHELv2PackageInfo` and add only a single `RHELv2Package`
2. [Vuln retrieval from the DB](https://github.com/stackrox/scanner/blob/2.26/database/pgsql/rhelv2_vulnerability.go#L251)
    * You can see here, we only add a single `RHELv2Package`, `pkg`, to the single `RHELv2PackageInfo` entry for the vuln.

This PR deprecates `RHELv2PackageInfo`, adds the two fields `FixedInVersion` and `ArchOperation` to `RHELv2Package`, and ensures compatibility with previous genesis dumps by converting the old `RHELv2PackageInfo` into a modern `RHELv2Package`


daynewlee (approved):
Looks good to me

PR #929, opened 2022-09-12
Skip data dump GCS upload for tagged builds

RTann:
Let's say, after releasing some Scanner version X, we add a column to a database, or change the links for all Alpine vulnerabilities, or something related to that.

Any PRs and data relied upon while developing X + 1 will assume the data is consistent with the latest format.

If, after this time, we patch X and make a new tag, then, currently, we overwrite the GCS data used in PRs with the data following the old format, which causes PRs to fail until that data is overwritten again.

Let's just stop uploading this data upon new creating a new tag. There is no need for it, and it just causes unnecessary errors.

### DIFF
--- a/scripts/ci/jobs/upload-db-dump.sh
+++ b/scripts/ci/jobs/upload-db-dump.sh
@@ -10,6 +10,9 @@ upload_db_dump() {
     if is_in_PR_context; then
         info "Skipping upload, as this is a PR"
     fi
+    if is_tagged; then
+        info "SKipping upload, as this is a tag"
+    fi
 
     info "Starting DB dump upload"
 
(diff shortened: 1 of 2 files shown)


scripts/ci/jobs/upload-db-dump.sh
```diff
@@ -10,6 +10,9 @@ upload_db_dump() {
     if is_in_PR_context; then
         info "Skipping upload, as this is a PR"
     fi
+    if is_tagged; then
```
RTann, line 13: Do I need to ensure this is not master as well, or will this suffice @gavin-stackrox ?

gavin-stackrox: It is unclear to me. But from the PR description I think you are trying to deal with inconsistency that can arise between main branch development and patch release development. If this DB artifact does not get used by patch release images or patch release CI then the check is more like `is_a_merge_to_the_main_branch()`

RTann:
Closing due to inactivity

PR #930, opened 2022-09-13
ROX-12350: Detect CVE-2022-22978

RTann:
This PR ensures we can detect CVE-2022-22978. When inspecting an example Spring application, I found the `spring-security` JARs were named more than simply `spring-security`. For example, there is `spring-security-web`. In NVD, the CPE only matches `spring-security`, so I add a manual entry to ensure we can detect it. I followed Snyk's lead https://security.snyk.io/vuln/SNYK-JAVA-ORGSPRINGFRAMEWORKSECURITY-2833359 and attributed the vuln to `spring-security-web`.

In this PR, I also try to future-proof and try to associate known Spring components to known vendors to ensure we can capture more Spring-related vulns.


cpe/attributes/java/java.go
```diff
@@ -90,6 +120,15 @@ func getPossibleVendors(origins []string) set.StringSet {
...
 		vendorSet.Add("apache")
 	}
+
+	// Add Spring-specific vendors.
+	for name := range names {
+		if knownSpringComponents.Contains(name) {
+			vendorSet.AddAll(knownSpringVendors...)
+			break
+		}
+	}
```
misberner, line 130: If performance is not a strong concern (it might be, though), you could also do

```suggestion
	if !names.Intersect(knownSpringComponents).IsEmpty() {
		vendorSet.AddAll(knownSpringVendors...)
	}
```
Or, to address the performance issue, maybe it would also be nice to add a `Intersects(other Set) bool` method to `pkg/set` (not in this PR, just as an idea).

cpe/attributes/java/java.go
```diff
@@ -90,6 +120,15 @@ func getPossibleVendors(origins []string) set.StringSet {
 	if vendorSet.Cardinality() == 0 {
 		vendorSet.Add("apache")
 	}
+
+	// Add Spring-specific vendors.
```
misberner, line 124: Does it make sense to do this only _after_ adding `"apache"` based on nothing but hope? Isn't the `knownSpringComponents` check more precise than the `"apache"` default and therefore should suppress the latter?

misberner: I guess another question would be, is it adequate to do this here even when `vendorSet` is non-empty (and wasn't populated via the apache fallback)?

misberner (changes requested):
To be honest, from reading the problem description in the PR description, the solution seems non-obvious to me.

If I understand the mechanics correctly, we now detect a vulnerability that is attributed to `spring_security` in an image that uses the `spring_security_web` package by essentially duplicating the vuln in a manual entry. Isn't this very prone to not ageing well? What if the severity or some other metadata of the vuln is updated upstream?

Rather than having manual vuln entries, wouldn't a targeted "CPE override" be the feature we want? Another possible alternative: should we have a notion of "virtual packages"? I.e., if a vuln is reported to affect `spring_security`, shouldn't we treat it as affecting `spring_security_*`? From the referenced Snyk entry, it seems quite obvious that it indeed only affects `spring_security_web`, so a CPE with that seems the more precise option. However, in the general case, this seems to be the right thing to do, because there isn't even a [spring_security package](https://mvnrepository.com/artifact/org.springframework.security) per se, just a lot of `spring-security-*` packages.

misberner:
Note: if this is time-sensitive and you share my view but deem those infeasible to implement in the timeframe, that's fine, just curious if you had considered those points.

misberner (approved):

PR #935, opened 2022-09-15
ROX-12226: Fetch RHELv2 unpatched CVE components resolution status and store them in scanner db

daynewlee:
In this pr, we are fetching resolution states of RHELv2 unpatched CVE components and store those data in the database.
In the future, we can work with the UI team and update the Stackrox dashboard to present these status of affected components in unpatched CVEs. In this way the customers can determine an appropriate course of action to mitigate these vulnerabilities.


pkg/rhelv2/ovalutil/rpm.go
```diff
@@ -223,3 +231,31 @@ func GetDefinitionType(def oval.Definition) (string, error) {
...
+		return nil
+	}
+	result := make(map[string]string)
+
+	for _, resolution := range resolutions {
+		state := resolution.State
+		components := resolution.Components
+
+		for j := 0; j < len(components); j++ {
+			component := components[j]
```
c-du, line 248: If the index is only used to get the component, you can use `for _, component := range` syntax.

c-du (approved):

PR #937, opened 2022-09-20
fix resolution state for packages with modules

RTann:
We were originally unsure what something like `maven:3.6/maven` meant when looking at a package's resolution state, so we ignored anything before the `/`. It turns out, this indicates the package (`maven`) plus its module (`maven:3.6`), so this PR ensures we correctly attribute the resolution state to the correct package/module combo.

This PR also includes various related cleaning and a fix for the md5 hash when inserting a vulnerability into the DB


database/models.go
```diff
@@ -163,10 +163,11 @@ type RHELv2PackageInfo struct {
...
 
-	Name    string `json:"name"`
-	Version string `json:"version,omitempty"`
-	Module  string `json:"module,omitempty"`
-	Arch    string `json:"arch,omitempty"`
+	Name            string `json:"name"`
+	Version         string `json:"version,omitempty"`
+	Module          string `json:"module,omitempty"`
+	Arch            string `json:"arch,omitempty"`
+	ResolutionState string `json:"resolution_state"`
```
RTann, line 170: Putting this closer to more relevant fields

database/pgsql/migrations/00019_add_vuln_state_rhel.go
```diff
@@ -6,7 +6,7 @@ func init() {
 	RegisterMigration(migrate.Migration{
 		ID: 19,
 		Up: migrate.Queries([]string{
-			`ALTER TABLE vuln_v2 ADD COLUMN IF NOT EXISTS package_resolution_state TEXT NOT NULL DEFAULT ''`,
+			`ALTER TABLE vuln_v2 ADD COLUMN IF NOT EXISTS resolution_state TEXT`,
```
RTann, line 9: no need to make this `NOT NULL`. Also, change the name to something more concise

database/pgsql/rhelv2_vulnerability.go
```diff
@@ -174,6 +174,7 @@ func md5Vuln(v *database.RHELv2Vulnerability, pkgInfoIdx, pkgIdx, cpeIdx int) []
 	b.WriteString(pkg.Module)
 	b.WriteString(pkg.Arch)
 	b.WriteString(cpe)
+	b.WriteString(pkg.ResolutionState)
```
RTann, line 177: this should be part of the md5 hash

pkg/rhelv2/ovalutil/rpm.go
```diff
@@ -247,6 +246,8 @@ func ParseUnpatchedCVEComponents(def oval.Definition) map[string]string {
 		for _, component := range components {
 			stringSlice := strings.Split(component, "/")
 			var componentName string
+			// Some components are prefixed with superfluous information followed by a /.
+			// For example, TODO.
```
RTann, line 250: @daynewlee what was the example you had for this again?

database/pgsql/migrations/00019_add_vuln_state_rhel.go
daynewlee: One of the tests failed because there's no "NOT NULL". I am guessing it's because not all package has an resolution state.

daynewlee (commented):
Looks good to me. As long as the tests pass, it should be good to go.

database/pgsql/migrations/00019_add_vuln_state_rhel.go
RTann: Yeah, there are `1777588` rows in `vuln_v2` with empty `resolution_state` :D . I don't think we need to worry about `NOT NULL` since Go would turn it into an empty string upon insert and read

pkg/rhelv2/ovalutil/rpm.go
```diff
@@ -247,6 +246,9 @@ func ParseUnpatchedCVEComponents(def oval.Definition) map[string]string {
 		for _, component := range components {
 			stringSlice := strings.Split(component, "/")
 			var componentName string
+			// Some components are prefixed with superfluous information followed by a /.
+			// For example, virt:rhel/qemu-guest-agent. This is prefixed with unnecessary information
+			// about the environment.
 			if len(stringSlice) > 1 {
```
RTann, line 252: I'm actually wondering if we should just skip these instead

RTann: This may also be the module. We should look into this more

daynewlee (approved):

PR #940, opened 2022-09-21
Update docker-entrypoint.sh to match latest version

RTann:
https://github.com/docker-library/postgres/pull/964


c-du (approved):

connorgorman:
@RTann Is this file actually necessary anymore? Like is the functionality useful or can we really slim it down to exactly what we need

RTann:
> @RTann Is this file actually necessary anymore? Like is the functionality useful or can we really slim it down to exactly what we need

I'm definitely no expert on what is 100% required vs good to have, but I think keeping this file as-is is good. If I started deleting stuff, I don't really know what effects it may have. I'd rather just keep it as-is

PR #941, opened 2022-09-21
ROX-12556: Always initialize from scratch

RTann:
Sometimes people choose to restart the nodes, which keeps the volumes but causes pods to restart. When this happens, scanner-db initialization does not work properly, as it expects to start from scratch every time. Let's enforce starting from scratch every time by deleting any data which may already exist prior to startup.


connorgorman (changes requested):
I would suggest a different approach which would be to `rm -rf` the database path and let it start from scratch. Wdyt? I worry that if we're initializing and get turned off then the database would be in a half state and it would be better to just be completely fresh

RTann:
> I would suggest a different approach which would be to `rm -rf` the database path and let it start from scratch. Wdyt? I worry that if we're initializing and get turned off then the database would be in a half state and it would be better to just be completely fresh

Good idea. We usually assume we start from scratch anyway, so we should now enforce it for correctness

connorgorman (approved):

PR #947, opened 2022-09-27
Update stackrox/stackrox dependency

RTann:
Our dependency was a bit dated, which, from a functionality standpoint, is ok, but this will also resolve vulnerabilities which have since been patched in dependencies.


RTann:
Verified if we were to update the version of scanner in stackrox's go.mod there will be no dependency issues.

misberner (commented):
In stackrox/stackrox, I once added an `exclude` directive to break the cycle of dependencies early, just to give `go mod tidy` less work. It's possible that this is no longer required in a newer go version - can you check whether adding `exclude github.com/stackrox/scanner <version imported by respective rox version>` has any material impact on the runtime of `go mod tidy` or the size of `go.sum`?

RTann:
@misberner I added `exclude github.com/stackrox/scanner v0.0.0-20220426001230-9ab6777c9581` to my go.mod just now, and I saw no change to the go.sum. I also searched the go.sum file for any reference to `scanner`, but I did not find anything, so it looks like adding the `exclude` doesn't do much

misberner (approved):
Yeah LG. The go folks added some module pruning to the resolution algorithm in 1.16 or 1.17, that's probably it.

PR #954, opened 2022-09-30
fix tag used for hourly CI runs

RTann:
I noticed "hourly" CI runs are pushing images with the nightly tag. I think this is because we set `CIRCLE_TAG` to `git tag --sort=creatordate --contains | tail -1`. I think if there have not been any merges to the `master` branch since the nightly ran, then this will return a nightly tag. For example, running this on the `master` branch right now gave me `2.26-nightly-20220930`.

This PR removes `CIRCLE_JOB` and `CIRCLE_TAG` and replaces `CIRCLE_TAG` with `NIGHTLY_TAG`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,12 @@ BASE_DIR=$(CURDIR)
 
 TAG := # make sure tag is never injectable as an env var
 
+ifdef CI
+ifneq ($(NIGHTLY_TAG),)
+TAG := $(NIGHTLY_TAG)
+endif
+endif
+
 ifeq ($(TAG),)
 TAG=$(shell git describe --tags --abbrev=10 --dirty --long)
 endif
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,6 @@ BASE_DIR=$(CURDIR)
 
 TAG := # make sure tag is never injectable as an env var
 
-ifdef CI
-ifneq ($(CIRCLE_TAG),)
-TAG := $(CIRCLE_TAG)
-endif
-endif
-
 ifeq ($(TAG),)
 TAG=$(shell git describe --tags --abbrev=10 --dirty --long)
 endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,12 +12,6 @@ BASE_DIR=$(CURDIR)
 
 TAG := # make sure tag is never injectable as an env var
 
-ifdef CI
```
RTann, line 15: I think it's ok if we just set the `tag` to the value below, especially since this PR changes `CIRCLE_TAG` to be the output of `make tag` anyway

gavin-stackrox (approved):
LGTM if this works for you but I think the change alters the meaning of CIRCLE_TAG which might cause confusion. It will now *always* be set. In the Circle CI world this was used only with CI that was triggered by the addition of a tag. Which was releases, rcs, nightlies.

RTann:
> LGTM if this works for you but I think the change alters the meaning of CIRCLE_TAG which might cause confusion. It will now _always_ be set. In the Circle CI world this was used only with CI that was triggered by the addition of a tag. Which was releases, rcs, nightlies.

That's what I was thinking, but I noticed that `git tag --sort=creatordate --contains | tail -1` returns a nightly tag, so it's possible this has been getting set by accident already?

gavin-stackrox:
> > LGTM if this works for you but I think the change alters the meaning of CIRCLE_TAG which might cause confusion. It will now _always_ be set. In the Circle CI world this was used only with CI that was triggered by the addition of a tag. Which was releases, rcs, nightlies.
> 
> That's what I was thinking, but I noticed that `git tag --sort=creatordate --contains | tail -1` returns a nightly tag, so it's possible this has been getting set by accident already?

I think with stackrox/stackrox using --contains works ok because there is no automation that runs on a cron against the main branch.

RTann:
@gavin-stackrox I updated it to set a new `NIGHTLY_TAG` and I removed `CIRCLE_TAG` since it's not really used (especially not as widely used as it is in the stackrox repo). Let me know if this makes more sense now, thanks

gavin-stackrox (approved):
Makes sense to me.

PR #957, opened 2022-10-04
Use multistage builds for base and slim images

janisz:


janisz:
It looks like multi stage builds are not supported by prow

RTann:
Yeah we originally had just a single Dockerfile for both Scanner/Scanner-slim and ScannerDB/ScannerDB-slim, but we had to separate them for the CI migration

PR #977, opened 2022-10-14
Bump hashstructure to v2

RTann:


RTann:
e2etest failure can be ignored. Fixed in https://github.com/stackrox/scanner/pull/982

daynewlee (approved):

PR #979, opened 2022-10-17
CI: gate "upload-db-dump" for tag

RTann:
This step should not run upon tag creation.


RTann:
e2etest failure can be ignored. Fixed in https://github.com/stackrox/scanner/pull/982

PR #980, opened 2022-10-17
CI: Cleanup dangling processes

RTann:
See https://github.com/stackrox/stackrox/pull/3440 for equivalent PR in the stackrox repo


scripts/ci/lib.sh
```diff
@@ -42,6 +42,51 @@ ci_exit_trap() {
...
+
+    handle_dangling_processes
+}
+
+# handle_dangling_processes() - The OpenShift CI ci-operator will not complete a
+# test job if there are processes remaining that were started by the job. While
+# processes _should_ be cleaned up by their creators it is common that some are
+# not, so this exists as a fail safe.
+handle_dangling_processes() {
+    if ! ( command -v ps >/dev/null ); then
```
RTann, line 54: I will deleted this, actually

RTann:
@gavin-stackrox Tests keep failing because `ps` is not installed. The issue is a lot of these steps in CI are based on the `src` image which does not have `ps`. I see `stackrox` uses [`test-bin`](https://github.com/openshift/release/blob/master/ci-operator/step-registry/stackrox/e2e/stackrox-e2e-ref.yaml#L15). Can you tell me more about this image?

gavin-stackrox:
- `src` == `build_root` + the cloned repo
- `test-bin` == `src` + whatever `test-binary-build-commands` create (IIRC)

I think you are hitting the problem that `build_root` is reconciled from `master` and not your PR. There is a painful update validation process detailed in https://github.com/stackrox/scanner/blob/master/.openshift-ci/Dockerfile.build_root. You could also go YOLO and revert if the merge CI fails horribly.

scripts/ci/lib.sh
gavin-stackrox, line 54:
```suggestion
    if ! command -v ps >/dev/null; then
```

gavin-stackrox (approved):
LGTM but the new build_root is not tested by this PR as mentioned earlier.

PR #987, opened 2022-10-24
Make gRPC service structs forward-compatible by embedding Unimplemented.. types

misberner:
gRPC code generation generates `Unimplemented...` structs that return an unimplemented status code for all methods. This struct should be embedded in all gRPC service implementations. While this is not currently enforced, future gRPC code generator versions will make this mandatory (this behavior can be disabled, but it is not recommended; see [here](https://github.com/grpc/grpc-go/issues/3669) for a lengthy discussion on this matter with the clear verdict that gRPC authors see embedding of `Unimplemented..` as something that should always have been required).

Alas, this makes service type assertions ineffective, but that seems to be the pill to swallow.

(description copied from https://github.com/stackrox/stackrox/pull/3538)

### DIFF
--- a/api/v1/imagescan/service.go
+++ b/api/v1/imagescan/service.go
@@ -38,6 +38,8 @@ func NewService(db database.Datastore, nvdCache nvdtoolscache.Cache) Service {
 }
 
 type serviceImpl struct {
+	v1.UnimplementedImageScanServiceServer
+
 	version  string
 	db       database.Datastore
 	nvdCache nvdtoolscache.Cache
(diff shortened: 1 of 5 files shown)


RTann (approved):
Seems weird we have to embed an `Unimplemented...` struct into our implementation of the interface but ok

misberner:
@RTann the E2E failures seem unrelated, amiright?

RTann:
> @RTann the E2E failures seem unrelated, amiright?

yeah, you can ignore it. This package has gotten quite a few vuln updates lately

PR #1001, opened 2022-11-02
Replace CVE-2022-3602 and CVE-2022-3786 with RHSA-2022:7288 for RHEL 9 and rescore CVE-2022-3602

RTann:


ext/vulnsrc/manual/manual.go
```diff
@@ -46,10 +46,10 @@ var Vulnerabilities = []database.Vulnerability{
 					"Vectors":             "",
 				},
 				"CVSSv3": map[string]interface{}{
-					"ExploitabilityScore": 2.8,
-					"ImpactScore":         5.9,
-					"Score":               8.8,
-					"Vectors":             "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H",
+					"ExploitabilityScore": 3.9,
+					"ImpactScore":         3.6,
+					"Score":               7.5,
```
theencee, line 51: which source are you using? NVD says 9.8 for CVSS 3.1 : https://nvd.nist.gov/vuln/detail/CVE-2022-3602

theencee: Your update is correct for the other CVE: https://nvd.nist.gov/vuln/detail/CVE-2022-3786. Am I mis-reading this?

RTann: This is from Red Hat. I noticed NVD added a score only after I requested a review 😄 . I'm going to let CI run and see what happens. I expect us to use NVD's results instead and the manual entry to not be used anymore, but we'll see. I'll remove the review request and then re-request a review once I confirm we see expected results.

RTann: @theencee I looked into it, and, even though the UI has the up-to-date data, the JSON feed which we read is not updated yet. Because of this, we still need our manual entries. I updated them to reflect NVD's scoring, and Ubuntu's new description

theencee (approved):

PR #1004, opened 2022-11-04
ROX-13627: Extend GetNodeVulnerabilities API by supporting Node Inventory

vikin91:
This PR extends the `GetNodeVulnerabilities` API to accept a `Node` message that contains `NodeInventory` (see https://github.com/stackrox/stackrox/pull/3755 and https://github.com/stackrox/stackrox/pull/3757). The respective Central part of this request is still in progress and there is no PR for it yet.

This is required to support RHCOS Node scanning over the `GetNodeVulnerabilities` API.

Additionally, this PR includes:
- A new feature flag `RHCOSNodeScanning`
- Improvements in the Makefile that would make it usable for:
  - Colleagues who have spaces ` ` in their `PATH`
  - ~Colleagues who do not have `wget` but have `curl`~

## How tested

- [x] Added an e2e and run it locally with:
  - `make image`
  - `make deploy-local`
  - Open port-forward to the scanner pod 8443:8443
  - Run tests `go test -tags e2e -timeout=10s -v -run ^TestGRPCGetRHCOSNodeVulnerabilities$  github.com/stackrox/scanner/e2etests`
- [x] Tested the changes in the Makefile by doing `make clean-proto-deps` and doing the steps described in  https://github.com/stackrox/scanner#steps
- [x] Deploying with Central locally and seeing the vulnerabilities in the UI
- [x] CI


vikin91:
Current dependencies on/for this PR:
* master
  * **PR #1004** <a href="https://app.graphite.dev/github/pr/stackrox/scanner/1004" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/scanner/1004?utm_source=stack-comment).

api/v1/nodescan/service.go
```diff
@@ -270,6 +272,15 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
...
 	}
 
+	// Handle the new format of the request and scan node inventory additionally
+	if features.RHCOSNodeScanning.Enabled() && req.GetNodeInventory() != nil {
+		// TODO(ROX-12968): resolve hardcoded value uncertifiedRHEL
+		layer, err := apiV1.GetVulnerabilitiesForComponents(s.db, req.GetNodeInventory(), false)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		resp.InventoryFeatures = imagescan.ConvertFeatures(layer.Features)
```
RTann, line 282: hm maybe we should put this in a more common package so we don't mix image and node scanning libraries

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,162 @@
...
+	conn := connectToScanner(t)
+	client := v1.NewNodeScanServiceClient(conn)
+
+	cases := []struct {
+		request          *v1.GetNodeVulnerabilitiesRequest
+		responseContains *v1.GetNodeVulnerabilitiesResponse
+	}{
+		{
+			request: &v1.GetNodeVulnerabilitiesRequest{
+				OsImage:          "Ubuntu 20.04.1 LTS",
```
RTann, line 64: If you want to make it more accurate, you can do something like "Red Hat Enterprise Linux CoreOS 45.82.202008101249-0 (Ootpa)" (I copied that from some other test I found)

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,162 @@
...
+			Vector:              "CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H",
+			ExploitabilityScore: 1,
+			ImpactScore:         5.9,
+		},
+	},
+	FixedBy:  "",
+	Severity: "Moderate",
+}
+
+func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
```
RTann, line 54: does this pass with the feature flag off right now? May need to enable it for tests?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+	}
+
+	for i, c := range cases {
+		t.Run(fmt.Sprintf("case-%d", i), func(t *testing.T) {
+			c := c
+			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
+			require.NoError(t, err)
+			for _, expectedFeat := range c.responseContains.InventoryFeatures {
+				found := false
```
RTann, line 131: don't care too much since this is a test, but usually `var found bool` is preferred

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,162 @@
...
+				OsImage:          "Ubuntu 20.04.1 LTS",
+				KernelVersion:    "0.0.1",
+				KubeletVersion:   "0.0.1",
+				KubeproxyVersion: "0.0.1",
+				Runtime: &v1.GetNodeVulnerabilitiesRequest_ContainerRuntime{
+					Name:    "docker",
+					Version: "0.0.1",
+				},
+				NodeInventory: &v1.Components{
+					Namespace: "Red Hat CoreOS",
```
RTann, line 73: just curious I wonder what this value actually would be

RTann (commented):
some comments. Also be sure to rebase to make the ose-jenkins E2E test pass

Makefile
```diff
@@ -1,7 +1,7 @@
 # Store tooling in a location that does not affect the system.
 GOBIN := $(CURDIR)/.gobin
 export GOBIN
-PATH := $(GOBIN):$(PATH)
+PATH := $(GOBIN):"$(PATH)"
```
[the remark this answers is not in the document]
vikin91: My PATH contains a space: `echo $PATH
...:/Applications/Visual Studio Code.app/Contents/Resources/app/bin:...` and it caused issues when building as the command was split into 3 separate commands without the quotes.

api/v1/nodescan/service.go
```diff
@@ -236,7 +239,6 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
 	if stringutils.AtLeastOneEmpty(req.GetKernelVersion(), req.GetOsImage()) {
 		return nil, status.Error(codes.InvalidArgument, "both os image and kernel version are required")
 	}
-
```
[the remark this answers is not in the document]
vikin91: No need to remove, there must have been a debug statement near that line and I accidentally removed both. I will readd this line to make the diff smaller.

make/protogen.mk
```diff
@@ -58,7 +58,11 @@ PROTOC_FILE := $(PROTOC_DOWNLOADS_DIR)/$(PROTOC_ZIP)
 
 $(PROTOC_FILE): $(PROTOC_DOWNLOADS_DIR)
 	@echo "+ $@"
+ifeq (, $(shell which wget))
```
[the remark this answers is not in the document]
vikin91: Not sure, it works for me in when I build stackrox without any additional changes. I will look how it is done there.

api/v1/nodescan/service.go
vikin91: We could - we kind of intentionally planned to call the imagescan package in the design doc, so having it here is a consequence and it is visible (maybe that is good?).

I would rely here on your suggestion here if you have some particular pkg in mind.

e2etests/node_scan_rhcos_test.go
vikin91: I will make sure that this is being executed with the FF.

e2etests/node_scan_rhcos_test.go
vikin91: Yeah, I guess it may be not fully accurate:

```
$ cat /host/etc/os-release
NAME="Red Hat Enterprise Linux CoreOS"
...
```

I will correct it.

e2etests/node_scan_rhcos_test.go
vikin91: Yes, this is being executed in the CI. I know this because the test has failed :)

```
{Failed      node_scan_rhcos_test.go:141: 
        	Error Trace:	/go/src/github.com/stackrox/scanner/e2etests/node_scan_rhcos_test.go:141
        	Error:      	Should be true
        	Test:       	TestGRPCGetRHCOSNodeVulnerabilities/case-0
        	Messages:   	expected to find feat 'libksba:1.3.5-7.el8.x86_64' in the reply, but got none
    node_scan_rhcos_test.go:141: 
        	Error Trace:	/go/src/github.com/stackrox/scanner/e2etests/node_scan_rhcos_test.go:141
        	Error:      	Should be true
        	Test:       	TestGRPCGetRHCOSNodeVulnerabilities/case-0
        	Messages:   	expected to find feat 'tar:1.27.1.el8.x86_64' in the reply, but got none}
```

Do you know why the vulns are being found when running this e2e test locally but not on CI?

vikin91:
I will rebase as well as suggested.

vikin91:
After enabling the feature flag in code (setting default to true), the test executes properly:

(from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/stackrox_scanner/1004/pull-ci-stackrox-scanner-master-e2e-tests/1595484736086609920/artifacts/e2e-tests/e2e-tests/build-log.txt)

```
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities/case-0
Handling connection for 8443
    node_scan_rhcos_test.go:141: gotFeat.InventoryFeatures: [name:"libksba" version:"1.3.5-7.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"RHSA-2022:7089" description:"KSBA (pronounced Kasbah) is a library to make X.509 certificates as well as the CMS easily accessible by other applications.  Both specifications are building blocks of S/MIME and TLS.\n\nSecurity Fix(es):\n\n* libksba: integer overflow may lead to remote code execution (CVE-2022-3515)\n\nFor more details about the security issue(s), including the impact, a CVSS score, acknowledgments, and other related information, refer to the CVE page(s) listed in the References section." link:"https://access.redhat.com/errata/RHSA-2022:7089" metadata_v2:<published_date_time:"2022-10-24T00:00Z" last_modified_date_time:"2022-10-24T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H" score:8.1 exploitability_score:2.2 impact_score:5.9 > > fixed_by:"0:1.3.5-8.el8_6" severity:"Important" > fixed_by:"1.3.5-8.el8_6"  name:"tar" version:"1.27.1.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"CVE-2005-2541" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: Tar 1.15.1 does not properly warn the user when extracting setuid or setgid files, which may allow local users or remote attackers to gain privileges. \n            STATEMENT: This CVE was assigned to what is the documented and expected behaviour of tar.  There are currently no plans to change tar behaviour to strip setuid and setgid bits when extracting archives." link:"https://access.redhat.com/security/cve/CVE-2005-2541" metadata_v2:<last_modified_date_time:"2021-06-28T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H" score:7 exploitability_score:1 impact_score:5.9 > > severity:"Moderate" > vulnerabilities:<name:"CVE-2019-9923" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: pax_decode_header in sparse.c in GNU Tar before 1.32 had a NULL pointer dereference when parsing certain archives that have malformed extended headers." link:"https://access.redhat.com/security/cve/CVE-2019-9923" metadata_v2:<last_modified_date_time:"2021-10-27T00:00Z" cvss_v3:<vector:"CVSS:3.0/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > vulnerabilities:<name:"CVE-2021-20193" description:"DOCUMENTATION: A flaw was found in the src/list.c of tar. This flaw allows an attacker who can submit a crafted input file to tar to cause uncontrolled consumption of memory. The highest threat from this vulnerability is to system availability." link:"https://access.redhat.com/security/cve/CVE-2021-20193" metadata_v2:<last_modified_date_time:"2022-04-17T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > ]
    node_scan_rhcos_test.go:141: gotFeat.InventoryFeatures: [name:"libksba" version:"1.3.5-7.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"RHSA-2022:7089" description:"KSBA (pronounced Kasbah) is a library to make X.509 certificates as well as the CMS easily accessible by other applications.  Both specifications are building blocks of S/MIME and TLS.\n\nSecurity Fix(es):\n\n* libksba: integer overflow may lead to remote code execution (CVE-2022-3515)\n\nFor more details about the security issue(s), including the impact, a CVSS score, acknowledgments, and other related information, refer to the CVE page(s) listed in the References section." link:"https://access.redhat.com/errata/RHSA-2022:7089" metadata_v2:<published_date_time:"2022-10-24T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H" score:8.1 exploitability_score:2.2 impact_score:5.9 > > fixed_by:"0:1.3.5-8.el8_6" severity:"Important" > fixed_by:"1.3.5-8.el8_6"  name:"tar" version:"1.27.1.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"CVE-2005-2541" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: Tar 1.15.1 does not properly warn the user when extracting setuid or setgid files, which may allow local users or remote attackers to gain privileges. \n            STATEMENT: This CVE was assigned to what is the documented and expected behaviour of tar.  There are currently no plans to change tar behaviour to strip setuid and setgid bits when extracting archives." link:"https://access.redhat.com/security/cve/CVE-2005-2541" metadata_v2:<last_modified_date_time:"2021-06-28T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H" score:7 exploitability_score:1 impact_score:5.9 > > severity:"Moderate" > vulnerabilities:<name:"CVE-2019-9923" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: pax_decode_header in sparse.c in GNU Tar before 1.32 had a NULL pointer dereference when parsing certain archives that have malformed extended headers." link:"https://access.redhat.com/security/cve/CVE-2019-9923" metadata_v2:<last_modified_date_time:"2021-10-27T00:00Z" cvss_v3:<vector:"CVSS:3.0/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > vulnerabilities:<name:"CVE-2021-20193" description:"DOCUMENTATION: A flaw was found in the src/list.c of tar. This flaw allows an attacker who can submit a crafted input file to tar to cause uncontrolled consumption of memory. The highest threat from this vulnerability is to system availability." link:"https://access.redhat.com/security/cve/CVE-2021-20193" metadata_v2:<last_modified_date_time:"2022-04-17T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > ]
--- PASS: TestGRPCGetRHCOSNodeVulnerabilities (4.51s)
    --- PASS: TestGRPCGetRHCOSNodeVulnerabilities/case-0 (4.51s)
```

Now I would be interested how can we enable the feature flags for e2e tests in the PRs and master, but keep it disabled (if the default is false) for releases.

e2etests/node_scan_rhcos_test.go
vikin91: Hoping to see the test being run in CI after applying https://github.com/stackrox/scanner/pull/1004/commits/229e12d3d1e2c2ec483cd6a94382a5a4ac56727d

vikin91: Apparently adding the feature-flag env to the chart did not enable it and the test was skipped. Do we have any other ideas how to make the test run on PRs and master?

RTann: Sorry, I should have mentioned adding the env var to the chart will just enable the feature in the Scanner deployment. The tests run on CI (not in the deployment), so we'd have to set the var there as well. We never really had a need to skip a test due to a feature flag before, so we do not really have that infrastructure as of yet. I'm curious if the stackrox repo has this. I know we used to with CircleCI by updating the `config.yml`. Maybe @gavin-stackrox knows and there exists a solution which, preferably, avoids OpenShift CI configuration changes. Perhaps just a bash script in the ci/scripts directory which sets environment variables at the beginning of each step?

RTann:
> After enabling the feature flag in code (setting default to true), the test executes properly:
> 
> (from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/stackrox_scanner/1004/pull-ci-stackrox-scanner-master-e2e-tests/1595484736086609920/artifacts/e2e-tests/e2e-tests/build-log.txt)
> 
> ```
> === RUN   TestGRPCGetRHCOSNodeVulnerabilities
> === RUN   TestGRPCGetRHCOSNodeVulnerabilities/case-0
> Handling connection for 8443
>     node_scan_rhcos_test.go:141: gotFeat.InventoryFeatures: [name:"libksba" version:"1.3.5-7.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"RHSA-2022:7089" description:"KSBA (pronounced Kasbah) is a library to make X.509 certificates as well as the CMS easily accessible by other applications.  Both specifications are building blocks of S/MIME and TLS.\n\nSecurity Fix(es):\n\n* libksba: integer overflow may lead to remote code execution (CVE-2022-3515)\n\nFor more details about the security issue(s), including the impact, a CVSS score, acknowledgments, and other related information, refer to the CVE page(s) listed in the References section." link:"https://access.redhat.com/errata/RHSA-2022:7089" metadata_v2:<published_date_time:"2022-10-24T00:00Z" last_modified_date_time:"2022-10-24T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H" score:8.1 exploitability_score:2.2 impact_score:5.9 > > fixed_by:"0:1.3.5-8.el8_6" severity:"Important" > fixed_by:"1.3.5-8.el8_6"  name:"tar" version:"1.27.1.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"CVE-2005-2541" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: Tar 1.15.1 does not properly warn the user when extracting setuid or setgid files, which may allow local users or remote attackers to gain privileges. \n            STATEMENT: This CVE was assigned to what is the documented and expected behaviour of tar.  There are currently no plans to change tar behaviour to strip setuid and setgid bits when extracting archives." link:"https://access.redhat.com/security/cve/CVE-2005-2541" metadata_v2:<last_modified_date_time:"2021-06-28T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H" score:7 exploitability_score:1 impact_score:5.9 > > severity:"Moderate" > vulnerabilities:<name:"CVE-2019-9923" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: pax_decode_header in sparse.c in GNU Tar before 1.32 had a NULL pointer dereference when parsing certain archives that have malformed extended headers." link:"https://access.redhat.com/security/cve/CVE-2019-9923" metadata_v2:<last_modified_date_time:"2021-10-27T00:00Z" cvss_v3:<vector:"CVSS:3.0/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > vulnerabilities:<name:"CVE-2021-20193" description:"DOCUMENTATION: A flaw was found in the src/list.c of tar. This flaw allows an attacker who can submit a crafted input file to tar to cause uncontrolled consumption of memory. The highest threat from this vulnerability is to system availability." link:"https://access.redhat.com/security/cve/CVE-2021-20193" metadata_v2:<last_modified_date_time:"2022-04-17T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > ]
>     node_scan_rhcos_test.go:141: gotFeat.InventoryFeatures: [name:"libksba" version:"1.3.5-7.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"RHSA-2022:7089" description:"KSBA (pronounced Kasbah) is a library to make X.509 certificates as well as the CMS easily accessible by other applications.  Both specifications are building blocks of S/MIME and TLS.\n\nSecurity Fix(es):\n\n* libksba: integer overflow may lead to remote code execution (CVE-2022-3515)\n\nFor more details about the security issue(s), including the impact, a CVSS score, acknowledgments, and other related information, refer to the CVE page(s) listed in the References section." link:"https://access.redhat.com/errata/RHSA-2022:7089" metadata_v2:<published_date_time:"2022-10-24T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H" score:8.1 exploitability_score:2.2 impact_score:5.9 > > fixed_by:"0:1.3.5-8.el8_6" severity:"Important" > fixed_by:"1.3.5-8.el8_6"  name:"tar" version:"1.27.1.el8.x86_64" feature_type:"rpm" vulnerabilities:<name:"CVE-2005-2541" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: Tar 1.15.1 does not properly warn the user when extracting setuid or setgid files, which may allow local users or remote attackers to gain privileges. \n            STATEMENT: This CVE was assigned to what is the documented and expected behaviour of tar.  There are currently no plans to change tar behaviour to strip setuid and setgid bits when extracting archives." link:"https://access.redhat.com/security/cve/CVE-2005-2541" metadata_v2:<last_modified_date_time:"2021-06-28T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H" score:7 exploitability_score:1 impact_score:5.9 > > severity:"Moderate" > vulnerabilities:<name:"CVE-2019-9923" description:"DOCUMENTATION: The MITRE CVE dictionary describes this issue as: pax_decode_header in sparse.c in GNU Tar before 1.32 had a NULL pointer dereference when parsing certain archives that have malformed extended headers." link:"https://access.redhat.com/security/cve/CVE-2019-9923" metadata_v2:<last_modified_date_time:"2021-10-27T00:00Z" cvss_v3:<vector:"CVSS:3.0/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > vulnerabilities:<name:"CVE-2021-20193" description:"DOCUMENTATION: A flaw was found in the src/list.c of tar. This flaw allows an attacker who can submit a crafted input file to tar to cause uncontrolled consumption of memory. The highest threat from this vulnerability is to system availability." link:"https://access.redhat.com/security/cve/CVE-2021-20193" metadata_v2:<last_modified_date_time:"2022-04-17T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L" score:3.3 exploitability_score:1.8 impact_score:1.4 > > severity:"Low" > ]
> --- PASS: TestGRPCGetRHCOSNodeVulnerabilities (4.51s)
>     --- PASS: TestGRPCGetRHCOSNodeVulnerabilities/case-0 (4.51s)
> ```
> 
> Now I would be interested how can we enable the feature flags for e2e tests in the PRs and master, but keep it disabled (if the default is false) for releases.

Scanner has never really differentiated much between release and dev before for this kind of thing. Something like this, for example, cannot even be used unless Central requests for it anyway, as the network policy dictates only Central can make requests to Scanner. Scanner does not even enforce use of the default, only, when it's a release vs dev (like the stackrox repo does) because it never really had such a need before

vikin91:
> > Now I would be interested how can we enable the feature flags for e2e tests in the PRs and master, but keep it disabled (if the default is false) for releases.
> 
> Scanner has never really differentiated much between release and dev before for this kind of thing. Something like this, for example, cannot even be used unless Central requests for it anyway, as the network policy dictates only Central can make requests to Scanner. Scanner does not even enforce use of the default, only, when it's a release vs dev (like the stackrox repo does) because it never really had such a need before

Would this suggest to use the feature flag in central and remove it from scanner?

RTann:
> > > Now I would be interested how can we enable the feature flags for e2e tests in the PRs and master, but keep it disabled (if the default is false) for releases.
> > 
> > 
> > Scanner has never really differentiated much between release and dev before for this kind of thing. Something like this, for example, cannot even be used unless Central requests for it anyway, as the network policy dictates only Central can make requests to Scanner. Scanner does not even enforce use of the default, only, when it's a release vs dev (like the stackrox repo does) because it never really had such a need before
> 
> Would this suggest to use the feature flag in central and remove it from scanner?

I'd say for now just enable the flag, and we can disable it before next release, if needed

proto/scanner/api/v1/node_scan_service.proto
```diff
@@ -15,21 +17,23 @@ message GetNodeVulnerabilitiesRequest {
...
     message ContainerRuntime {
         string name    = 1;
         string version = 2;
     }
     ContainerRuntime runtime              = 5;
 }
 
-// Next tag: 9
+// Next tag: 10
 message GetNodeVulnerabilitiesResponse {
```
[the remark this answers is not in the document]
msugakov: After checking usages, only `ScanStatus_SUCCEEDED` is used. We don't need this attribute.

pkg/features/list.go
```diff
@@ -5,4 +5,6 @@ var (
 	ContinueUnknownOS = registerFeature("Enable continuation upon detecting unknown OS", "ROX_CONTINUE_UNKNOWN_OS", true)
 	// RHEL9Scanning enables support for scanning RHEL9-based images.
 	RHEL9Scanning = registerFeature("Enable support for scanning RHEL9 images", "ROX_RHEL9_SCANNING", true)
+	// RHCOSNodeScanning enables phase 1 functions of "Full host level vulnerability scanning for RHCOS nodes" (ROX-10818)
+	RHCOSNodeScanning = registerFeature("Enable RHCOS node scanning of OS and installed packages", "ROX_RHCOS_NODE_SCANNING", false)
 )
```
msugakov, line 10: It looks like you don't need a feature flag for this given how the code is structured.

make/protogen.mk
```diff
@@ -58,7 +58,11 @@ PROTOC_FILE := $(PROTOC_DOWNLOADS_DIR)/$(PROTOC_ZIP)
...
 	$(SILENT)wget -q "https://github.com/protocolbuffers/protobuf/releases/download/v$(PROTOC_VERSION)/$(PROTOC_ZIP)" -O "$@"
+else
+	$(SILENT)curl -sSL "https://github.com/protocolbuffers/protobuf/releases/download/v$(PROTOC_VERSION)/$(PROTOC_ZIP)" -o "$@"
+endif
```
msugakov, line 65: 1. It makes little sense to have two ways of doing the same thing under a conditional. If you for any reason don't want to install `wget`, make `curl` command to be the only one here.
2. Please don't forget to add `--fail` command-line option to curl. Otherwise, it will happily download an error page.

chart/templates/deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
               fieldPath: metadata.name
         - name: ROX_SKIP_PEER_VALIDATION
           value: "true"
+        - name: ROX_RHCOS_NODE_SCANNING
+          value: "true"
         resources:
```
msugakov, line 33: I think this only would have an effect in `stackrox/scanner` CI. `stackrox/stackrox` deploys Scanner using its Helm chart. Therefore, environment value override should also be done there if you would like to see how this API functions when called from Central.

api/v1/nodescan/service.go
```diff
@@ -270,6 +273,15 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	// Handle the new format of the request and scan node inventory additionally
+	if features.RHCOSNodeScanning.Enabled() && req.GetNodeInventory() != nil {
+		// TODO(ROX-12968): resolve hardcoded value uncertifiedRHEL
```
msugakov, line 278: I don't think that ROX-12968 is appropriate task for this.

Moreover, you'll get all necessary info in `GetNodeVulnerabilitiesRequest.notes` once you add that attribute. Therefore, this TODO can be addressed as part of this same PR.

api/v1/nodescan/service.go
```diff
@@ -270,6 +273,15 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	// Handle the new format of the request and scan node inventory additionally
+	if features.RHCOSNodeScanning.Enabled() && req.GetNodeInventory() != nil {
```
msugakov, line 277: If you go the path of removing feature flag, this line will turn into

```suggestion
	if req.GetNodeInventory() != nil {
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+		},
+	}
+
+	for i, c := range cases {
+		t.Run(fmt.Sprintf("case-%d", i), func(t *testing.T) {
+			c := c
+			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
+			require.NoError(t, err)
+			for _, expectedFeat := range c.responseContains.InventoryFeatures {
```
msugakov, line 134: 1. I suggest to assert that `resp.(Inventory)Features` length equals to `c.responseContains.(Inventory)Features` length so that the output does not have more than we expect.
2. Please use Getters: `c.responseContains.GetInventoryFeatures()`.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+				KernelVersion:    "0.0.1",
+				KubeletVersion:   "0.0.1",
+				KubeproxyVersion: "0.0.1",
+				Runtime: &v1.GetNodeVulnerabilitiesRequest_ContainerRuntime{
+					Name:    "docker",
+					Version: "0.0.1",
+				},
+				NodeInventory: &v1.Components{
+					Namespace: "Namespace-Name",
+					RhelComponents: []*v1.RHELComponent{
```
msugakov, line 78: Having an additional component without vulnerabilities in this test would also be very nice.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
+			require.NoError(t, err)
+			for _, expectedFeat := range c.responseContains.InventoryFeatures {
+				var found bool
+				for _, gotFeat := range resp.InventoryFeatures {
+					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
+						found = true
+						assert.NotNil(t, gotFeat)
+						assert.NotNil(t, gotFeat.Vulnerabilities, "Expected to find vulnerabilities for %s:%s", expectedFeat.GetName(), expectedFeat.GetVersion())
+						t.Logf("gotFeat.InventoryFeatures: %+v\n", resp.InventoryFeatures)
```
msugakov, line 141: Is this dev-time logging? Looks like it will produce excessive output by dumping the same info over and over.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	for i, c := range cases {
+		t.Run(fmt.Sprintf("case-%d", i), func(t *testing.T) {
+			c := c
+			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
+			require.NoError(t, err)
+			for _, expectedFeat := range c.responseContains.InventoryFeatures {
+				var found bool
+				for _, gotFeat := range resp.InventoryFeatures {
```
msugakov, line 136: Please use getters: `resp.GetInventoryFeatures()`.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
+						found = true
+						assert.NotNil(t, gotFeat)
+						assert.NotNil(t, gotFeat.Vulnerabilities, "Expected to find vulnerabilities for %s:%s", expectedFeat.GetName(), expectedFeat.GetVersion())
```
msugakov, line 140: Please use getters: `gotFeat.GetVulnerabilities()`

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+						t.Logf("gotFeat.InventoryFeatures: %+v\n", resp.InventoryFeatures)
+						assertContainsVuln(t, gotFeat.Vulnerabilities, expectedFeat.Vulnerabilities)
```
msugakov, line 142: Please use getters: `assertContainsVuln(t, gotFeat.GetVulnerabilities(), expectedFeat.GetVulnerabilities())`

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+					}
+				}
+				assert.Truef(t, found, "expected to find feat '%s:%s' in the reply, but got none. Features in the reply: %+v", expectedFeat.GetName(), expectedFeat.GetVersion(), resp.InventoryFeatures)
```
msugakov, line 145: Please use getters: `resp.GetInventoryFeatures()`

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+			assert.Equal(t, c.responseContains.Notes, resp.Notes)
+		})
+	}
+}
+
+func assertContainsVuln(t *testing.T, foundVulns, expectedContains []*v1.Vulnerability) {
```
msugakov, line 152: 1. Please use getters inside.
2. Please assert that lengths of `foundVulns` and `expectedContains` are equal. Unless you mean that `expectedContains` can be a subset of `foundVulns`.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+	// Prune last modified time
+	for _, v := range foundVulns {
+		v.MetadataV2.LastModifiedDateTime = ""
+	}
+	if expectedContains != nil {
```
msugakov, line 157: Why is this condition needed?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+		for _, contains := range expectedContains {
+			contains.MetadataV2.LastModifiedDateTime = ""
+			if !assert.Contains(t, foundVulns, contains) {
+				fmt.Printf("Found vulns: %v\n", foundVulns)
+				fmt.Printf("Expected vuln: %v\n", contains)
+			}
```
msugakov, line 163: I think both `fmt.Printf` contents can be a function argument of some `assert.Contains` flavor. No strong feelings though.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
+//go:build e2e
+// +build e2e
+
+package e2etests
+
+import (
+	"context"
+	"fmt"
+	"github.com/stackrox/scanner/pkg/features"
+	"testing"
```
msugakov, line 10: [nit] Imports aren't ordered properly.

e2etests/node_scan_rhcos_test.go
vikin91: Yes, I meant to assert that the expected vulnerability exists in the `foundVulns`.  Maybe I rename this function to make it clear.

e2etests/node_scan_rhcos_test.go
vikin91: I finally decided to put `rhcos:4.11` here.

e2etests/node_scan_rhcos_test.go
vikin91: I decided to remove the feature-flag. The feature will be used only if the arriving request to Scanner has non-empty node inventory. This does the same job that the feature flag is doing.

proto/scanner/api/v1/node_scan_service.proto
```diff
@@ -15,21 +17,23 @@ message GetNodeVulnerabilitiesRequest {
     string             kernel_version     = 2;
     string             kubelet_version    = 3;
     string             kubeproxy_version  = 4;
+    Components         node_inventory     = 6;
```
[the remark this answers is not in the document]
vikin91: How strong is your opinion about renaming this to `components`? We call this inventory everywhere on the path from Compliance to Scanner but here.

make/protogen.mk
vikin91: I am torn here, as I know `curl` much better than `wget` and I am sure that there are people that have totally opposite opinion. Let me install `wget` then

api/v1/nodescan/service.go
vikin91: I moved it into a separate package now. Many existing pkgs caused import cycle, so I created a separate one in e693353e46875ad3f44dc2fd388dfea1edbe080e

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,166 @@
...
+						found = true
+						assert.NotNil(t, gotFeat)
+						assert.NotNil(t, gotFeat.Vulnerabilities, "Expected to find vulnerabilities for %s:%s", expectedFeat.GetName(), expectedFeat.GetVersion())
+						t.Logf("gotFeat.InventoryFeatures: %+v\n", resp.InventoryFeatures)
+						assertContainsVuln(t, gotFeat.Vulnerabilities, expectedFeat.Vulnerabilities)
+					}
+				}
+				assert.Truef(t, found, "expected to find feat '%s:%s' in the reply, but got none. Features in the reply: %+v", expectedFeat.GetName(), expectedFeat.GetVersion(), resp.InventoryFeatures)
+			}
+			assert.Equal(t, c.responseContains.Notes, resp.Notes)
```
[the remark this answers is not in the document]
vikin91: For the test cases that we have now this will be nil. In the future there maybe some different values expected than nil. I allowed myself to keep the current form of the assertion as it covers nil and non-nil cases.

proto/scanner/api/v1/node_scan_service.proto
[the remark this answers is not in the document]
msugakov: Let's keep `components` as discussed

e2etests/node_scan_rhcos_test.go
```diff
@@ -52,65 +50,77 @@ var vulnTar = &v1.Vulnerability{
...
+
 func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
 	conn := connectToScanner(t)
 	client := v1.NewNodeScanServiceClient(conn)
-	if !features.RHCOSNodeScanning.Enabled() {
-		t.Skipf("Feature RHCOS Node Scanning is disabled")
-	}
 
 	cases := []struct {
+		name             string
```
msugakov, line 115: Please switch to `map[string]struct`.

e2etests/node_scan_rhcos_test.go
```diff
@@ -121,46 +131,56 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
+					},
 				},
 			},
 		},
+		{
+			name:    "Uncertified scan is unsupported for RHCOS and returns no features",
+			request: buildRequestCase([]v1.Note{v1.Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE}),
+			responseContains: &v1.GetNodeVulnerabilitiesResponse{
+				Features: []*v1.Feature{},
+			},
```
msugakov, line 147: Suggestion:
```suggestion
			responseContains: &v1.GetNodeVulnerabilitiesResponse{
				Features: []*v1.Feature{},
				NodeNotes: nil,
			},
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -121,46 +131,56 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
-	if expectedContains != nil {
-		for _, contains := range expectedContains {
-			contains.MetadataV2.LastModifiedDateTime = ""
-			if !assert.Contains(t, foundVulns, contains) {
-				fmt.Printf("Found vulns: %v\n", foundVulns)
-				fmt.Printf("Expected vuln: %v\n", contains)
-			}
-		}
+	for _, v := range expectedToExist {
+		v.MetadataV2.LastModifiedDateTime = ""
```
msugakov, line 183: This line is redundant.
```suggestion
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -121,46 +131,56 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
 					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
 						found = true
 						assert.NotNil(t, gotFeat)
-						assert.NotNil(t, gotFeat.Vulnerabilities, "Expected to find vulnerabilities for %s:%s", expectedFeat.GetName(), expectedFeat.GetVersion())
-						t.Logf("gotFeat.InventoryFeatures: %+v\n", resp.InventoryFeatures)
-						assertContainsVuln(t, gotFeat.Vulnerabilities, expectedFeat.Vulnerabilities)
+						if len(expectedFeat.GetVulnerabilities()) > 0 {
+							assert.NotNil(t, gotFeat.GetVulnerabilities(), "Expected to find vulnerabilities for %s:%s", expectedFeat.GetName(), expectedFeat.GetVersion())
+							assertIsSubset(t, gotFeat.GetVulnerabilities(), expectedFeat.GetVulnerabilities())
+						}
```
msugakov, line 166:
```suggestion
assertIsSubset(t, gotFeat.GetVulnerabilities(), expectedFeat.GetVulnerabilities())
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -121,46 +131,56 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
 			}
-			assert.Equal(t, c.responseContains.Notes, resp.Notes)
+			assert.Equal(t, c.responseContains.GetNodeNotes(), resp.GetNodeNotes())
 		})
 	}
 }
 
-func assertContainsVuln(t *testing.T, foundVulns, expectedContains []*v1.Vulnerability) {
+// assertIsSubset asserts that every element of expectedToExist exists in gotVulns
+func assertIsSubset(t *testing.T, gotVulns, expectedToExist []*v1.Vulnerability) {
```
msugakov, line 177: Please add
```go
assert.GreaterOrEqual(t, len(gotVulns), len(expectedToExist))
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -121,46 +131,56 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
 			c := c
 			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
 			require.NoError(t, err)
-			for _, expectedFeat := range c.responseContains.InventoryFeatures {
+			assert.Len(t, resp.GetFeatures(), len(c.responseContains.GetFeatures()))
+			for _, expectedFeat := range c.responseContains.GetFeatures() {
 				var found bool
-				for _, gotFeat := range resp.InventoryFeatures {
+				for _, gotFeat := range resp.GetFeatures() {
 					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
```
msugakov, line 160: Suggestion, to reduce so many levels of indentation:

```go
for _, expectedFeat := range ... {
	var gotFeat Blah
	for _, x := range ... {
		if expectedFeat.GetName() == x.GetName() && ... {
			gotFeat = x
		}
	}
	assert.NotNil(t, gotFeat, "not found blah")
	assertIsSubset(t, gotFeat.GetVulnerabilities(), expectedFeat.GetVulnerabilities())
}
```

msugakov (approved):
I need to re-check the test after comments.

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,186 @@
...
+	FixedBy:  "0:1.3.5-8.el8_6",
+	Severity: "Important",
+}
+
+// vulnTar is an example of a non-fixable vulnerability
+var vulnTar = &v1.Vulnerability{
+	Name:        "CVE-2005-2541",
+	Description: "DOCUMENTATION: The MITRE CVE dictionary describes this issue as: Tar 1.15.1 does not properly warn the user when extracting setuid or setgid files, which may allow local users or remote attackers to gain privileges. \n            STATEMENT: This CVE was assigned to what is the documented and expected behaviour of tar.  There are currently no plans to change tar behaviour to strip setuid and setgid bits when extracting archives.",
+	Link:        "https://access.redhat.com/security/cve/CVE-2005-2541",
+	MetadataV2: &v1.Metadata{
```
msugakov, line 40: Is it ok that `PublishedDateTime` is absent here or would it only be present for a fixable vuln?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Version: "0.0.1",
+		},
+		Notes: notes,
+		Components: &v1.Components{
+			Namespace: "rhcos:4.11",
+			RhelComponents: []*v1.RHELComponent{
+				{
+					Id:        int64(1),
+					Name:      "libksba",
+					Namespace: "rhel:8",
```
msugakov, line 77: `Namespace` at the `Components` level is different from `Namespace` at the `RHELComponent` level. Which one is used by Scanner when it also sees `v1.Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE` note?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,186 @@
...
+						Name:            "tar",
+						Version:         "1.27.1.el8.x86_64",
+						Vulnerabilities: []*v1.Vulnerability{vulnTar},
+					},
+					{
+						Name:            "grep",
+						Version:         "3.1-6.el8.x86_64",
+						Vulnerabilities: []*v1.Vulnerability{},
+					},
+				},
```
msugakov, line 139:
```suggestion
				},
				NodeNotes: nil,
```

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,186 @@
...
+			resp, err := client.GetNodeVulnerabilities(context.Background(), c.request)
+			require.NoError(t, err)
+			assert.Len(t, resp.GetFeatures(), len(c.responseContains.GetFeatures()))
+			for _, expectedFeat := range c.responseContains.GetFeatures() {
+				var found bool
+				for _, gotFeat := range resp.GetFeatures() {
+					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
+						found = true
+						assert.NotNil(t, gotFeat)
+						if len(expectedFeat.GetVulnerabilities()) > 0 {
```
msugakov, line 163: One other thing to make this check more reliable would be to ignore any vulnerabilities for `gotFeat` that are newer than a hardcoded date of, say, 2022-12-07 that you can put in this test, if there is such a date attribute.

This way, you can even check equality of `gotFeat.GetVulnerabilities()` and `expectedFeat.GetVulnerabilities()` (ignoring order of elements).

Does something like creation date exist for vulnerabilities?
Judging by the name, `LastModifiedDateTime` seems more like an update date, and it may change for already known vulns causing the test to start failing when that happens.

msugakov (commented):
Went through the test

e2etests/node_scan_rhcos_test.go
vikin91: I can check the creation date, but the reason I am checking only for one vulnerability out of many is to save some space. I am not sure how many vulnerabilities exist for that version, I picked one as an example and I am asserting on that one.

All in all we are not testing here if Scanner returns the correct set of vulnerabilities, but if the API under test returns something that makes sense.

e2etests/node_scan_rhcos_test.go
vikin91: Need to check that...

vikin91: In this case, scanner is missing the `PublishedDateTime`. The vulnerability metadata looks like this:
`metadata_v2:<cvss_v3:<vector:"CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H" score:7 exploitability_score:1 impact_score:5.9>>`.
I am assuming that we cannot rely on this field being set all the time.

e2etests/node_scan_rhcos_test.go
vikin91: We seem to use `req.GetComponents()..GetNamespace()` inside of `func GetVulnerabilitiesForComponents` but it is used only to set the field inside of the `layers` struct and it does not seem to be used for scanning.

e2etests/node_scan_rhcos_test.go
msugakov: I'm fine either way.

e2etests/node_scan_rhcos_test.go
msugakov: How does scanner know that it is a RHEL-based OS and produces no features when you set `Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE` note?

e2etests/node_scan_rhcos_test.go
msugakov: I thought that you check for all vulns and the code did not tell otherwise. I think that deserves a comment somewhere around the place you define expected vulns that this is a spot-checking of vulns just to make sure some results come back.

e2etests/node_scan_rhcos_test.go
vikin91: This seems to be already addressed (I got confused trying to find the place in code).

e2etests/node_scan_rhcos_test.go
vikin91: The note `Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE` is defined at the level of `GetNodeVulnerabilitiesRequest`:

```go
	// Handle the new format of the request and scan node inventory additionally
	if req.GetComponents() != nil {
		layer, err := apiV1.GetVulnerabilitiesForComponents(s.db, req.GetComponents(), common.HasUncertifiedRHEL(req.GetNotes()))
```

The behavior of Scanner in the scope of this API seem to not rely on the value of `Components.Namespace` - I confirmed that by using several random arbitrary chosen values of namespaces and was able to obtain the same scan results.

Regarding the `Components.RHELComponent[].Namespace`: the entire `Components.RHELComponent` is passed deeper into scanner code and I have not analyzed whether the behavior depends on the namespace value, however a quick experiment with inputs:

```
RHELComponent.Namespace: "xxx:8",
RHELComponent.Namespace: "yyy:8",
RHELComponent.Namespace: "xxx:999",
RHELComponent.Namespace: "xxx:1",
```

produces the same scan results as for `RHELComponent.Namespace: "rhel:8"`. The information about namespace (which is just copied from the input to the output) returned from `apiV1.GetVulnerabilitiesForComponents` is lost in `features.ConvertFeatures`, so we cannot even add an assertion for its value in this test.

vikin91:
Asked about the e2e failure in https://srox.slack.com/archives/C033Z8KMZAM/p1670597763818699

proto/scanner/api/v1/node_scan_service.proto
```diff
@@ -37,7 +41,8 @@ message GetNodeVulnerabilitiesResponse {
     }
     KernelComponent       kernel_component           = 5;
 
-    repeated NodeNote     notes                      = 8;
+    repeated NodeNote     node_notes                 = 8;
```
RTann, line 44: how come we changed the name?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+func buildRequestCase(notes []v1.Note) *v1.GetNodeVulnerabilitiesRequest {
+	// From: https://www.redhat.com/security/data/metrics/repository-to-cpe.json
+	// "rhel-8-for-x86_64-appstream-rpms": {"cpes": ["cpe:/a:redhat:enterprise_linux:8::appstream", "cpe:/a:redhat:rhel:8.3::appstream"]},
+	// "rhel-8-for-x86_64-baseos-rpms": {"cpes": ["cpe:/o:redhat:enterprise_linux:8::baseos", "cpe:/o:redhat:rhel:8.3::baseos"]}
+	cpes := []string{
+		"cpe:/a:redhat:enterprise_linux:8::appstream", "cpe:/a:redhat:rhel:8.3::appstream",
+		"cpe:/a:redhat:enterprise_linux:8::baseos", "cpe:/a:redhat:rhel:8.3::baseos",
+	}
+	return &v1.GetNodeVulnerabilitiesRequest{
```
RTann, line 61: nit: can you comment that we use dummy values for the fields we aren't testing here (KernelVersion, Kubelet, etc)

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,187 @@
...
+			Vector:              "CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:H/I:H/A:H",
+			ExploitabilityScore: 1,
+			ImpactScore:         5.9,
+		},
+	},
+	FixedBy:  "",
+	Severity: "Moderate",
+}
+
+func buildRequestCase(notes []v1.Note) *v1.GetNodeVulnerabilitiesRequest {
```
RTann, line 53: nit: maybe just `buildRequest`?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,187 @@
...
+			responseContains: &v1.GetNodeVulnerabilitiesResponse{
+				Features:  []*v1.Feature{},
+				NodeNotes: nil,
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			c := c
```
RTann, line 158: I don't think this is needed when it's not a pointer, right?

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,187 @@
...
+				var feat *v1.Feature
+				for _, gotFeat := range resp.GetFeatures() {
+					if expectedFeat.GetName() == gotFeat.GetName() && expectedFeat.GetVersion() == gotFeat.GetVersion() {
+						feat = gotFeat
+					}
+				}
+				assert.NotNil(t, feat, "expected to find feat '%s:%s' in the reply, but got none. Features in the reply: %+v", expectedFeat.GetName(), expectedFeat.GetVersion(), resp.GetFeatures())
+				assertIsSubset(t, feat.GetVulnerabilities(), expectedFeat.GetVulnerabilities())
+			}
+			assert.Equal(t, c.responseContains.GetNodeNotes(), resp.GetNodeNotes())
```
RTann, line 172: I believe it's meant to be `assert.Equal(t, expected, actual)`

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,187 @@
...
+		})
+	}
+}
+
+// assertIsSubset asserts that every element of expectedToExist exists in gotVulns
+func assertIsSubset(t *testing.T, gotVulns, expectedToExist []*v1.Vulnerability) {
```
RTann, line 178: can we make it `assertIsSubset(t, expected, actual)`?

e2etests/node_scan_rhcos_test.go
```diff
@@ -175,12 +176,12 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
 
 // assertIsSubset asserts that every element of expectedToExist exists in gotVulns
 func assertIsSubset(t *testing.T, gotVulns, expectedToExist []*v1.Vulnerability) {
+	assert.GreaterOrEqual(t, len(gotVulns), len(expectedToExist), "Expected %d vulnerabilities to be a subset of a set that has %d elements", len(gotVulns), len(expectedToExist))
```
msugakov, line 179: The message is misleading as the "subset"/order of arguments are mixed.
Either it should be changed to
```go
	assert.GreaterOrEqual(t, len(gotVulns), len(expectedToExist), "Expected %d vulnerabilities to be a superset of a set that has %d elements", len(gotVulns), len(expectedToExist))
```
or just drop `%d`s and provide a text. I think the library will print `e1` and `e2` argument values.

e2etests/node_scan_rhcos_test.go
```diff
@@ -137,13 +140,15 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
 				},
+				NodeNotes: nil,
 			},
 		},
 		{
 			name:    "Uncertified scan is unsupported for RHCOS and returns no features",
 			request: buildRequestCase([]v1.Note{v1.Note_CERTIFIED_RHEL_SCAN_UNAVAILABLE}),
 			responseContains: &v1.GetNodeVulnerabilitiesResponse{
-				Features: []*v1.Feature{},
+				Features:  []*v1.Feature{},
```
msugakov, line 150: Given what we have discussed in comments (https://github.com/stackrox/scanner/pull/1004#discussion_r1043601763), this test does not actually validate that the scan returns **no features** (unless I'm reading the code incorrectly).

Basically, the test checks that an empty set is a subset of returned features. By definition, the empty set is a subset of any possible set.

Maybe we over-edited the thing? Either way, some assertions have to be modified now.

e2etests/node_scan_rhcos_test.go
vikin91: `func(t *testing.T)` does not accept `c` as a param, so this is needed. 
It is not the best explanation of the problem, but I usually share this link as a reference material http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#closure_for_it_vars

e2etests/node_scan_rhcos_test.go
vikin91: You are right, this needed a refresh. 
In [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39) I reworked this a bit, so that if we expect 0 features or 0 vulnerabilities, then we use `assert.Equals` on the slice lengths. 
In the other cases we use:
- `assertIsSubset` on vulns
-  each `expected` feature must be found within the `got`, and their length must match.

I know the cleanest solution would be to always assert equality, but this approach could lead to the tests failing in the future when new vulnerabilities are discovered. 

Let me know if that is enough for this case.

e2etests/node_scan_rhcos_test.go
vikin91: Improved in [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39)

e2etests/node_scan_rhcos_test.go
vikin91: I changed it to make it more clear, PTAL at [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39)

e2etests/node_scan_rhcos_test.go
vikin91: Yes, this is correct. I tried to unify this a bit and used `assert.Len` (which uses different order). PTAL at  [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39) and let me know what you think.

e2etests/node_scan_rhcos_test.go
vikin91: Done in  [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39)

e2etests/node_scan_rhcos_test.go
vikin91: Done in  [6692db4](https://github.com/stackrox/scanner/pull/1004/commits/6692db429eba7737b9c090e679f5fe5dad836a39)

e2etests/node_scan_rhcos_test.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// assertIsSubset asserts that every element of 'subset' exists in 'set'
+func assertIsSubset(t *testing.T, subset, set []*v1.Vulnerability) {
+	assert.GreaterOrEqual(t, len(set), len(subset), "Expected to find at least %d vulnerabilities", len(subset))
+	// Prune last modified time
+	for _, v := range set {
+		if v == nil {
+			continue
+		}
```
msugakov, line 198: Hm. Usually a list (slice) should not contain `nil` entries because "absent" (or "empty") values can be just not included in the list. If this list doesn't actually contain `nil`-s, than this condition can be dropped.
```suggestion
```

e2etests/node_scan_rhcos_test.go
msugakov: Well, the problem is that `grep` which currently serves as an example package without vulnerabilities may get vulnerabilities detected in the future. The comparison with `0` will start failing then.
We can either:
- do nothing now and address the issue by changing the package when tests start breaking,
- find a different package that does so little that it will not have vulnerabilities (e.g. in Debian there are meta-packages that have empty contents but bring many transitive dependencies once installed),
- try see how to use a snapshot of vulnerability data in test so that you can be confident that detected vulnerabilities don't change.

I'm fine with any of these options.

e2etests/node_scan_rhcos_test.go
msugakov: This tiny improvement can still be done.

e2etests/node_scan_rhcos_test.go
msugakov: Interesting. This means that namespace doesn't play role in rhel vulnerabilities detection. Looks like only CPEs matter. No CPEs => no vulnerabilities.

e2etests/node_scan_rhcos_test.go
vikin91: I can remove this as this is only a test code, but an API response in shape of `[]*v1.Vulnerability{nil, nil}` can theoretically happen and then the test will panic instead of failing. 

Theoretically, the goproto generated code should not allow to send `[]*v1.Vulnerability{nil, nil}` as a reply... and it does that by panicking. I hope the scanner will never return such value.

e2etests/node_scan_rhcos_test.go
vikin91: > No CPEs => no vulnerabilities

Yes, this is what I observed experimentally.

vikin91:
I wanted to answer to this comment, but cannot find a way to do so :/

> Well, the problem is that grep which currently serves as an example package without vulnerabilities may get vulnerabilities detected in the future. The comparison with 0 will start failing then.
We can either:
>- do nothing now and address the issue by changing the package when tests start breaking,
>- find a different package that does so little that it will not have vulnerabilities (e.g. in Debian there are meta-packages that have empty contents but bring many transitive dependencies once installed),
> -  see how to use a snapshot of vulnerability data in test so that you can be confident that detected vulnerabilities don't change.
> I'm fine with any of these options.

I acknowledge these options. While I tend to move on with the first option (keep as it is), I will add a comment stating that these tests may start failing in the future when new vulnerabilities are discovered for `grep` and the test would need to be adapted.

Regarding the different package than `grep` - I am open for suggestions. Running `rpm -qa` did not return any virtual packages as known from Debian, but I see that people ask in the Internet how to build a virtual package on their own using rpm.

e2etests/node_scan_rhcos_test.go
vikin91: Here: [d26da56](https://github.com/stackrox/scanner/pull/1004/commits/d26da561cb06ab166a9ffb23919e904972ad8b93)

vikin91:
The e2e test failure is consistently appearing but unrelated to this PR:
```
TestGRPCGetImageVulnerabilities/quay.io/rhacs-eng/qa:ose-jenkins/jenkins-2-plugins/4.10.1650890594-1.el8.noarch
 Messages:   	Expected to find CVE-2022-30945 in scan results
```

jvdm (approved):
LGTM.

msugakov:
> Regarding the different package than grep - I am open for suggestions. Running rpm -qa did not return any virtual packages as known from Debian, but I see that people ask in the Internet how to build a virtual package on their own using rpm.

The answer was couple more attempts away (10 minutes).
```
$ docker run --rm -it --entrypoint=/bin/bash registry.access.redhat.com/ubi8/ubi:8.7
$ dnf install less
$ dnf search '*' | grep noarch | sort | less
```

The ones which seemed most innocent:
```
git-core-doc.noarch
golang-docs.noarch
iso-codes.noarch
tzdata.noarch
```

e2etests/node_scan_rhcos_test.go
msugakov: Panic in test is perfectly valid outcome. Our life will be impossibly hard if we have to check slices/lists for `nil` elements.

e2etests/node_scan_rhcos_test.go
msugakov: The page fails to call out explicitly is that this matters when you start goroutines (or the code that's called starts goroutines). We don't do start goroutines here and can get by without `c := c`.

e2etests/node_scan_rhcos_test.go
```diff
@@ -136,16 +134,18 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
 						Vulnerabilities: []*v1.Vulnerability{vulnTar},
 					},
 					{
-						Name:            "grep",
-						Version:         "3.1-6.el8.x86_64",
+						Name:    "grep",
```
msugakov, line 137: See my `.noarch` suggestion https://github.com/stackrox/scanner/pull/1004#issuecomment-1373488790 . Optionally, up to you.

msugakov (approved):
Looks good to ship. Feel free to treat new comments as optional suggestions.

e2etests/node_scan_rhcos_test.go
vikin91: This is affected not only by Goroutines but also by closures. If we ever decide to run tests in parallel, then only one test case will be run and this bug will be difficult to find out. 

https://go.dev/play/p/HDF-sPHnTjd (comment and uncomment the `c :=c` line)

I know that probability of running into this issue is low, but there is no harm in keeping this line.
Hence, I would like to keep it for now, but I am open to discussing this topic when needed.

e2etests/node_scan_rhcos_test.go
vikin91: Unfortunately `golang-docs.noarch` returns a vulnerability (the same that can be found for `golang-docs:1.19.2.noarch`).


```
Error Trace:	/Users/prygiels/go/src/github.com/stackrox/scanner/e2etests/node_scan_rhcos_test.go:192
Error:      	"[name:"CVE-2022-23806" description:"DOCUMENTATION: A flaw was found in the elliptic package of the crypto library in golang (...)" metadata_v2:<last_modified_date_time:"2022-10-24T00:00Z" cvss_v3:<vector:"CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:L/A:H" score:8.2 exploitability_score:3.9 impact_score:4.2 > > severity:"Moderate" ]" should have 0 item(s), but has 1
Test:       	TestGRPCGetRHCOSNodeVulnerabilities/Selected_vulnerabilities_should_be_returned_by_the_certified_scan
Messages:   	Expected to find 0 vulnerabilities for feature 'golang-docs:.noarch'
```

vikin91: It looks like the CVE is for golang and not for the golang-docs, but this is what our scanner returns 🤷

msugakov: How about these?

```
git-core-doc.noarch
iso-codes.noarch
tzdata.noarch
```

vikin91: Following up in https://srox.slack.com/archives/C04213W844U/p1673027151155069

PR #1013, opened 2022-11-14
Minor update for Istio scanning

daynewlee:


api/v1/convert/istio.go
```diff
@@ -30,9 +30,10 @@ func IstioVulnerabilities(vStr string, istioVulns []types.Vuln) []*v1.Vulnerabil
 		}
 
 		link := stringutils.OrDefault(istioVuln.Link, "https://istio.io/latest/news/security/")
+		// Only second returned value is needed for fixed by version in response
```
RTann, line 33: Can you explain why this is? It's because we already know it is affected, right?

RTann (approved):
approving, but one comment

PR #1014, opened 2022-11-15
Support following absolute symlinks for node analysis

RTann:


pkg/fsutil/fileinfo/util.go
```diff
@@ -0,0 +1,13 @@
...
+import "io/fs"
+
+// IsFileExecutable returns true if the file is an executable regular file.
+func IsFileExecutable(fileInfo fs.FileInfo) bool {
+	return fileInfo.Mode().IsRegular() && fileInfo.Mode()&0111 != 0
+}
+
+// IsFileSymlink returns true if fileInfo represents a symlink.
+func IsFileSymlink(fileInfo fs.FileInfo) bool {
+	return fileInfo.Mode().Type() == fs.ModeSymlink
```
msugakov, line 12: I think, more accurate would be
```suggestion
	return fileInfo.Mode().Type()&fs.ModeSymlink != 0
```
because symlink bit can potentially be combined with some other flags.

msugakov (approved):
Looks legit to me

vikin91:
I managed to test it only locally on Colima, here it the failure obtained using the code from https://github.com/stackrox/stackrox/pull/3722 (Scanner `2.26.x-32-g3bd1d35930`)

https://github.com/stackrox/stackrox/pull/3722#issuecomment-1317371731

vikin91 (approved):
LGTM assuming that we handle the panic when exploring the filesystem on Docker in a separate PR.

RTann:
> LGTM assuming that we handle the panic when exploring the filesystem on Docker in a separate PR.

I'm going to merge because I think they are somewhat unrelated and can be handled in separate PRs. Plus, this is not used in prod yet

PR #1020, opened 2022-11-17
fix release tagging

RTann:
https://github.com/stackrox/scanner/pull/954 ensured hourly CI runs are not tagged with "nightly"; however, that change caused releases to be tagged improperly (the tag is the "long" version).

This PR adds a `RELEASE_TAG` which does not use the "long" version is the CI build is discovered to be for a release tag.

These changes were already used to tag `2.27.2`. Please see the CI run for this commit: https://github.com/stackrox/scanner/commit/e365ba197d3e2a0e459cc40cf6d2a00e08226701


gavin-stackrox (approved):
LGTM

PR #1026, opened 2022-12-02
ROX-12967: Fix RHCOS detection and namespace generation

Maddosaurus:
This PR fixes the bug that RHCOS is detected by the CentOS RegEx.
Additionally, it introduces an alternative rpm db location, as RHCOS keeps its db at `/usr/share/rpm` as opposed to RHEL, which keeps it at `/var/lib/rpm`.
Finally, it extends the `util` function with another namespace check, as RHCOS should also be treated as a well known certified platform.

A test was extended to ensure the correct detection and namespace format of RHCOS.
As discussed in Thursdays sync meeting with @RTann, we explicitly return the version including minor (e.g. `rhcos:4.11` as opposed to `rhel:8`).
I decided for a standalone RegEx instead of extending the RHEL RegEx, as I didn't want to change how RHEL versions are handled and still return a major.minor version for RHCOS.

Caveat: As the `range matcher.GetCommonPrefixDirs()` call leads to an extremely intensive filesystem walk, I introduced a very limited set of directories to search for now. We already have a follow-up item planned to investigate and improve the runtime behaviour of this code.

For these changes to run successfully in a Compliance container, an additional `emptyDir` mount is needed for Compliance at `/tmp`, as the code copies the RPM DB and opens the copy in /tmp

Testing: 
I verified this on a UBI8 image as well as an unpacked RHCOS filesystem (with the help of local-nodescanner, #1027 ).

- [x] UBI8 Image running local-nodescanner, scanning UBI8 (#1027)
- [x] UBI8 Image running local-nodescanner, scanning an unpacked RHCOS ISO image on filesystem
- [x] Compliance image with feature flag enabled, running this Sensor branch on an OS 4.11.17 cluster (Infra-provisioned)

### DIFF
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -124,7 +124,7 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 
 	notes := getNotes(layer.NamespaceName, uncertifiedRHEL)
 
-	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName)) {
+	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName) || namespaces.IsRHCOSNamespace(layer.NamespaceName)) {
 		for _, dbFeatureVersion := range dbLayer.Features {
 			feature := featureFromDatabaseModel(dbFeatureVersion, opts.GetUncertifiedRHEL(), depMap)
 
@@ -135,7 +135,7 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 			updateFeatureWithVulns(feature, dbFeatureVersion.AffectedBy, dbFeatureVersion.Feature.Namespace.VersionFormat)
 			layer.Features = append(layer.Features, *feature)
 		}
-		if !uncertifiedRHEL && namespaces.IsRHELNamespace(layer.NamespaceName) {
+		if !uncertifiedRHEL && (namespaces.IsRHELNamespace(layer.NamespaceName) || namespaces.IsRHCOSNamespace(layer.NamespaceName)) {
 			certified, err := addRHELv2Vulns(db, &layer)
 			if err != nil {
 				return layer, notes, err
@@ -202,7 +202,7 @@ func ComponentsFromDatabaseModel(db database.Datastore, dbLayer *database.Layer,
 		}
 	}
 
-	if !uncertifiedRHEL && namespaces.IsRHELNamespace(namespaceName) {
+	if !uncertifiedRHEL && (namespaces.IsRHELNamespace(namespaceName) || namespaces.IsRHCOSNamespace(namespaceName)) {
 		var certified bool
 		var err error
 		rhelv2PkgEnvs, certified, err = getRHELv2PkgEnvs(db, dbLayer.Name)
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -124,7 +124,7 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 
 	notes := getNotes(layer.NamespaceName, uncertifiedRHEL)
 
-	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName) || namespaces.IsRHCOSNamespace(layer.NamespaceName)) {
+	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName)) {
 		for _, dbFeatureVersion := range dbLayer.Features {
 			feature := featureFromDatabaseModel(dbFeatureVersion, opts.GetUncertifiedRHEL(), depMap)
 
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -135,7 +135,7 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 			updateFeatureWithVulns(feature, dbFeatureVersion.AffectedBy, dbFeatureVersion.Feature.Namespace.VersionFormat)
 			layer.Features = append(layer.Features, *feature)
 		}
-		if !uncertifiedRHEL && (namespaces.IsRHELNamespace(layer.NamespaceName) || namespaces.IsRHCOSNamespace(layer.NamespaceName)) {
+		if !uncertifiedRHEL && namespaces.IsRHELNamespace(layer.NamespaceName) {
 			certified, err := addRHELv2Vulns(db, &layer)
 			if err != nil {
 				return layer, notes, err
@@ -202,7 +202,7 @@ func ComponentsFromDatabaseModel(db database.Datastore, dbLayer *database.Layer,
 		}
 	}
 
-	if !uncertifiedRHEL && (namespaces.IsRHELNamespace(namespaceName) || namespaces.IsRHCOSNamespace(namespaceName)) {
+	if !uncertifiedRHEL && namespaces.IsRHCOSNamespace(namespaceName) {
 		var certified bool
 		var err error
 		rhelv2PkgEnvs, certified, err = getRHELv2PkgEnvs(db, dbLayer.Name)
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -202,7 +202,7 @@ func ComponentsFromDatabaseModel(db database.Datastore, dbLayer *database.Layer,
 		}
 	}
 
-	if !uncertifiedRHEL && namespaces.IsRHCOSNamespace(namespaceName) {
+	if !uncertifiedRHEL && namespaces.IsRHELNamespace(namespaceName) {
 		var certified bool
 		var err error
 		rhelv2PkgEnvs, certified, err = getRHELv2PkgEnvs(db, dbLayer.Name)
--- a/ext/featurens/redhatrelease/redhatrelease.go
+++ b/ext/featurens/redhatrelease/redhatrelease.go
@@ -32,7 +32,7 @@ import (
 var (
 	amazonReleaseRegexp = regexp.MustCompile(`(?P<os>Amazon) (Linux release|Linux AMI release) (?P<version>[\d]+\.[\d]+|[\d]+)`)
 	oracleReleaseRegexp = regexp.MustCompile(`(?P<os>Oracle) (Linux Server release) (?P<version>[\d]+)`)
-	centosReleaseRegexp = regexp.MustCompile(`(?P<os>[^\s]*) (Linux release|release) (?P<version>[\d]+)`)
+	centosReleaseRegexp = regexp.MustCompile(`(?P<os>CentOS) (Linux release|release) (?P<version>[\d]+)`)
 	redhatReleaseRegexp = regexp.MustCompile(`(?P<os>Red Hat Enterprise Linux) (Client release|Server release|Workstation release|release) (?P<version>[\d]+)`)
 	rockyReleaseRegexp  = regexp.MustCompile(`(?P<os>Rocky) (Linux release) (?P<version>[\d]+)`)
 
--- a/ext/featurens/redhatrelease/redhatrelease.go
+++ b/ext/featurens/redhatrelease/redhatrelease.go
@@ -34,6 +34,7 @@ var (
 	oracleReleaseRegexp = regexp.MustCompile(`(?P<os>Oracle) (Linux Server release) (?P<version>[\d]+)`)
 	centosReleaseRegexp = regexp.MustCompile(`(?P<os>CentOS) (Linux release|release) (?P<version>[\d]+)`)
 	redhatReleaseRegexp = regexp.MustCompile(`(?P<os>Red Hat Enterprise Linux) (Client release|Server release|Workstation release|release) (?P<version>[\d]+)`)
+	rhcosReleaseRegexp  = regexp.MustCompile(`(?P<os>Red Hat Enterprise Linux) (CoreOS release) (?P<version>[\d]+[\.]?[\d]*)`)
 	rockyReleaseRegexp  = regexp.MustCompile(`(?P<os>Rocky) (Linux release) (?P<version>[\d]+)`)
 
 	// RequiredFilenames defines the names of the files required to identify the RHEL-based release.
@@ -87,6 +88,15 @@ func (d detector) Detect(files analyzer.Files, opts *featurens.DetectorOptions) 
 				namespace.Name = "rhel:" + r[3]
 			}
 			return namespace
+		}
+
+		// Attempt to match CoreOS.
+		r = rhcosReleaseRegexp.FindStringSubmatch(string(f.Contents))
+		if len(r) == 4 {
+			return &database.Namespace{
+				Name:          "rhcos:" + r[3],
+				VersionFormat: rpm.ParserName,
+			}
 		}
 
 		// Attempt to match Rocky.
--- a/ext/featurens/redhatrelease/redhatrelease.go
+++ b/ext/featurens/redhatrelease/redhatrelease.go
@@ -32,7 +32,7 @@ import (
 var (
 	amazonReleaseRegexp = regexp.MustCompile(`(?P<os>Amazon) (Linux release|Linux AMI release) (?P<version>[\d]+\.[\d]+|[\d]+)`)
 	oracleReleaseRegexp = regexp.MustCompile(`(?P<os>Oracle) (Linux Server release) (?P<version>[\d]+)`)
-	centosReleaseRegexp = regexp.MustCompile(`(?P<os>CentOS) (Linux release|release) (?P<version>[\d]+)`)
+	centosReleaseRegexp = regexp.MustCompile(`(?P<os>[^\s]*) (Linux release|release) (?P<version>[\d]+)`)
 	redhatReleaseRegexp = regexp.MustCompile(`(?P<os>Red Hat Enterprise Linux) (Client release|Server release|Workstation release|release) (?P<version>[\d]+)`)
 	rhcosReleaseRegexp  = regexp.MustCompile(`(?P<os>Red Hat Enterprise Linux) (CoreOS release) (?P<version>[\d]+[\.]?[\d]*)`)
 	rockyReleaseRegexp  = regexp.MustCompile(`(?P<os>Rocky) (Linux release) (?P<version>[\d]+)`)
@@ -90,7 +90,7 @@ func (d detector) Detect(files analyzer.Files, opts *featurens.DetectorOptions) 
 			return namespace
 		}
 
-		// Attempt to match CoreOS.
+		// Attempt to match Red Hat CoreOS.
 		r = rhcosReleaseRegexp.FindStringSubmatch(string(f.Contents))
 		if len(r) == 4 {
 			return &database.Namespace{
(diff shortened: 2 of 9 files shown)


Maddosaurus:
Current dependencies on/for this PR:
* master
  * **PR #1026** <a href="https://app.graphite.dev/github/pr/stackrox/scanner/1026" target="_blank">Graphite</a>  👈
    * **PR #1027** <a href="https://app.graphite.dev/github/pr/stackrox/scanner/1027" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/stackrox/scanner/1026?utm_source=stack-comment).

ext/featurens/redhatrelease/redhatrelease.go
```diff
@@ -32,8 +32,9 @@ import (
 var (
 	amazonReleaseRegexp = regexp.MustCompile(`(?P<os>Amazon) (Linux release|Linux AMI release) (?P<version>[\d]+\.[\d]+|[\d]+)`)
 	oracleReleaseRegexp = regexp.MustCompile(`(?P<os>Oracle) (Linux Server release) (?P<version>[\d]+)`)
-	centosReleaseRegexp = regexp.MustCompile(`(?P<os>[^\s]*) (Linux release|release) (?P<version>[\d]+)`)
```
vikin91, line 35: This change limits the strings that we are matching - would be interested in opinion of @RTann on this.

pkg/wellknownnamespaces/util.go
```diff
@@ -20,7 +20,7 @@ func IsCentOSNamespace(namespace string) bool {
 // The namespace is expected to be of form `namespacename:version` but only `namespacename` is required.
 // For example: rhel:7, rhel:8, centos:8, ubuntu:14.04.
 func IsRHELNamespace(namespace string) bool {
-	return strings.HasPrefix(namespace, "rhel")
+	return (strings.HasPrefix(namespace, "rhel") || strings.HasPrefix(namespace, "rhcos"))
```
vikin91, line 23: So we are sure that `rhcos` is the correct string that we will get here? 
Could we simply return `rhel` from the regexp and skip this modification?

I am not suggesting anything, just asking as I truly don't know myself :) @RTann help :)

vikin91 (commented):
This looks really good!
Could we test it on Openshift nodes before merging?

pkg/wellknownnamespaces/util.go
Maddosaurus: IIRC, we ended our sync meeting yesterday (Thursday) with the notion that we want to identify RHCOS as `rhcos:4.11`, hence this check, to ensure we identify it as discussed and still enter the verified scanning functionality.
But I'm open to changing this should I have misunderstood that :)

Maddosaurus:
We do run into problems when executing this on a real running OS4 node.
The OS is correctly identified as `rhcos:4.11`, but no Components are discovered.
I'm tracking down the problem ATM. What I know so far:  The package DB location (`/var/lib/rpm`) is never added to the `filesMap.files` ([node.go#43](https://github.com/stackrox/scanner/blob/8b886b04ddbc24b496fe79c008c2506165d7069b/pkg/analyzer/nodes/node.go#L43)) array, so the code that reads the components from that DB is not able to find it.

ext/featurens/redhatrelease/redhatrelease.go
```diff
@@ -89,6 +90,15 @@ func (d detector) Detect(files analyzer.Files, opts *featurens.DetectorOptions)
 			return namespace
 		}
 
+		// Attempt to match CoreOS.
```
RTann, line 93: nit: Red Hat CoreOS

ext/featurens/redhatrelease/redhatrelease.go
RTann: I'm thinking let's revert this to ensure that continues to work as before

pkg/wellknownnamespaces/util.go
RTann: I'm thinking we make a separate `IsRHCOSNamespace` and then do the `||` when used by the client. What do you think?

Maddosaurus:
The updated code can now run and produce results if executed as local-nodescanner directly on the Openshift 4 node.
If run in Compliance, we run into issues:

With the additional changes and with a manually mounted `emptyDir` at `/tmp`, I am one step further.
The scan now successfully identifies the DB, but is unable to restore it. I will continue the investigation from here.

```
No certificates found in /usr/local/share/ca-certificates
No certificates found in /etc/pki/injected-ca-trust
main: 2022/12/06 17:22:28.636143 main.go:249: Info: Running StackRox Version: 3.73.x-95-g3b4b3561a9-dirty
main: 2022/12/06 17:22:28.637229 main.go:257: Info: Initialized Sensor gRPC stream connection
main: 2022/12/06 17:22:28.637350 main.go:274: Info: Node Rescan interval: 4h0m0s
main: 2022/12/06 17:22:28.637475 main.go:285: Info: Using NodeScan
time="2022-12-06T17:22:28Z" level=info msg="add files from directory" directory=etc/ root=/host
main: 2022/12/06 17:22:28.697936 main.go:243: Info: Successfully connected to Sensor at sensor.stackrox.svc:443
main: 2022/12/06 17:22:28.698723 main.go:94: Info: Starting audit log reader on node mat-0612-xmsd7-master-0.c.srox-temp-dev-test.internal in cluster d48d4377-f3c8-408d-b0cd-559073e8f28c using previously saved state: {
  "collectLogsSince": "2022-12-06T17:21:11.653779Z",
  "lastAuditId": "89753d51-e918-4bda-bf7b-5e314f77cc51"
})
time="2022-12-06T17:22:28Z" level=info msg="add files from directory" directory=usr/share/rpm root=/host
time="2022-12-06T17:22:28Z" level=info msg="add files from directory" directory=var/lib/rpm root=/host
added following files to filesMap: map[etc/os-release:0xc000a20c9c etc/redhat-release:0xc000a213d8 etc/system-release:0xc000e081dc etc/system-release-cpe:0xc000e081f0 usr/share/rpm/Packages:0xc000e085fb]
filesMap.Get full absPath: /host/etc/redhat-release
filesMap.Get full absPath: /host/etc/redhat-release
Trying known DB paths: map[Packages:{} rpmdb.sqlite:{}]
checking whether DB exists at var/lib/rpm/Packages
filesMap.Get full absPath: /host/usr/share/rpm/Packages
RHCOS DB exists!
checking whether DB exists at var/lib/rpm/rpmdb.sqlite
Created rpm db /tmp/rpm3788663974/Packages
time="2022-12-06T17:22:29Z" level=warning msg="failed to rebuild the rpm database: "
collection/nodescanv2: 2022/12/06 17:22:29.090198 nodescan.go:23: Info: Finished node inventory /host scan
collection/nodescanv2: 2022/12/06 17:22:29.090295 nodescan.go:27: Info: Components found under /host: <nil>
main: 2022/12/06 17:22:29.091558 main.go:243: Info: Successfully connected to Sensor at sensor.stackrox.svc:443
```

pkg/analyzer/nodes/node.go
```diff
@@ -89,7 +89,7 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 		files: make(map[string]*fileMetadata),
 	}
 	m := metrics.FileExtractionMetrics{}
-	for _, dir := range matcher.GetCommonPrefixDirs() {
+	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm"} { //TODO(ROX-13771): Use range matcher.GetCommonPrefixDirs() again after fixing
```
vikin91, line 92: To make the style-check happy:

```suggestion
        // TODO(ROX-13771): Use range matcher.GetCommonPrefixDirs() again after fixing
	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm"} { 
```

pkg/rpm/database.go
```diff
@@ -100,6 +103,7 @@ func DatabaseFiles() []string {
 	paths := make([]string, 0, databaseFiles.Cardinality())
 	for filename := range databaseFiles {
 		paths = append(paths, path.Join(databaseDir, filename))
+		paths = append(paths, path.Join(rhcosDbDir, filename))
```
vikin91, line 106: This can be a single `append` with two items:

```
paths = append(paths, path.Join(rhcosDbDir, filename), path.Join(databaseDir, filename))
```

vikin91 (approved):
Looks great!

How does it work without this?

> For these changes to run successfully in a Compliance container, an additional emptyDir mount is needed for Compliance at /tmp, as the code copies the RPM DB and opens the copy in /tmp

Maddosaurus:
> Looks great!
> 
> How does it work without this?

So, technically this is a requirement of this code running in the Compliance container.
If you run it, e.g. as local scanner, it will use your hosts' `/tmp/` folder, so I would consider this requirement more for stackrox and less for this scanner code

vikin91:
Could you maybe also handle the issue with [this statement](https://github.com/stackrox/scanner/blob/a92d02a279c4bdb5b85634d6a14ffd10b8c716ac/pkg/analyzer/nodes/node.go#L69-L71) in this PR? 

I would like to be aware of this error when it happens, so I would suggest either:

```go
	if err != nil {
		return nil, err
	}
```

or 

```go
	if err != nil {
		log.Infof("Ignoring filesystem analysis error: %v", err)
		return nil, nil
	}
```

I think it is a bug and the first suggestion should be used. If you want to move more carefully, I would be okay with adding the log message only.

pkg/rpm/database.go
```diff
@@ -41,6 +41,9 @@ const (
 	// databaseDir is the directory where the RPM database is expected to be in
 	// the container filesystem.
 	databaseDir = "var/lib/rpm"
+	// rhcosDbDir is an alternative to databaseDir, as RHCOS saves its RPM DB in
+	// a different path.
+	rhcosDbDir = "usr/share/rpm"
```
msugakov, line 46: 1. What about symlink story? I think it is worth mentioning it here in comments. More precisely, my question is about `var/lib/rpm` being a symlink to `usr/share/rpm`. Won't we end up with duplicate entries due to the same content getting discovered from two seemingly different paths?

2. Also, [nit], previous items were separated by a blank line.

pkg/rpm/database.go
```diff
@@ -99,7 +102,7 @@ func init() {
 func DatabaseFiles() []string {
 	paths := make([]string, 0, databaseFiles.Cardinality())
```
msugakov, line 103: [nit] The preallocated size now should be twice as large.

api/v1/models.go
```diff
@@ -124,7 +124,7 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 
 	notes := getNotes(layer.NamespaceName, uncertifiedRHEL)
 
-	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName)) {
+	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName) || namespaces.IsRHCOSNamespace(layer.NamespaceName)) {
```
msugakov, line 127: Why did you change this file? Did you see that the code in Compliance hits these lines?

msugakov (commented):
Sorry that it took me so long to provide this feedback. I need more changes/conversations to understand what happens here.

pkg/analyzer/detection/detection.go
```diff
@@ -27,7 +27,7 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
 	var featureVersions []database.FeatureVersion
 	var rhelfeatures *database.RHELv2Components
 
-	if namespace != nil && wellknownnamespaces.IsRHELNamespace(namespace.Name) {
+	if namespace != nil && (wellknownnamespaces.IsRHELNamespace(namespace.Name) || wellknownnamespaces.IsRHCOSNamespace(namespace.Name)) {
 		// This is a RHEL-based image that must be scanned in a certified manner.
 		// Use the RHELv2 scanner instead.
```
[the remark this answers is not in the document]
Maddosaurus: I have added the question if an uncertified RHCOS exists to our meeting question catalogue for today and created ROX-13906 to track and implement the changes (if needed). I am also adding a comment and ticket reference in code.

api/v1/models.go
Maddosaurus: Thank you for spotting this.
I added this in debugging the empty component list.
I have verified that package detection works without this change, so I have rolled it back.

pkg/analyzer/detection/detection.go
```diff
@@ -27,8 +27,12 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
 	var featureVersions []database.FeatureVersion
 	var rhelfeatures *database.RHELv2Components
 
-	if namespace != nil && wellknownnamespaces.IsRHELNamespace(namespace.Name) {
-		// This is a RHEL-based image that must be scanned in a certified manner.
+	// In the current state, RHCOS will always be handled as certified system,
+	// no matter what value uncertifiedRHEL is.
+	// This is in contrast to RHEL, where uncertifiedRHEL determines whether we scan for RHEL or CentOS.
+	// TODO(ROX-13906): Verify RHCOS logic
+	if namespace != nil && (wellknownnamespaces.IsRHELNamespace(namespace.Name) || wellknownnamespaces.IsRHCOSNamespace(namespace.Name)) {
```
RTann, line 34: Maybe for another time, but perhaps we can make something like `UseCertifiedWorkflow` which checks if the namespace is `rhel` or `rhcos`. Minimizes all the `this || that`

pkg/rpm/database.go
```diff
@@ -41,6 +41,12 @@ const (
 	// databaseDir is the directory where the RPM database is expected to be in
 	// the container filesystem.
 	databaseDir = "var/lib/rpm"
+
+	// rhcosDbDir is an alternative to databaseDir, as RHCOS saves its RPM DB in
+	// a different path that symlinks to var/lib/rpm.
+	// Since PR #1014, Scanner code follows symlinks , but usr/share/rpm will get filtered out
+	// as it is not a well known database dir unless defined here.
+	rhcosDbDir = "usr/share/rpm"
```
RTann, line 49: nit: `rhcosDBDir`

RTann: Another option is to do something like:

```
databaseDirs = []string{"var/lib/rpm", "usr/share/rpm"}
```

and have nested for loops through `darabaseDirs` and `databaseFiles`

RTann (approved):
LGTM with some comments to consider

pkg/analyzer/detection/detection.go
Maddosaurus: That is a good idea!
I have added a follow up ticket (ROX-14028) to track the refactoring.

pkg/rpm/database.go
Maddosaurus: I chose a separate var to improve understandabiliy, combined with the detailed code comment explaining the var.
That said, I'm open to changing it to a nested loop if that fits the rest of the codebase better.
WDYT?

api/v1/models.go
msugakov: You've reverted one change, but I still see two changed spots in `api/v1/models.go`.

pkg/rpm/database.go
```diff
@@ -41,9 +41,12 @@ const (
...
 	// the container filesystem.
 	databaseDir = "var/lib/rpm"
-	// rhcosDbDir is an alternative to databaseDir, as RHCOS saves its RPM DB in
-	// a different path.
-	rhcosDbDir = "usr/share/rpm"
+
+	// rhcosDBDir is an alternative to databaseDir, as RHCOS saves its RPM DB in
+	// a different path that symlinks to var/lib/rpm.
+	// Since PR #1014, Scanner code follows symlinks , but usr/share/rpm will get filtered out
+	// as it is not a well known database dir unless defined here.
```
msugakov, line 48: I spent about an hour trying to understand this. Unrelated thing, but I found a security vulnerability <https://issues.redhat.com/browse/ROX-14035>.

I think what's described here is not how #1014 works. Moreover, #1014 did not actually touch the code which has this effect of skipping `usr/share/rpm`.

The reason why `var/lib/rpm` symlink is not followed is because `node.filesMap` (implementation of `analyzer.Files` interface) is populated initially via `addFiles()` function which uses `filepath.WalkDir()` library function.

https://github.com/stackrox/scanner/blob/3432ef813ebc20a007d185a533bebbb949d41d14/pkg/analyzer/nodes/node.go#L101-L107

In the [documentation](https://pkg.go.dev/path/filepath#WalkDir) of `filepath.WalkDir()` it is said:

> WalkDir does not follow symbolic links.

Basically, `WalkDir` will see that `var/lib/rpm` is a symlink and will not traverse it.

On one hand, `usr/share/rpm` is a viable substitute for now, but in order to move with it long-term, we need to obtain a confirmation that it will stay as a location where CoreOS team will store RPMDB in all future versions.

On the other hand, `var/lib/rpm` is a _convention_ and we'll do better job if we teach our code work with symlinks and respect conventions rather than add more and more special cases.
Unfortunately, dealing with symlinks is difficult and like many other difficult things isn't implemented in Go standard library. It is also additionally complicated by the fact that our symlinks have custom root. If there's no good library solving this problem (I wonder how is it solved in the container scanning!?), I suggest to get a promise from the CoreOS team that they maintain `/usr/share/rpm` as a contract and move on with this.

I'm fine with a TODO for now.

pkg/analyzer/nodes/node.go
```diff
@@ -99,6 +99,7 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 }
 
 // addFiles searches the directory for files using the matcher and adds them to the file map.
+// TODO(ROX-14050): Improve handling of symlinks - if possible, follow instead of ignore them
```
msugakov, line 102:
```suggestion
// TODO(ROX-14050): Improve handling of symlinks - if possible, follow instead of ignoring them
```

msugakov (approved):
LGTM. I hope the testing also completes successfully.

Maddosaurus:
Thanks for the suggestions, I have accepted them.
With the latest set of changes, we are still able to produce valid results on an OS 4.11 cluster.

I have pushed a set of images with the latest changes in this PR as well as the stackrox/stackrox PR 3722.
You can get a running deployment by doing the following (e.g. with a Kubeconfig pointing to an OS 4.11 cluster):

```
export MAIN_IMAGE_REPO=quay.io/mmeiding/playground
export MAIN_IMAGE_TAG=main-g87a5cdaf05-dirty

export CENTRAL_DB_IMAGE_TAG=3.73.x-176-g5e618a632d

export ROXCTL_IMAGE_REPO=quay.io/mmeiding/playground
export ROXCTL_IMAGE_TAG=roxctl-g87a5cdaf05-dirty

./deploy/openshift/deploy.sh
```

pkg/rpm/database.go
RTann: Great catch with the potential vuln! If we are to keep our ability to read symlinks, we should definitely handle any known potential vulnerability before releasing. Luckily, we have handled a similar case before when it came to reading ZIP files (see https://security.snyk.io/research/zip-slip-vulnerability for more information about ZIP-slip): https://github.com/stackrox/scanner/blob/2.27.3/pkg/ziputil/open.go#L60

If we keep the symlink stuff, then let's definitely ensure the file is under the `root`. 

Similarly, I have been thinking about just simply looking at `/usr/share/rpm` instead, and preferring that over `/var/lib/rpm` when we know we are looking at RHCOS. Perhaps that could be a good followup after the MVP, so we don't delay this

PR #1033, opened 2022-12-16
Added fix for ROX-12979- Vulnerability scan doesn't include dependency libs

ksurabhi91:
This change will enable scanner to show vulnerabilities for source package for alpine linux.
Added a new e2e openssl vuln test for this scenario.

### DIFF
--- a/e2etests/testcase_test.go
+++ b/e2etests/testcase_test.go
@@ -20,7 +20,7 @@ type testCase struct {
 	namespace          string
 	expectedFeatures   []apiV1.Feature
 	unexpectedFeatures []apiV1.Feature
-	// This specifies that the features only need to contain at least the vulnerabilities specified
+	// This specifies that the features only need to contain at least the vulnerabilities specified.
 	onlyCheckSpecifiedVulns  bool
 	uncertifiedRHEL          bool
 	checkProvidedExecutables bool
@@ -2570,9 +2570,8 @@ var testCases = []testCase{
 						Path: "/lib/libapk.so.3.12.0",
 						RequiredFeatures: []*v1.FeatureNameVersion{
 							{Name: "apk-tools", Version: "2.12.0-r4"},
-							{Name: "libcrypto1.1", Version: "1.1.1i-r0"},
-							{Name: "libssl1.1", Version: "1.1.1i-r0"},
 							{Name: "musl", Version: "1.2.2_pre7-r0"},
+							{Name: "openssl", Version: "1.1.1i-r0"},
 							{Name: "zlib", Version: "1.2.11-r3"},
 						},
 					},
@@ -2580,8 +2579,7 @@ var testCases = []testCase{
 						Path: "/sbin/apk",
 						RequiredFeatures: []*v1.FeatureNameVersion{
 							{Name: "apk-tools", Version: "2.12.0-r4"},
-							{Name: "libcrypto1.1", Version: "1.1.1i-r0"},
-							{Name: "libssl1.1", Version: "1.1.1i-r0"},
+							{Name: "openssl", Version: "1.1.1i-r0"},
 							{Name: "musl", Version: "1.2.2_pre7-r0"},
 							{Name: "zlib", Version: "1.2.11-r3"},
 						},
@@ -2622,6 +2620,15 @@ var testCases = []testCase{
 				AddedBy: "sha256:596ba82af5aaa3e2fd9d6f955b8b94f0744a2b60710e3c243ba3e4a467f051d1",
 				FixedBy: "1.32.1-r9",
 				Executables: []*v1.Executable{
+					{
+						Path: "/usr/bin/ssl_client",
+						RequiredFeatures: []*v1.FeatureNameVersion{
+							{Name: "busybox", Version: "1.32.1-r0"},
+							{Name: "libtls-standalone", Version: "2.9.1-r1"},
+							{Name: "musl", Version: "1.2.2_pre7-r0"},
+							{Name: "openssl", Version: "1.1.1i-r0"},
+						},
+					},
 					{
 						Path: "/etc/network/if-up.d/dad",
 						RequiredFeatures: []*v1.FeatureNameVersion{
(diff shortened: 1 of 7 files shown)


e2etests/testcase_test.go
```diff
@@ -2768,6 +2768,34 @@ var testCases = []testCase{
 				AddedBy: "sha256:2408cc74d12b6cd092bb8b516ba7d5e290f485d3eb9672efc00f0583730179e8",
 				FixedBy: "1.35.0-r15",
 			},
+			{
+				Name:          "openssl",
+				NamespaceName: "alpine:v3.16",
+				VersionFormat: "apk",
+				Version:       "1.1.1a-r0",
```
RTann, line 2775: this should be `1.1.1o-r0`, right?

e2etests/testcase_test.go
```diff
@@ -2768,6 +2768,34 @@ var testCases = []testCase{
...
+				Vulnerabilities: []apiV1.Vulnerability{
+					{
+						Name:          "CVE-2018-0734",
```
RTann, line 2778: This vuln is fixed by `1.1.1a`. I'd expect us to find `CVE-2022-2097`, which is fixed by `1.1.1q-r0`

RTann (commented):
I think the branch needs to be remerged to resolve a bunch of the unrelated E2E test failures. Let's do that so we can make sure any failures are related. Also, take a look at the unit tests defined in `apk_test.go`. They seem to be failing at the moment.

Also, can we rename the PR to something like `ROX-12979: <description>`?

ext/featurefmt/apk/apk.go
```diff
@@ -58,6 +58,8 @@ func (l lister) ListFeatures(files analyzer.Files) ([]database.FeatureVersion, e
 	libToDeps := make(database.StringToStringsMap)
 	scanner := bufio.NewScanner(bytes.NewBuffer(file.Contents))
 	var dir string
+	source := ""
```
RTann, line 61: nit: `var source string`

RTann (commented):
Just the one nit, then let's get CI passing. Once that's done, then should be goo to go!

ext/featurefmt/apk/apk.go
```diff
@@ -79,6 +82,13 @@ func (l lister) ListFeatures(files analyzer.Files) ([]database.FeatureVersion, e
 					Name:    pkg.Feature.Name,
 					Version: pkg.Version,
 				}
+
+				if source != "" {
+					pkg.Feature.Name = source
+					source = ""
+				}
```
RTann, line 89: this should be done before the `key` is created

RTann (commented):
Took a look and left a few comments. Check out `apk.go` again. I left a comment there, which should help resolve some issues. Also, I'm going to request reviews from others, so there isn't a bottleneck of one reviewer

ext/featurefmt/apk/apk.go
```diff
@@ -79,6 +81,11 @@ func (l lister) ListFeatures(files analyzer.Files) ([]database.FeatureVersion, e
 					Name:    pkg.Feature.Name,
 					Version: pkg.Version,
 				}
+
+				if source != "" {
```
RTann, line 85: Doing it this way will ultimately end up with `ListFeatures` potentially returning multiple packages with the same name and version, but different `ExecutableToDependencies` and `LibraryToDependencies`. In this scenario, it is likely only one of those will be inserted into the database, so we may potentially end up with non-deterministic behavior when it comes to figuring out a package's provided executables and those executables' dependencies.

For this reason, we should update the `pkg.Feature.Name` to be `source` prior to adding it to the `pkgSet`. However, this now means we will only be using the `ExecutableToDependencies` and `LibraryToDependencies` from a single package. We should combine them via `Merge`.

You can try something like this:

```
if source != "" {
	pkg.Feature.Name = source
	source = "" // This can be moved outside the outer `if` where the other variables are reset
}

key := featurefmt.PackageKey{
	Name:    pkg.Feature.Name,
	Version: pkg.Version,
}
if oldPkg, exists := pkgSet[key]; exists {
	pkg.ExecutableToDependencies.Merge(oldPkg.ExecutableToDependencies)
	pkg.LibraryToDependencies.Merge(oldPkg.LibraryToDependencies)
}

pkgSet[key] = pkg
```

ext/featurefmt/apk/apk_test.go
```diff
@@ -64,27 +64,27 @@ func TestAPKFeatureDetection(t *testing.T) {
...
 					Version: "2.6.7-r0",
 				},
 				{
-					Feature: database.Feature{Name: "scanelf"},
+					Feature: database.Feature{Name: "pax-utils"},
 					Version: "1.1.6-r0",
 				},
 				{
-					Feature: database.Feature{Name: "musl-utils"},
+					Feature: database.Feature{Name: "musl"},
```
RTann, line 83: we already check `musl` at the top, so we can remove this

e2etests/testcase_test.go
```diff
@@ -2570,18 +2575,16 @@ var testCases = []testCase{
...
 							{Name: "zlib", Version: "1.2.11-r3"},
 						},
 					},
 					{
 						Path: "/sbin/apk",
 						RequiredFeatures: []*v1.FeatureNameVersion{
 							{Name: "apk-tools", Version: "2.12.0-r4"},
-							{Name: "libcrypto1.1", Version: "1.1.1i-r0"},
-							{Name: "libssl1.1", Version: "1.1.1i-r0"},
+							{Name: "openssl", Version: "1.1.1i-r0"},
```
RTann, line 2587: if you move this below `musl`, there won't be a need to add the sorting, but adding the sorting does prevent future confusion

e2etests/testcase_test.go
```diff
@@ -2505,6 +2509,7 @@ var testCases = []testCase{
 		registry:                 "https://registry-1.docker.io",
 		source:                   "NVD",
 		onlyCheckSpecifiedVulns:  true,
+		run:                      true,
```
RTann, line 2512: be sure to remove this prior to merging

ext/featurefmt/apk/apk_test.go
```diff
@@ -64,27 +64,27 @@ func TestAPKFeatureDetection(t *testing.T) {
 					Version: "1.2.8-r2",
 				},
 				{
-					Feature: database.Feature{Name: "libcrypto1.0"},
+					Feature: database.Feature{Name: "openssl"},
 					Version: "1.0.2h-r1",
 				},
 				{
-					Feature: database.Feature{Name: "libssl1.0"},
+					Feature: database.Feature{Name: "openssl"},
```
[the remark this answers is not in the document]
RTann: unit tests are failing because of this and the `musl` below

ext/featurefmt/apk/apk.go
```diff
@@ -74,11 +76,20 @@ func (l lister) ListFeatures(files analyzer.Files) ([]database.FeatureVersion, e
 				if len(libToDeps) != 0 {
 					pkg.LibraryToDependencies = libToDeps
 				}
+				if source != "" {
+					pkg.Feature.Name = source
+					source = "" // This can be moved outside the outer `if` where the other variables are reset
```
RTann, line 81: nit: that comment was more of a note for you. Let's either remove the comment or move `source = ""` down

ext/featurefmt/apk/apk.go
```diff
@@ -91,6 +102,7 @@ func (l lister) ListFeatures(files analyzer.Files) ([]database.FeatureVersion, e
 		case line[:2] == "P:":
 			// Start of a package definition.
 			pkg.Feature.Name = line[2:]
+			log.Println("pkgnm:", pkg.Feature.Name)
```
RTann, line 105: remove

RTann (commented):
just two minor thing then good to go!

RTann (approved):

PR #1036, opened 2022-12-19
remove timeutil package

RTann:
noticed this isn't used. No need for excess bloat in our repo/Go binary :)


daynewlee (approved):
It looks fine to me but I am not sure if the e2e-test failures are related to this change.

PR #1038, opened 2022-12-21
fix apple silicon local scanner build

dcaravel:
Fixes build (`make image`) errors on Apple Silicon Macs.

The hardcoded `BUILD_IMAGE` for `arm64` in `Makefile` has go version `1.17.2` which fails to build the latest scanner binary (most errors related to introduction of generics). A newer version of the `apollo-ci-scanner-build-X-arm64` image does not exist in the `sandbox` repo - it's unclear how the `sandbox` repo is updated

Removing the hardcode and letting `make build` use the `amd64` build image succeeds when combined with fixing a bug in the `IMAGE_BUILD_FLAGS` environment flags. This is a temporary fix as it relies on `qemu` emulation during build which has [known issues](https://docs.docker.com/desktop/troubleshoot/known-issues/):

> Some images do not support the ARM64 architecture. You can add --platform linux/amd64 to run (or build) an Intel image using emulation.
> 
> However, attempts to run Intel-based containers on Apple silicon machines under emulation can crash as qemu sometimes fails to run the container. In addition, filesystem change notification APIs (inotify) do not work under qemu emulation. Even when the containers do run correctly under emulation, they will be slower and use more memory than the native equivalent.
> 
> In summary, running Intel-based containers on Arm-based machines should be regarded as “best effort” only. We recommend running arm64 containers on Apple silicon machines whenever possible, and encouraging container authors to produce arm64, or multi-arch, versions of their containers. We expect this issue to become less common over time, as more and more images are rebuilt [supporting multiple architectures](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/).

The locally built `scanner` and `scanner-db` images are built for the `arm64` architecture - as a result the `qemu` known issues should have no/minimal impact on `scanner`/`scanner-db` at runtime.


vladbologa (commented):
Unfortunately the builder images for arm64 are not built automatically by CI. So far they were built and uploaded manually to `sandbox` by @connorgorman.

I have a temporary workaround for this issue here https://github.com/stackrox/scanner/pull/1040, if you'd still prefer to use the native builder image.

We will soon have to build arm64 images for Go v1.19. I have built them before btw, I just don't know how to upload them to sandbox.

Makefile
```diff
@@ -34,7 +34,7 @@ BUILD_IMAGE := $(DEFAULT_IMAGE_REGISTRY)/apollo-ci:$(BUILD_IMAGE_VERSION)
 
 ifeq ($(shell uname -ms),Darwin arm64)
 	# TODO(ROX-12064) build these images in the CI pipeline
-	BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-scanner-build-0.3.44-arm64
+	# BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-scanner-build-0.3.44-arm64
```
vladbologa, line 37:
```suggestion
	# Uncomment the line below to enable native arm64 builds
	# BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-stackrox-build-0.3.49-arm64
```

You could leave this here instead of just commenting out the old image, otherwise fine by me.

vladbologa (approved):

Makefile
```diff
@@ -44,7 +45,7 @@ endif
 
 LOCAL_VOLUME_ARGS := -v$(CURDIR):/src:delegated -v $(GOPATH):/go:delegated
 GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
-IMAGE_BUILD_FLAGS := -e CGO_ENABLED=0,GOOS=linux,GOARCH=${GOARCH}
```
RTann, line 47: In case you were wondering why this was never caught: Scanner was (and still is) traditionally built in a amd64 linux container, so GOOS and GOARCH were set correctly anyway. Also, I think CGO_ENABLED defaults to 0 anyway

PR #1040, opened 2022-12-22
Use Stackrox builder to build Scanner for arm64 / apple silicon

vladbologa:
This is an alternative to fix Scanner builds (`make image`) errors on Apple silicon Macs (see also https://github.com/stackrox/scanner/pull/1038), that uses a native arm64 builder image.

The current builder image has Go v17 and cannot build the current source.

This PR changes the builder image to use the current Stackrox builder, which has Go 18.4. It should be acceptable to use this image (the main difference between the Scanner and the Stackrox builder images is that the Stackrox one also contains the RocksDB binaries and its dependencies).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ BUILD_IMAGE := $(DEFAULT_IMAGE_REGISTRY)/apollo-ci:$(BUILD_IMAGE_VERSION)
 
 ifeq ($(shell uname -ms),Darwin arm64)
 	# TODO(ROX-12064) build these images in the CI pipeline
-	BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-scanner-build-0.3.44-arm64
+	BUILD_IMAGE = quay.io/rhacs-eng/sandbox:apollo-ci-stackrox-build-0.3.49-arm64
 	ARCH := aarch64
 	GOARCH := arm64
 else


dcaravel (commented):
Thanks for looking into #1038  @vladbologa!

A few thoughts for consideration:

`quay.io/rhacs-org` is not accessible to outside contributors, relying instead on the `quay.io/stackrox` builder image(s) + emulation may be more inviting until multi-arch build images are made avail

I believe the `IMAGE_BUILD_FLAGS` env bug fix (from #1038) should be made regardless to avoid future confusion / debt - even though it will not have a noticable impact when using the `arm64` builder image

Leveraging the `amd64` build image now may prevent having to make future changes if/when the `scanner` image diverges or becomes multi-arch

If you still prefer to use the proposed stackrox `BUILD_IMAGE`, I'll update #1038 accordingly - either way the local builds for apple silicon are fixed which works for me! 😉

vladbologa:
> Thanks for looking into #1038 @vladbologa!
> 
> A few thoughts for consideration:
> 
> `quay.io/rhacs-org` is not accessible to outside contributors, relying instead on the `quay.io/stackrox` builder image(s) + emulation may be more inviting until multi-arch build images are made avail
> 
> I believe the `IMAGE_BUILD_FLAGS` env bug fix (from #1038) should be made regardless to avoid future confusion / debt - even though it will not have a noticable impact when using the `arm64` builder image
> 
> Leveraging the `amd64` build image now may prevent having to make future changes if/when the `scanner` image diverges or becomes multi-arch
> 
> If you still prefer to use the proposed stackrox `BUILD_IMAGE`, I'll update #1038 accordingly - either way the local builds for apple silicon are fixed which works for me! 😉

You're right, if it works properly, I guess the only advantage of the arm64 build is speed. I should've have just commented on your PR instead of making a new one anyway, so I'll just close this one.

PR #1042, opened 2023-01-05
ROX-14035: only read file if within given root

RTann:
It is possible a symlink in a node references a file outside of the "root". It's possible the symlink is a path which attempts to link to a file outside the bounds of its own "root".

It should be noted that we should, perhaps, drop symlink support in favor of determining and reading the expected paths, directly. This fix is added in case this does not happen or it happens after a release in which this function may be used.

### DIFF
--- a/pkg/analyzer/nodes/node.go
+++ b/pkg/analyzer/nodes/node.go
@@ -1,6 +1,7 @@
 package nodes
 
 import (
+	"github.com/stackrox/scanner/pkg/fsutil"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


Maddosaurus (approved):
Looks good to me!

pkg/analyzer/nodes/node.go
```diff
@@ -181,35 +182,46 @@ func extractFile(path string, entry fs.DirEntry, pathMatcher matcher.Matcher, m
...
+		// Symlinks to relative paths are followed correctly.
+		if filepath.IsAbs(linkDest) {
+			absPath = filepath.Join(n.root, filepath.FromSlash(linkDest))
 		}
 	}
-	return analyzer.FileData{}, false
+
+	// Protect against potentially reading from a path outside root.
+	// This is possible when reading a symlinked file.
+	if !fsutil.Within(n.root, absPath) {
```
RTann, line 215: @jvdm just want to confirm we don't need to set `n.readError` here

jvdm: Do you have a corner-case scenario in mind? Otherwise, any symlink pointing to a file above the node's root is invalid by definition and should be ignored the same way as any other invalid symlink. So I don't see why we should set `n.readError`.

In the future, if we consider scanning sub-trees of the node filesystem instead of the whole node, any symlinks pointing to parents will be ignored in the scan. Suppose the file is relevant to the scan. In that case, customers might complain the node scanning is incorrect without evidence to us engineers that the way we are mounting the node's FS is breaking access to specific symlinks that were previously accessible. In any case, I don't think that blocks your PR.

jvdm (approved):
LGTM, with some remarks in the inline comment.

PR #1044, opened 2023-01-07
update binaries used in Makefile

RTann:
In the Makefile, we define a few binaries which we download. However, when using them, we may not have necessarily used the correct binary versions (they may have been pulled from the system instead). This PR ensures we use the freshly downloaded versions of the binaries. See https://github.com/stackrox/stackrox/blob/master/Makefile to see that the stackrox repo does it this way.

This PR also updates the version of `github.com/sourcegraph/go-diff` used by `golangci-lint`. See https://github.com/stackrox/stackrox/pull/4320 for more information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,14 +116,14 @@ ifdef CI
 	@echo 'The environment indicates we are in CI; running linters in check mode.'
 	@echo 'If this fails, run `make style`.'
 	@echo "Running with no tags..."
-	golangci-lint run
+	$(GOLANGCILINT_BIN) run
 	@echo "Running with release tags..."
 	@# We use --tests=false because some unit tests don't compile with release tags,
 	@# since they use functions that we don't define in the release build. That's okay.
-	golangci-lint run --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
+	$(GOLANGCILINT_BIN) run --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
 else
-	golangci-lint run --fix
-	golangci-lint run --fix --build-tags "$(subst $(comma),$(space),$(RELEASE_GOTAGS))" --tests=false
+	$(GOLANGCILINT_BIN) run --fix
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -389,6 +389,11 @@ clean-test:
 	rm -rf test-output/
 	rm -rf junit-reports/
 
+.PHONY: clean-gobin
+clean-gobin:
+	@echo "+ $@"
+	rm -rf .gobin
```
dcaravel, line 395:
```suggestion
	rm -rf $(GOBIN)
```

Makefile
```diff
@@ -380,7 +380,7 @@ clean-helm-rendered:
 .PHONY: clean-pprof
 clean-pprof:
 	@echo "+ $@"
-	rm /tmp/pprof.zip || true
+	rm -f /tmp/pprof.zip
```
dcaravel, line 383:
```suggestion
	rm -f /tmp/pprof-out/pprof.zip
```
Needs verification, but I don't believe `/tmp/pprof.zip` is ever created.  The `scale-tests` target however does appear to create `/tmp/pprof-out/pprof.zip`

dcaravel (changes requested):

tools/linters/go.mod
```diff
@@ -169,3 +170,10 @@ require (
 	mvdan.cc/lint v0.0.0-20170908181259-adc824a0674b // indirect
 	mvdan.cc/unparam v0.0.0-20220706161116-678bad134442 // indirect
 )
+
+// TODO: Remove once golangci is updated.
```
jvdm, line 174: This is better tracked as a ticket in the backlog, for visibility so anyone can follow up.

jvdm (changes requested):

tools/linters/go.mod
RTann: I'll make one, but I didn't bother at first because we have dependabot running weekly, so I figured whoever reviews the PR which updates `github.com/golangci/golangci-lint` would notice the TODO commented above it. I should also clarify it won't be fixed until `> 1.50.1`

dcaravel (approved):

jvdm (approved):
LGTM.

PR #1045, opened 2023-01-10
Improve e2e testcase for RHCOS

vikin91:
This is a follow-up to #1004.
In the test case, we need a package that has high probability of being free from vulnerabilities because we assert on it to be recognized by scanner and have exactly 0 vulnerabilities.

In this PR, I replace the previously used `grep` by `tzdata` guesstimating that the chance of finding a vulnerability in the latter is lower than in the former.

## Tested

On CI and locally with:
- `make image deploy-local`
- Opening a port-forward to scanner:8443
- Running `go test -tags e2e -timeout=10s -count=1 -v -run ^TestGRPCGetRHCOSNodeVulnerabilities$  github.com/stackrox/scanner/e2etests`

### DIFF
--- a/e2etests/node_scan_rhcos_test.go
+++ b/e2etests/node_scan_rhcos_test.go
@@ -113,7 +113,6 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
 	cases := map[string]struct {
 		request          *v1.GetNodeVulnerabilitiesRequest
 		expectedResponse *v1.GetNodeVulnerabilitiesResponse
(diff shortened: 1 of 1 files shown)


msugakov (approved):

vikin91:
The modified test case passed in CI. All failures in the recent CI run are unrelated.

RTann:
I think another option is to add some flag to the test to indicate if you want to bother looking at the vuln data or not (or even just check for specific vulns). We have this precedent set in other parts of the e2e tests (see testcase_test.go)

PR #1046, opened 2023-01-10
e2e: update jenkins-2-plugins fixed by

RTann:
See https://access.redhat.com/errata/RHSA-2022:9098 as to why this package's fixedby is updated

### DIFF
--- a/e2etests/testcase_test.go
+++ b/e2etests/testcase_test.go
@@ -3816,7 +3816,7 @@ Applications using RegexRequestMatcher with '.' in the regular expression are po
 				VersionFormat: "rpm",
 				Version:       "4.10.1650890594-1.el8.noarch",
 				AddedBy:       "sha256:3fa3f612bdcb92746bf76be1b9c9e1c1c80de777aedaf48b7068f4a129ded3c2",
-				FixedBy:       "4.10.1667388055-1.el8",
+				FixedBy:       "4.10.1670851835-1.el8",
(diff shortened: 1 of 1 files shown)


dcaravel:
@RTann I'm not understanding the association between this update and the linked security advisory.  The CVE and RHSA mentioned in the advisory are not listed as `Vulnerabilities` for `jenkins-2-plugins` in the test case.

Is there an association I'm missing? (in other words: how did you come to the conclusion that based on the advisory this needed updating?)

jvdm:
@RTann What triggered this update? Did you see testing failing somewhere? Especially curious because I want to understand if this should be handled by @acs-scanner-primary.

dcaravel (approved):
LGTM - after our chat - this update was triggered by watching CI for failures.  

I'm still wrapping my head around how the 'fixed by' version gets updated for something thats already fixed - almost as if the vuln wasn't actually fixed previously.

RTann:
> I'm still wrapping my head around how the 'fixed by' version gets updated for something thats already fixed - almost as if the vuln wasn't actually fixed previously.

ah ok so we have two `FixedBy` fields. One associated with a vulnerability for a package and one associated with the package outside of any vulnerabilities. This is the latter. This particular `FixedBy` answers the following: "What is the minimum version to which I can update the package such that the package is not affected by any fixable vulnerabilities?" This is not the `FixedBy` for a specific vulnerability affecting the package. The only reason that would be updated would be if we had a bug or the source feed had a bug.

PR #1050, opened 2023-01-11
Refactor e2e RHCOS tests

vikin91:
This PR aims at refactoring the e2e tests for node scanning, so that the assertions (equals vs. is subset) are clearer to read.

This is a follow-up to the discussion in https://github.com/stackrox/scanner/pull/1045#issuecomment-1377950931


vikin91:
Ping @RTann and @jvdm - This is a pure refactor PR, but I don't want it to be forgotten. What do you think?

jvdm (approved):
LGTM.

PR #1053, opened 2023-01-12
ROX-13794: Add content sets to Components proto

jvdm:
## Description and rationale

Add a field to `scannerV1.Components` to carry the list of Content Sets found in a node inventory scan.

`scannerV1.Component` is used by [Compliance node inventorizer](https://github.com/stackrox/stackrox/blob/ed17f4de82a70995e7d4e17cfdc8a6d75548f0c2/compliance/collection/nodeinventorizer/nodeinventory.go#L46) to send the list of components detected in a node inventory scan. The Scanner initially used that payload to hold CPEs mapped during the detection/analysis phase. But to map CPEs, Scanner needs access to [`repository-to-cpe.json`](https://www.redhat.com/en/blog/how-accurately-match-oval-security-data-installed-rpms). Compliance doesn't have access to that file, yet.

To avoid adding one additional channel to Compliance to keep repo-to-CPE data up-to-date, we will return the Content Sets found in the node. That will allow Scanner to do the mapping when Central calls [Scanner's GetNodeVulnerabilities()](https://github.com/stackrox/scanner/blob/9e56a0b37a23edc2ca9b9244dce0b8e80828c1b5/api/v1/nodescan/service.go#L277-L283).

### Why `scannerV1.Components`

Because content sets are currently attached to a single O.S. installation or an image layer for node scanning. This is acceptable for nodes because there are no layer compositions. But for images, this might be limiting. I foresee the possibility of multiple images adding different content sets, and it will be hard to tell, com `scannerV1.Components` which ones are for each package. But hopefully, this field will be ignored by image scanning.

By using `scannerV1.Components` as part of the node inventory scan, I feel we are coupling image scanning with node scanning concerning the design of the proto message.

I am open to discussing alternatives. 

## Tests

- UTs.


vikin91 (approved):
The code look great. The change is okay as well, but let me propose an alternative to it. 

Adding the field will mainly have influence on the Scanner repo, as `scannerV1.Components` proto will be translated into a stackrox proto type when we merge #3757 - see `func protoComponentsFromScanComponents(c *nodes.Components) *storage.NodeInventory_Components` in #3757.

As the life of `scannerV1.Components` will be short within the scope of stackrox code (only few lines in the `inventorizer.go`) we might as well return the content sets as an additional parameter from `Analyze` - would that maybe help keeping things cleaner in the `component.proto`?

daynewlee (approved):

jvdm:
See follow up changes at: https://github.com/stackrox/scanner/pull/1055

jvdm:
@vikin91 Based on https://github.com/stackrox/stackrox/pull/4392 I don't think we need this anymore. I am closing. I will move the small `cpe` interface changes to another PR.

vikin91:
I have a follow-up to this change: https://github.com/stackrox/scanner/pull/1059
PTAL

PR #1054, opened 2023-01-13
ROX-12966: disable node scanning for non RHCOS nodes

daynewlee:


daynewlee:
Tested on GKE cluster:


```
time="2023-01-16T16:23:21Z" level=warning msg="Not able to start node scanning for this namespace" detected namespace="ubuntu:20.04" layer=gke-yl-01-16-relieved-en-default-pool-49b42e9d-2hbn
collection/nodeinventorizer: 2023/01/16 16:23:21.145016 nodeinventory.go:27: Info: Finished node inventory
collection/nodeinventorizer: 2023/01/16 16:23:21.145055 nodeinventory.go:29: Error: Error scanning node /host inventory: Node scanning unavailable
main: 2023/01/16 16:23:21.145141 main.go:173: Error: error running scanNode: Node scanning unavailable
collection/nodeinventorizer: 2023/01/16 16:24:20.839038 nodeinventory.go:23: Info: Started node inventory
```

daynewlee:
Tested on Openshift cluster:

```
main: 2023/01/16 18:35:43.554169 main.go:243: Info: Successfully connected to Sensor at sensor.stackrox.svc:443
collection/nodeinventorizer: 2023/01/16 18:35:49.130227 nodeinventory.go:27: Info: Finished node inventory
main: 2023/01/16 18:35:49.146348 main.go:243: Info: Successfully connected to Sensor at sensor.stackrox.svc:443
collection/nodeinventorizer: 2023/01/16 18:36:43.461517 nodeinventory.go:23: Info: Started node inventory
time="2023-01-16T18:36:43Z" level=info msg="add files from directory" directory=etc/ root=/host
time="2023-01-16T18:36:43Z" level=info msg="add files from directory" directory=usr/share/rpm root=/host
time="2023-01-16T18:36:43Z" level=info msg="add files from directory" directory=var/lib/rpm root=/host
collection/nodeinventorizer: 2023/01/16 18:36:47.538836 nodeinventory.go:27: Info: Finished node inventory

```

pkg/analyzer/detection/detection.go
```diff
@@ -21,9 +22,12 @@ const LogLayerName = "layer"
 // layers, or the root layer, use `nil`. Notice that language components are not
 // extracted by DetectComponents, but if provided they are annotated with
 // certified RHEL dependencies, and returned.
-func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
+func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool, isRHCOSRequired bool) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
```
RTann, line 25: I think it may be better to have something like `DetectComponentOpts` which has attributes `uncertifiedRHEL` and maybe a slice of supported namespaces. This way, we don't have to keep adding parameters to the function. It also makes it more explicit what the parameters mean

pkg/analyzer/detection/detection.go
```diff
@@ -15,15 +16,27 @@ import (
...
 // extracted by DetectComponents, but if provided they are annotated with
 // certified RHEL dependencies, and returned.
-func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
-	namespace := DetectNamespace(name, files, parent, uncertifiedRHEL)
-
+func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, opts DetectComponentOpts) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
+	namespace := DetectNamespace(name, files, parent, opts.UncertifiedRHEL)
+	if namespace != nil && opts.IsRHCOSRequired && !wellknownnamespaces.IsRHCOSNamespace(namespace.Name) {
+		logrus.WithFields(logrus.Fields{LogLayerName: name, "detected namespace": namespace.Name}).Warning("Not able to start node scanning for this namespace")
+		return nil, nil, nil, nil, errors.New("Node scanning unavailable")
```
vikin91, line 38: Will this error lead to setting a Note so that the user is informed that we do not support node scanning, or would that be a part of a follow-up?

pkg/analyzer/nodes/node.go
```diff
@@ -66,7 +66,8 @@ func Analyze(nodeName, rootFSdir string, uncertifiedRHEL bool) (*Components, err
 	}
 	c := &Components{}
 	c.OSNamespace, c.OSComponents, c.CertifiedRHELComponents, _, err =
-		detection.DetectComponents(nodeName, files, nil, nil, uncertifiedRHEL)
+		detection.DetectComponents(nodeName, files, nil, nil,
+			detection.DetectComponentOpts{UncertifiedRHEL: uncertifiedRHEL, IsRHCOSRequired: false})
```
vikin91, line 70: `IsRHCOSRequired: false` is hardcoded here, but we want to have ability to set it to `true` when we call `Analyze` from Compliance, right?

pkg/analyzer/detection/detection.go
daynewlee: > Will this error lead to setting a Note so that the user is informed that we do not support node scanning, or would that be a part of a follow-up?

My current plan is to log and return an error in the compliance level. If the node scanning stops, I am not sure if we will still need to generate the node inventory.

pkg/analyzer/detection/detection.go
```diff
@@ -15,15 +16,27 @@ import (
 // LogLayerName is the name of the log field holding the detection target.
 const LogLayerName = "layer"
 
+// DetectComponentOpts contains namespace information for DetectComponents
```
RTann, line 19: at the moment yes, but `opts` implies these are just options we can set to configure how we want the detection to work

pkg/analyzer/detection/detection.go
```diff
@@ -15,15 +16,27 @@ import (
...
+type DetectComponentOpts struct {
+	// UncertifiedRHEL is boolean to decide if OS is uncertified RHEL
+	UncertifiedRHEL bool
+
+	// IsRHCOSRequired: if CoreOS is required for DetectComponents
```
RTann, line 24: what happens if it's false?

pkg/analyzer/nodes/node.go
```diff
@@ -56,7 +56,7 @@ type Components struct {
 }
 
 // Analyze performs analysis of node's hosts filesystem and return the detected components.
-func Analyze(nodeName, rootFSdir string, uncertifiedRHEL bool) (*Components, error) {
+func Analyze(nodeName, rootFSdir string, opts detection.DetectComponentOpts) (*Components, error) {
```
RTann, line 59: not sure if this is overkill, but I think technically we should probably have a separate set of options for `Analysis` which currently, coincidentally matches the options in `detection.DetectComponentOpts`

vikin91: This is okay, but I personally would prefer having a separate options type (i.e. not reuse `detection.DetectComponentOpts`) as Ross suggested.

pkg/analyzer/detection/detection.go
```diff
@@ -15,15 +16,27 @@ import (
 // LogLayerName is the name of the log field holding the detection target.
 const LogLayerName = "layer"
 
+// DetectComponentOpts contains configurations how components detection works
+type DetectComponentOpts struct {
+	// UncertifiedRHEL is boolean to decide if OS is uncertified RHEL
+	UncertifiedRHEL bool
+
+	// IsRHCOSRequired: if CoreOS is required for DetectComponents, Node scanning is disabled if false
+	IsRHCOSRequired bool
```
vikin91, line 25: I would be a bit more explicit here. My suggestion:

```suggestion
	// StopEarlyForNonRHCOS: if true, the analysis will run only for namespace starting with `rhcos:`. If false, the analysis will stop after detecting the namespace.
	StopEarlyForNonRHCOS bool
```

vikin91 (approved):
This looks good, my comments are optional to address.
I would suggest to maybe get a green review from Scanner team member before merging.

I think, we should also discuss with PM the topic of setting a Note when non-RHCOS system is detected.

pkg/analyzer/detection/detection.go
Maddosaurus: Nit: I found the naming of `IsRHCOSRequired` a bit confusing as well. 
If I read this as consumer of the lib, I would ask myself "in which case would RHCOS be required?".

@vikin91's proposal seems fine to me. Alternatively I would pitch something like `RHCOSMode`  or `StrictMode`, although I'm not strongly in favor of any of these 🙂

pkg/analyzer/detection/detection.go
```diff
@@ -15,15 +16,27 @@ import (
...
 // DetectComponents detects the namespace and extracts the components present in
 // the files of a filesystem or image layer. For layers, the parent layer should
 // be specified. For filesystems, which don't have the concept of intermediate
 // layers, or the root layer, use `nil`. Notice that language components are not
 // extracted by DetectComponents, but if provided they are annotated with
 // certified RHEL dependencies, and returned.
-func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, uncertifiedRHEL bool) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
-	namespace := DetectNamespace(name, files, parent, uncertifiedRHEL)
-
+func DetectComponents(name string, files analyzer.Files, parent *database.Layer, languageComponents []*component.Component, opts DetectComponentOpts) (*database.Namespace, []database.FeatureVersion, *database.RHELv2Components, []*component.Component, error) {
```
Maddosaurus, line 34: Nit: Should we maybe update the comment above to include the info about the early exit on non-RHCOS namespaces if the scan opts are set?

Maddosaurus (approved):
Thanks for the change! Feel free to TAL at the small, optional things that came to mind while reading this change.

vikin91:
@daynewlee one more question, it would be nice to edit the PR description to include there a sentence or two about how you tested this change. We want to make sure that it works after merging ;)

vikin91:
Could we also fix style before merging? (see "style-checks" in CI)

PR #1055, opened 2023-01-13
ROX-13794: Make node analyzer return content sets, not cpes

jvdm:
## Description and rationale

Populate RHEL Content Sets in the return value of certified RHEL components detection.

This will be consumed by Compliance node scanning and eventually sent to Central Scanner through `scannerV1.Component` to be mapped to CPEs. Analysis detection used to perform the CPE mapping on detection, and that is left for backward compatibility. Still, the mapping results in `nil` gracefully if CPE mapping is not available, which is the case for node scanning.

## Tests

- UTs.


vikin91:
We also must make sure that the `GetNodeVulnerabilitiesRequest.Components` from https://github.com/stackrox/scanner/blob/master/proto/scanner/api/v1/node_scan_service.proto contains the `content_sets` instead/additionally to CPEs.

Otherwise Central will not be able to pass the content_sets to Scanner through the `GetNodeVulnerabilities` API.

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
...
-		}
+		logrus.WithFields(logrus.Fields{
+			LogLayerName:    name,
+			"feature count": len(featureVersions),
+		}).Debug("detected features")
+	}
+	return namespace, featureVersions, rhelFeatures, languageComponents, nil
+}
+
+func isCertifiedRHELNamespace(namespace *database.Namespace) bool {
```
RTann, line 60: I'm thinking maybe we put this in `wellknownnamespaces`?

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
 	namespace := DetectNamespace(name, files, parent, uncertifiedRHEL)
 
 	var featureVersions []database.FeatureVersion
-	var rhelfeatures *database.RHELv2Components
+	var rhelFeatures *database.RHELv2Components
+	var err error
 
 	// In the current state, RHCOS will always be handled as certified system.
 	// If no CPEs are found on RHCOS, a note needs to be added that informs users of it.
```
RTann, line 33: this comment may need to be adjusted since we are looking at content sets now

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
...
-		// Detect features.
 		featureVersions, err = detectFeatureVersions(name, files, namespace, parent)
 		if err != nil {
 			return nil, nil, nil, nil, err
 		}
-		if len(featureVersions) > 0 {
-			logrus.WithFields(logrus.Fields{LogLayerName: name, "feature count": len(featureVersions)}).Debug("detected features")
-		}
+		logrus.WithFields(logrus.Fields{
+			LogLayerName:    name,
```
RTann, line 53: I'm actually not familiar with this key

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
...
+	// This is a certified image that needs to be scanned differently.
+	// Use the RHELv2 scanner instead.
+	packages, contentSets, err := rpm.ListFeatures(files)
+	if err != nil {
+		return nil, nil, err
+	}
+	rhelfeatures := &database.RHELv2Components{
+		Dist:     namespace.Name,
+		Packages: packages,
+		// CPEs are mapped and returned with content sets for backward compatibility.
```
RTann, line 78: Needed for scanning images in OCR. Perhaps we can update that path, too, which would also eliminate the need to pass those over from Central -> Sensor -> Scanner in another PR, if it works out

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
...
+		CPEs:        repo2cpe.Singleton().Get(contentSets),
+		ContentSets: contentSets,
+	}
+	if err := rpm.AnnotateComponentsWithPackageManagerInfo(files, languageComponents); err != nil {
+		logrus.WithError(err).Errorf("Failed to analyze package manager info for language components: %s", name)
```
RTann, line 83: I wonder if we should make this a warning since we do ignore it?

pkg/rhelv2/rpm/rpm.go
```diff
@@ -17,42 +17,47 @@ import (
...
-	contentManifests = `root/buildinfo/content_manifests`
-
 	pkgFmt = `rpmv2`
 )
 
 // AllRHELRequiredFiles lists all the names of the files required to identify RHEL-based releases.
 var AllRHELRequiredFiles set.StringSet
 
+// contentManifests set with all known directories that might contain content manifest files.
+var contentManifestDirs set.StringSet
```
RTann, line 31:
```suggestion
var (
	// AllRHELRequiredFiles lists all the names of the files required to identify RHEL-based releases.
	var AllRHELRequiredFiles set.StringSet

	// contentManifests set with all known directories that might contain content manifest files.
	var contentManifestDirs set.StringSet
)
```

RTann: also nit: comment name should start with the variable's name

pkg/rhelv2/rpm/rpm.go
```diff
@@ -77,33 +82,26 @@ func AddToDependencyMap(filename string, fileData analyzer.FileData, execToDeps,
...
-	// be translated using mapping file provided by Red Hat's PST team.
-	contents := getContentManifestFileContents(files)
+// getContentManifestSets returns the list of content sets defined in the content
+// manifest, if found in the files. Otherwise, `nil` or empty slice is returned.
+func getContentManifestSets(files analyzer.Files) ([]string, error) {
+	var contents []byte
+	for _, prefix := range contentManifestDirs.AsSlice() {
+		for name, file := range files.GetFilesPrefix(prefix) {
+			if strings.HasSuffix(name, ".json") {
+				// Return the first one found, as there should be only one dir per layer/node.
```
RTann, line 92: nit: file not dir

Also, I'm not 100% sure we can guarantee it's 1 per node. An image can have more than one (I've seen up to three in the wild), and if a node is essentially a running container, then I think it's possible a node can have more than one.

pkg/rhelv2/rpm/rpm_test.go
```diff
@@ -45,11 +45,11 @@ func Test_listFeatures(t *testing.T) {
 		isActiveVulnMgmt bool
 
 		// Expected values for assertion.
-		sampleExpectedPkgs []*database.RHELv2Package
-		unexpectedPkgs     []*database.RHELv2Package
-		expectedCPEs       []string
```
RTann, line 50: do we test the mapping from content set to CPE in repo2cpe? If not, we should do that there, so we don't lose this test

RTann (changes requested):
I'm not sure we can guarantee there is a single content_manifest per node since its a culmination of all layers of an image. We knew there was one per layer, but there may be more for the image as a whole

pkg/analyzer/detection/detection.go
jvdm: Yes, it may. Obviously, the comment mentions CPE, which will no longer be handled in RHCOS detection. But other aspects of the comment might still be relevant. What is your suggestion?

pkg/analyzer/detection/detection.go
jvdm: Do you mean we should change it or remove it? Not clear what your suggestion is.

pkg/analyzer/detection/detection.go
jvdm: Maybe. At the same time, this tests a namespace attribute rather than if it's a known name or not.

Perhaps leaving it here makes more sense since it is close to the context of where the function is probably going to be used. That is, during detection, to know whether a namespace is certified. This could be better than adding a public interface to a package that is not likely to be used anywhere other than here.

pkg/analyzer/detection/detection.go
jvdm: I see where you coming from. When looking at this patch, it is natural to try to eliminate CPE mapping in detection altogether. But the handling outside RHCOS scanning might be more complex.

I intentionally haven't investigated further. Perhaps database models need to be adjusted in Central Scanner, for example. In any case, I already created an issue ROX-14414 to track this, as we should consider it in a different context.

pkg/analyzer/detection/detection.go
jvdm: It could be a warning. What is your suggestion?

pkg/rhelv2/rpm/rpm.go
jvdm: See 0ee22232

pkg/rhelv2/rpm/rpm.go
jvdm: Oh, interesting. The logic was kept the same as in the previous patch. If, in fact, there can be multiple Content Sets per image/node, we need to know how to process all these content sets.

I noticed that there is some other metadata associated with each content set file, so it might be possible that when working with multiple, we need to process them differently. Or maybe concatenating them on a list. Hopefully.

I am going to keep the original data and follow up separately.

pkg/rhelv2/rpm/rpm_test.go
jvdm: Do you mean https://github.com/stackrox/scanner/blob/0ee22232a00d3ca6d5ea53356d3a33bf9ca00428/pkg/repo2cpe/mapping_test.go#L11

pkg/rhelv2/rpm/rpm.go
jvdm: I actually noticed that the current patch had a bug where instead of returning the first content set we would return the last one. This was fixed at e096fa77

pkg/analyzer/detection/detection.go
```diff
@@ -25,40 +26,63 @@ func DetectComponents(name string, files analyzer.Files, parent *database.Layer,
...
+	// Use the RHELv2 scanner instead.
+	packages, contentSets, err := rpm.ListFeatures(files)
+	if err != nil {
+		return nil, nil, err
+	}
+	rhelfeatures := &database.RHELv2Components{
+		Dist:     namespace.Name,
+		Packages: packages,
+		// CPEs are mapped and returned with content sets for backward compatibility.
+		CPEs:        repo2cpe.Singleton().Get(contentSets),
```
RTann, line 79: Just realized this may panic if it attempts to load a file and fails. It won't panic as long as `REPO_TO_CPE_DIR` is not defined (or is empty). Maybe we should reconsider the panic. It made sense before Scanner started to be used as a library like this, but now if we mess up, Compliance pods can panic. At the moment, it won't, but it could in the future if we aren't careful

pkg/rhelv2/rpm/rpm.go
```diff
@@ -77,33 +84,31 @@ func AddToDependencyMap(filename string, fileData analyzer.FileData, execToDeps,
...
-	// be translated using mapping file provided by Red Hat's PST team.
-	contents := getContentManifestFileContents(files)
+// getContentManifestSets returns the list of content sets defined in the content
+// manifest, if found in the files. Otherwise, `nil` or empty slice is returned.
+func getContentManifestSets(files analyzer.Files) ([]string, error) {
+	var contents []byte
+	for _, prefix := range contentManifestDirs.AsSlice() {
+		for name, file := range files.GetFilesPrefix(prefix) {
+			if strings.HasSuffix(name, ".json") {
+				// Return the first one found, as we are currently assuming there is only one per
```
RTann, line 94: This is what I'm wondering about. Is this a valid assumption for RHCOS nodes? Not sure if this is something confirmed by the RHCOS team

RTann: Ignore this, I didn't see you followed up in the original thread. I'll continue the conversation there https://github.com/stackrox/scanner/pull/1055#discussion_r1072957726

pkg/rhelv2/rpm/rpm.go
RTann: Maybe we should update this to allow for more than one content set per image layer (therefore node as well). That should not hurt image scanning results, and will allow for multiple content sets per node (maybe in a separate PR if we mark this with a TODO or something). I believe just appending each file's `ContentSets` into a slice is sufficient.

You bring up another point: are we sure RHCOS nodes' content_sets use the same format we are currently expecting? I'm not sure. Has this been discussed?

pkg/rhelv2/rpm/rpm.go
```diff
@@ -77,33 +84,31 @@ func AddToDependencyMap(filename string, fileData analyzer.FileData, execToDeps,
...
+				// layer/node.
+				contents = file.Contents
+				break
+			}
+		}
+		if contents != nil {
+			break
+		}
```
vikin91, line 102: So if we have multiple locations to search for - `len(contentManifestDirs.AsSlice()) > 1`  and we find a content in the first location `contentManifestDirs[0]`, why are we are breaking the search here? Shouldn't we check also the other ones - i.e., `contentManifestDirs[1]`, `contentManifestDirs[2]` and so on?

The current behavior would suggest that we are searching all location but we expect that the first hit is correct. Is it safe to assume that?

pkg/analyzer/detection/detection.go
jvdm: I reviewed this, and the comment needs to be refactored. Not only the mention of CPE. It is fixed in 3d1a547e.

pkg/rhelv2/rpm/rpm.go
jvdm: > Shouldn't we check also the other ones - i.e., contentManifestDirs[1], contentManifestDirs[2] and so on?

We want to check all the `contentManifestDirs` but stop on the first file found. That is because we don't expect to find content sets in more than one directory.

Initially, this code would only search in one directory, the default directory for content sets in certified RHEL-based images. Now we are adding RHCOS, which uses a different location -- they have a limitation in their build that prevents them from writing to `/root` at compose time -- but we don't expect to see content sets in both locations.

> The current behavior would suggest that we are searching all locations but we expect that the first hit is correct. Is it safe to assume that?

For RHCOS node scanning, we are starting to debate that assumption.

Based on the [original discussion](https://github.com/coreos/coreos-assembler/issues/2540), it seems the image build would collect all repos used and populate [one single file](https://github.com/coreos/coreos-assembler/blob/main/src/cmdlib.sh#L497). But [there are some reports](https://github.com/stackrox/scanner/pull/1055#discussion_r1072829203) of multiple content sets found "in the wild". Which might be caused by RHCOS customizations? We are not sure, but we are planning to follow up. In any case, maybe we should not try to capture that scenario without knowing more about it.

What do you think, @vikin91?

vikin91: Okay this convinces me. I understand that the potential case of multiple content sets json files will be handled in a separate PR.

vikin91 (approved):
Looks good to me!

I would love to see it in action on OpenShift to see that it works - we can make such test before of after this is merged as this feat is behind a feature flag, so I am leaving the choice to you.

jvdm:
@vikin91:

> I would love to see it in action on OpenShift to see that it works - we can make such test before of after this is merged as this feat is behind a feature flag, so I am leaving the choice to you.

I was planning to do that after working on the Scanner APIs. Glad [you pushed towards it already](https://github.com/stackrox/scanner/pull/1059). I will explore once that PR is merged.

pkg/rhelv2/rpm/rpm.go
RTann: I haven't seen it in RHCOS, but I have in RHEL-based images, which makes me think it could always be possible in RHCOS unless told otherwise. For example, in https://catalog.redhat.com/software/containers/redhat-openjdk-18/openjdk18-openshift/58ada5701fbe981673cd6b10?container-tabs=overview I see this:

```
[jboss@3ae40f064f3c ~]$ ls /root/buildinfo/content_manifests/
redhat-openjdk-18-openjdk18-openshift-container-1.15-2.json
rhel-server-container-7.9-848.1669829396.json
```

RTann: > we are planning to follow up

Do we have a ticket to track this?

RTann (approved):

pkg/rhelv2/rpm/rpm.go
jvdm: > we are planning to follow up

https://issues.redhat.com/browse/ROX-13794

PR #1057, opened 2023-01-17
update github.com/go-git/go-git/v5 version

RTann:
Not sure why this has been ignored by dependabot...


RTann:
> What prompted you to make this update?

I noticed this dependency was not up-to-date when looking at the go.mod file.
 
> Also any idea why CircleCI triggered for this PR but not for other PRs submitted recently?

CircleCI has always been triggered, but it just never showed up before. We still have a configuration file for it: https://github.com/stackrox/scanner/blob/master/.circleci/config.yml

It is just giving an error due to the security incident CircleCI had recently. I think we'd need to update the integration creds if we wanted to continue using it, but we don't use it anymore, so we don't care. We are in the process of removing this integration, so we can ignore the failure for now.

PR #1059, opened 2023-01-18
Support RHEL content sets in a call to GetNodeVulnerabilities

vikin91:
As a result of https://github.com/stackrox/scanner/pull/1053 we decided that `Analyze` will not provide CPEs but instead would include RHEL content sets. This PR adapts the `GetNodeVulnerabilities` to that change, so that content sets are translated to CPEs in Scanner.

## How tested
- [x] CI with the e2e test that covers the `GetNodeVulnerabilities`
- [x] Manually deployed to local cluster (along with https://github.com/stackrox/stackrox/pull/4430) and saw that feature works - see screenshot

Screenshot 2023-01-18 at 18 05 45

### DIFF
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -299,7 +299,7 @@ func (s *serviceImpl) getNodeInventoryVulns(components *v1.Components, isUncerti
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	log.Debugf("Scanned %d components from NodeInventory and returning %d features", len(components.GetRhelComponents()), len(layer.Features))
+	log.Infof("Matched vulnerabilities on %d RHEL components in node inventory", len(components.GetRhelComponents()))
 	return features.ConvertFeatures(layer.Features), nil
 }
 
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -291,7 +291,9 @@ func (s *serviceImpl) getNodeInventoryVulns(components *v1.Components, isUncerti
 	cpes := s.repoToCPE.Get(components.GetRhelContentSets())
 	log.Debugf("Converted content sets '%v' to CPEs '%v'", components.GetRhelContentSets(), cpes)
 	for _, comp := range components.GetRhelComponents() {
-		comp.Cpes = append(comp.Cpes, cpes...)
+		// TODO(ROX-14414): Handle situation when CPEs are provided in parallel to content sets
+		// Overwrite any potential CPEs and stick to content sets to sanitize the API input
+		comp.Cpes = cpes
 	}
 	layer, err := apiV1.GetVulnerabilitiesForComponents(s.db, components, isUncertifiedRHEL)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


api/v1/nodescan/service.go
```diff
@@ -273,17 +276,31 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
...
+	cpes := s.repoToCPE.Get(components.GetRhelContentSets())
+	log.Debugf("Converted content sets '%v' to CPEs '%v'", components.GetRhelContentSets(), cpes)
+	for _, comp := range components.GetRhelComponents() {
+		comp.Cpes = append(comp.Cpes, cpes...)
+	}
+	layer, err := apiV1.GetVulnerabilitiesForComponents(s.db, components, isUncertifiedRHEL)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	log.Debugf("Scanned %d components from NodeInventory and returning %d features", len(components.GetRhelComponents()), len(layer.Features))
```
jvdm, line 300: Nit: I would adjust the terminology. We usually consider "Vulnerability Scanning" as a two parts process: Analysis (detection and inventory creation) and Vulnerability Matching. What you are doing here is the latter:

"Matched vulnerabilities on %d RHEL components in node inventory."

Also, would make this a INFO rather than DEBUG too much in the logs?  Maybe having this sort of info in the bundles would be valuable.

api/v1/nodescan/service.go
```diff
@@ -273,17 +276,31 @@ func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeV
...
 	return resp, nil
 }
 
+func (s *serviceImpl) getNodeInventoryVulns(components *v1.Components, isUncertifiedRHEL bool) ([]*v1.Feature, error) {
+	log.Debugf("Scanning NodeInventory")
+	// Convert content sets to CPEs
+	cpes := s.repoToCPE.Get(components.GetRhelContentSets())
+	log.Debugf("Converted content sets '%v' to CPEs '%v'", components.GetRhelContentSets(), cpes)
+	for _, comp := range components.GetRhelComponents() {
+		comp.Cpes = append(comp.Cpes, cpes...)
```
jvdm, line 294: Maybe ensure `comp.Cpes` is empty. Or were you purposefully appending not to overwrite?

We will revisit this piece of code as part of ROX-14414. By then, we will have to decide how to support input that has CPEs set and inputs that don't. But until then, overwriting to sanitize input seems the right course of action.

jvdm (commented):
I have one question that might require a change, so still pending. I also have a Nit that is not a blocker.

api/v1/nodescan/service.go
jvdm: Also, making this an INFO rather than a DEBUG would cause too much impact in the logs? Having this sort of info in the bundles would be valuable, but I am not sure of the impact. Just push me back if that is not feasible.

api/v1/nodescan/service.go
vikin91: Okay, I was appending not to overwrite - having more CPEs seems not bad to me. But if we have a followup to tackle this, then I will follow your suggestion too overwrite and leave a comment about the follow-up ticket.

jvdm (approved):
LGTM.

PR #1065, opened 2023-01-26
ROX-14592: Ensure that RHCOS Node scan is not skipped by Scanner

vikin91:
## Problem
Currently Scanner does not run v1 Node Scan when RHCOS is detected. It displays the following warning when Node Scanning is run on RHCOS:

Screenshot 2023-01-27 at 11 29 16

This is by design as the v1 scan data from NVD was inaccurate and lead to confusion of users.

## Goal
The goal of this ticket is to disable v1 Node Scan when RHCOS is detected and provide v2 scan instead.
This contains:

- remove the quit-early branch if RHCOS is detected
- do not run v1 scan if and only if: 
    - (1) the FF is enabled, 
    - (2) node inventory is present, 
    - (3) RHCOS system is detected. (Note that 2 implies 3).

## How This was tested?

- [x] CI
    - ⚠️ I needed to skip the existing e2e tests if the feature flag it disabled. Otherwise they fail. I know this isn't great, but if someone knows a solution to enabling the FF in the e2e tests, then please let me know. 
- [x] Run the e2e test `TestGRPCGetRHCOSNodeVulnerabilities` locally with feature flag enabled. 

```
$ go test -tags e2e -timeout=10s -count=1 -v -run ^TestGRPCGetRHCOSNodeVulnerabilities$  github.com/stackrox/scanner/e2etests
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities/Uncertified_scan_is_unsupported_for_RHCOS_and_returns_no_features
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities/Selected_vulnerabilities_should_be_returned_by_the_certified_scan
--- PASS: TestGRPCGetRHCOSNodeVulnerabilities (0.05s)
    --- PASS: TestGRPCGetRHCOSNodeVulnerabilities/Uncertified_scan_is_unsupported_for_RHCOS_and_returns_no_features (0.02s)
    --- PASS: TestGRPCGetRHCOSNodeVulnerabilities/Selected_vulnerabilities_should_be_returned_by_the_certified_scan (0.02s)
PASS
ok  	github.com/stackrox/scanner/e2etests	0.583s
```
- [x] Deployed with feature flag disabled and made sure that nothing that should work is broken
- [x] Deployed along with https://github.com/stackrox/stackrox/pull/4492 and seen this in the UI (still no vulns, but for that, we need content sets)

Screenshot 2023-01-27 at 13 26 01

## What I observed in the UI

- `kernel` component is empty - this should be fixed in https://issues.redhat.com/browse/ROX-14288

Screenshot 2023-01-27 at 13 37 47

- Kubelet is present in the node inventory as: `kubelet v1.24.6+5658434`
- Kube-proxy is present in the node inventory as: `kube-proxy v1.24.6+5658434`
- Kernel is provided in the node overview

Screenshot 2023-01-27 at 13 36 41

### DIFF
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -265,14 +265,15 @@ func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNod
 }
 
 func (s *serviceImpl) getNodeVulnerabilitiesLegacy(_ context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
-	resp := &v1.GetNodeVulnerabilitiesResponse{
-		ScannerVersion: s.version,
-	}
 	if stringutils.AtLeastOneEmpty(req.GetKernelVersion(), req.GetOsImage()) {
 		return nil, status.Error(codes.InvalidArgument, "both os image and kernel version are required")
 	}
 
 	var err error
+	resp := &v1.GetNodeVulnerabilitiesResponse{
+		ScannerVersion: s.version,
+	}
+
 	resp.OperatingSystem, resp.KernelVulnerabilities, resp.KernelComponent, err = s.evaluateLinuxKernelVulns(req)
 	switch err {
 	case nil: // Normal
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -250,7 +250,8 @@ func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNod
 	}
 
 	if !wellknownnamespaces.IsRHCOSNamespace(req.GetComponents().GetNamespace()) {
-		// Non-RHCOS system detected, we can provide list of pkgs but cannot scan them -> setting a note to inform the user
+		// Non-RHCOS system with node inventory detected!
+		// We can provide a list of pkgs but cannot scan them -> setting a note to inform the user.
 		resp.NodeNotes = append(resp.GetNodeNotes(), v1.NodeNote_NODE_UNSUPPORTED)
 	}
 
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -21,7 +21,7 @@ import (
 	"github.com/stackrox/scanner/ext/versionfmt"
 	v1 "github.com/stackrox/scanner/generated/scanner/api/v1"
 	k8scache "github.com/stackrox/scanner/k8s/cache"
-	"github.com/stackrox/scanner/pkg/env"
+	featureFlags "github.com/stackrox/scanner/pkg/features"
 	"github.com/stackrox/scanner/pkg/repo2cpe"
 	"github.com/stackrox/scanner/pkg/version"
 	"github.com/stackrox/scanner/pkg/wellknownnamespaces"
@@ -242,7 +242,7 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
 
 func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
 	// If NodeInventory is empty `req.GetComponents() == nil` then fallback to v1 scanning
-	if req.GetComponents() == nil || !env.RHCOSNodeScanning.Enabled() {
+	if req.GetComponents() == nil || !featureFlags.RHCOSNodeScanning.Enabled() {
 		return s.getNodeVulnerabilitiesLegacy(ctx, req)
 	}
 
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -241,6 +241,7 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
 }
 
 func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
+	log.Infof("Received call to GetNodeVulnerabilities. RHCOSNodeScanning=%t (%s)", env.RHCOSNodeScanning.Enabled(), env.RHCOSNodeScanning.Value())
 	// If NodeInventory is empty `req.GetComponents() == nil` then fallback to v1 scanning
 	if req.GetComponents() == nil || !env.RHCOSNodeScanning.Enabled() {
 		return s.getNodeVulnerabilitiesLegacy(ctx, req)
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -13,6 +13,7 @@ import (
 	apiV1 "github.com/stackrox/scanner/api/v1"
 	"github.com/stackrox/scanner/api/v1/common"
 	"github.com/stackrox/scanner/api/v1/convert"
+	"github.com/stackrox/scanner/api/v1/features"
 	"github.com/stackrox/scanner/cpe/nvdtoolscache"
 	"github.com/stackrox/scanner/database"
 	"github.com/stackrox/scanner/ext/kernelparser"
@@ -20,7 +21,7 @@ import (
 	"github.com/stackrox/scanner/ext/versionfmt"
 	v1 "github.com/stackrox/scanner/generated/scanner/api/v1"
 	k8scache "github.com/stackrox/scanner/k8s/cache"
-	"github.com/stackrox/scanner/pkg/features"
+	featureFlags "github.com/stackrox/scanner/pkg/features"
 	"github.com/stackrox/scanner/pkg/repo2cpe"
 	"github.com/stackrox/scanner/pkg/version"
 	"google.golang.org/grpc"
@@ -247,7 +248,7 @@ func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNod
 		ScannerVersion: s.version,
 	}
 	// If NodeInventory is empty `req.GetComponents() == nil` then fallback to v1 scanning
-	if req.GetComponents() == nil || !features.RHCOSNodeScanning.Enabled() {
+	if req.GetComponents() == nil || !featureFlags.RHCOSNodeScanning.Enabled() {
 		return s.getNodeVulnerabilitiesLegacy(ctx, req, resp)
 	}
 
@@ -297,6 +298,7 @@ func (s *serviceImpl) getNodeVulnerabilitiesLegacy(_ context.Context, req *v1.Ge
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	return resp, nil
 }
 
 func (s *serviceImpl) getNodeInventoryVulns(components *v1.Components, isUncertifiedRHEL bool) ([]*v1.Feature, error) {
--- a/api/v1/nodescan/service.go
+++ b/api/v1/nodescan/service.go
@@ -241,12 +241,13 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
 }
 
 func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
-	resp := &v1.GetNodeVulnerabilitiesResponse{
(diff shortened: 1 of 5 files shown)


api/v1/nodescan/service.go
```diff
@@ -238,16 +240,35 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
 	return nil, nil
 }
 
-func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
-	if stringutils.AtLeastOneEmpty(req.GetKernelVersion(), req.GetOsImage()) {
-		return nil, status.Error(codes.InvalidArgument, "both os image and kernel version are required")
+func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
+	resp := &v1.GetNodeVulnerabilitiesResponse{
```
RTann, line 244: nit: this isn't needed until after the legacy check. Let's move this below that check

e2etests/node_scan_rhcos_test.go
```diff
@@ -5,6 +5,7 @@ package e2etests
 
 import (
 	"context"
+	"github.com/stackrox/scanner/pkg/features"
```
RTann, line 8: nit: this probably failed style checks

e2etests/node_scan_rhcos_test.go
```diff
@@ -150,6 +151,9 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
 	for name, c := range cases {
 		t.Run(name, func(t *testing.T) {
 			c := c
+			if !features.RHCOSNodeScanning.Enabled() {
```
RTann, line 154: is this needed inside the loop or can this be checked at the very beginning of `func Test...`?

api/v1/nodescan/service.go
```diff
@@ -238,16 +240,35 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
...
+	}
+	// If NodeInventory is empty `req.GetComponents() == nil` then fallback to v1 scanning
+	if req.GetComponents() == nil || !featureFlags.RHCOSNodeScanning.Enabled() {
+		return s.getNodeVulnerabilitiesLegacy(ctx, req, resp)
+	}
+
+	if !wellknownnamespaces.IsRHCOSNamespace(req.GetComponents().GetNamespace()) {
+		// Non-RHCOS system with node inventory detected!
+		// We can provide a list of pkgs but cannot scan them -> setting a note to inform the user.
+		resp.NodeNotes = append(resp.GetNodeNotes(), v1.NodeNote_NODE_UNSUPPORTED)
```
jvdm, line 255: Should we add the same note for inventories in certified RHEL namespaces (including RHCOS) but without content sets?

e2etests/node_scan_rhcos_test.go
```diff
@@ -150,6 +151,9 @@ func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
...
+				t.Skip("This test requires ROX_RHCOS_NODE_SCANNING=true")
```
jvdm, line 155: Can we forcibly enable the feature flag in the UT and always test the code? Would it make more sense, given we can enable the flag in dev with environment variables?

e2etests/node_scan_rhcos_test.go
vikin91: It is so annoying that Goland doesn't do that automatically :( I somehow configured it for stackrox but I do not remember how to repeat this for scanner repo.

e2etests/node_scan_rhcos_test.go
vikin91: Yes that would be great! The thing is that I tried to do this already and failed to enable it only for the e2e tests. We would need to tell the CI to build and deploy with that FF enabled. Do you maybe know where we could do this?

api/v1/nodescan/service.go
vikin91: Like this? [a5f7af3](https://github.com/stackrox/scanner/pull/1065/commits/a5f7af3b570a8b84564da9244ec7a9d373a124c8)

vikin91:
The e2e test failure seems unrelated: `FAIL: TestDistrolessVulnImages/gcr.io/distroless/python2.7@sha256:6d3895c4a1629ac99e73c7dc9cbe0ad8cb213d6cdebf3e835c2c388fc5aab1b2 (18.95s)`

e2etests/node_scan_rhcos_test.go
```diff
@@ -102,6 +103,9 @@ func buildRequest(notes []v1.Note) *v1.GetNodeVulnerabilitiesRequest {
 }
 
 func TestGRPCGetRHCOSNodeVulnerabilities(t *testing.T) {
+	if !features.RHCOSNodeScanning.Enabled() {
```
RTann, line 106: you can set it via `t.Setenv(features.RHCOSNodeScanning.EnvVar())`. Think of it as the same thing as `envisolator` which we used to have, but it resets the env var for us at the end of the test

e2etests/node_scan_rhcos_test.go
RTann: I left a comment. You can enable it on a per-unit test basis

api/v1/nodescan/service.go
```diff
@@ -238,7 +240,31 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
...
+		return s.getNodeVulnerabilitiesLegacy(ctx, req)
+	}
+
+	resp := &v1.GetNodeVulnerabilitiesResponse{
+		ScannerVersion: s.version,
+	}
+
+	if !wellknownnamespaces.IsRHCOSNamespace(req.GetComponents().GetNamespace()) || len(req.GetComponents().GetRhelContentSets()) == 0 {
+		// Non-RHCOS system with node inventory detected!
+		// We can provide a list of pkgs but cannot scan them -> setting a note to inform the user.
```
jvdm, line 255: Should we update this comment?

```suggestion
		// Node components exist in the request, but we cannot perform vulnerability matching
		// if (a.) its not an RHCOS namespaces or, (b.) if it is an RHCO namespace missing content
		// sets. The latter occurs for old RHCO versions we don't support.
```

jvdm (changes requested):

api/v1/nodescan/service.go
```diff
@@ -238,7 +240,31 @@ func (s *serviceImpl) getRuntimeVulns(containerRuntime *v1.GetNodeVulnerabilitie
 	return nil, nil
 }
 
-func (s *serviceImpl) GetNodeVulnerabilities(_ context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
+func (s *serviceImpl) GetNodeVulnerabilities(ctx context.Context, req *v1.GetNodeVulnerabilitiesRequest) (*v1.GetNodeVulnerabilitiesResponse, error) {
+	// If NodeInventory is empty `req.GetComponents() == nil` then fallback to v1 scanning
+	if req.GetComponents() == nil || !featureFlags.RHCOSNodeScanning.Enabled() {
```
RTann, line 245: just curious, why not just `!featureFlags.RHCOSNodeScanning.Enabled()`?

api/v1/nodescan/service.go
RTann: I think ^ is missing an S in RHCOS :)

api/v1/nodescan/service.go
vikin91: A request with `req.GetComponents() == nil` means that this is the legacy request (data coming not from Compliance), so I wanted this piece of code to behave exactly as before the change.

The second reason is that all the code below this line relies on `req.GetComponents()` not being nil, so this serves as panic protection.

e2etests/node_scan_rhcos_test.go
vikin91: Unfortunately this didn't seem to work: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/stackrox_scanner/1065/pull-ci-stackrox-scanner-master-e2e-tests/1620680953611423744

vikin91: Nope, it does not work. The scanner-under-test for e2e would need to be deployed with that flag enabled and this is not the case currently. See:

```
    Environment:
      POD_NAME:                  scanner-767f474d7-kc8hz (v1:metadata.name)
      ROX_SKIP_PEER_VALIDATION:  true
      ROX_RHCOS_NODE_SCANNING:   false
```

in the `roxci-artifacts/scanner/33227a261a3131cf9921c383a4846feff252d4ac/1620694223667859456-pull-ci-stackrox-scanner-master-e2e-tests/e2e-tests/k8s-service-logs/stackrox/scanner-767f474d7-kc8hz_describe.log` form the [recent CI run ](https://console.cloud.google.com/storage/browser/_details/roxci-artifacts/scanner/33227a261a3131cf9921c383a4846feff252d4ac/1620694223667859456-pull-ci-stackrox-scanner-master-e2e-tests/e2e-tests/k8s-service-logs/stackrox/scanner-767f474d7-kc8hz_describe.log;tab=live_object?authuser=1) (assumes authuser=1 for accessing the logs produced in the CI)

vikin91: I think the `t.Setenv` is good for setting the values in `pkg/env/list.go` (at least that is what the code search suggests) which can be changed at runtime. The vars from `pkg/features/list.go` seem to be built-time variables and maybe they do not allow setting them like that?

Let me try moving the feature flag from `pkg/features/list.go` to `pkg/env/list.go`. This would result in the potential possibility of our end users enabling it accidentally, but may be this is not a real risk?

vikin91: Damn, that didn't work either. Is this possible that it does not work for e2e tests?

I even added logging in the function that handles this event:

```
"Received call to GetNodeVulnerabilities. RHCOSNodeScanning=false (false)"
```
while the tests claim:

```
$ go test -tags e2e -timeout=10s -count=1 -v -run ^TestGRPCGetRHCOSNodeVulnerabilities$  github.com/stackrox/scanner/e2etests
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities
    node_scan_rhcos_test.go:108: FF set to: true
(...)
=== RUN   TestGRPCGetRHCOSNodeVulnerabilities/Uncertified_scan_is_unsupported_for_RHCOS_and_returns_no_features
--- FAIL: TestGRPCGetRHCOSNodeVulnerabilities (0.66s)
    --- FAIL: TestGRPCGetRHCOSNodeVulnerabilities/Selected_vulnerabilities_should_be_returned_by_the_certified_scan (0.47s)
    --- PASS: TestGRPCGetRHCOSNodeVulnerabilities/Uncertified_scan_is_unsupported_for_RHCOS_and_returns_no_features (0.19s)
FAIL
FAIL	github.com/stackrox/scanner/e2etests	1.347s
FAIL
```

vikin91: I will see what CI says, maybe this is an issue with the local running of those tests.

vikin91: Nope, it didn't help. I reverted the excessive logging, (rebased), and I am curious 🤔  whether we have any other ideas to resolve this.

jvdm (approved):
LGTM.

Maddosaurus (approved):
LGTM!

Edit: You had to skip the e2e-tests, which is why they fail in prow, correct?

vikin91:
> LGTM!
> 
> Edit: You had to skip the e2e-tests, which is why they fail in prow, correct?

They fail because I am unable to enable the feature flag for the e2e test. I could skip that test but we would loose visibility. If I keep it the way it is, it will be constantly failing in CI.

The solution here would be to enable the FF for the Scanner-under-test in CI, but I don't know how to do this :) I will see, maybe I dig something out, but If not, we can discuss it in Today's RHCOS sync.

vikin91:
I rebased to accommodate changes from https://github.com/stackrox/scanner/pull/1074 and use the branch from this PR in Today's demo

vikin91 (commented):
Move feature flag back to features pkg

vikin91 (commented):
Set ROX_RHCOS_NODE_SCANNING to true in the chart

vikin91:
This is good to be merged after CI is green and Ross gives his review.

RTann (approved):

PR #1066, opened 2023-01-27
Add usr/share/buildinfo to node scan

jvdm:
## Description

Explicitly add `usr/share/buildinfo` to the list of directories to detect files in node scanning. UT added to cover the new directory.

In node scan, we hard-code the directories to look for files during detection. To detect content sets, we need to include `usr/share/buildinfo/content_manifests.json`, hence the addition.

## Testing

- UT
- Openshift 4.12 cluster.

### DIFF
--- a/pkg/analyzer/nodes/node.go
+++ b/pkg/analyzer/nodes/node.go
@@ -101,7 +101,8 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 		files: make(map[string]*fileMetadata),
 	}
 	m := metrics.FileExtractionMetrics{}
-	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm"} { // TODO(ROX-13771): Use range matcher.GetCommonPrefixDirs() again after fixing
+	// TODO(ROX-13771): Use `range matcher.GetCommonPrefixDirs()` again after fixing.
+	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm", "usr/share/buildinfo"} {
 		if err := n.addFiles(filepath.FromSlash(dir), matcher, &m); err != nil {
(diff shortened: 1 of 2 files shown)


vikin91 (approved):

vikin91:
Just maybe give the style another look before merging ;)

vikin91:
I tested this PR with my other branches (https://github.com/stackrox/stackrox/pull/4571 + https://github.com/stackrox/scanner/pull/1065 + https://github.com/stackrox/stackrox/pull/4492 ) on OpenShift 4.11 and it looks that it works 🎉  
Screenshot 2023-01-27 at 15 11 00

pkg/analyzer/nodes/node.go
```diff
@@ -101,7 +101,8 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 		files: make(map[string]*fileMetadata),
 	}
 	m := metrics.FileExtractionMetrics{}
-	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm"} { // TODO(ROX-13771): Use range matcher.GetCommonPrefixDirs() again after fixing
+	// TODO(ROX-13771): Use `range matcher.GetCommonPrefixDirs()` again after fixing.
+	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm", "usr/share/buildinfo"} {
```
RTann, line 105: Perhaps for another PR, but since we are now looking at the RHCOS-specific locations, do we still need to look at `etc/` or `var/lib/rpm` here? Those ultimately just symlink to the other two paths. If we do get rid of it, we should also remove the symlink file reading stuff we added, as it would no longer be needed (and could be potentially dangerous if not handled correctly. We already fixed a bug where it was possible for us to read our own file outside of the given `root`)

PR #1068, opened 2023-01-31
ROX-13107: Hardcode content sets in RHCOS 4.7 to 4.9

jvdm:
## Description

Hardcode content sets and use standard CPE resolution logic to provide CPE info on older OpenShifts (4.7, 4.8, 4.9). Other versions are ignored, and no content sets are returned.

OpenShift 4.10 onwards provides `/root/content_manifests/*.json.` -- the manifest that lists repositories used to install packages in the O.S. used to find CPEs to match against OVAL. But OCP < 4.10 does not. Nevertheless, we can construct the content sets by hand by following the rules used by RHCOS itself. This will allow scanner to match vulnerabilities for OCP 4.7, 4.8 and 4.9.

## Tests

- UTs
- Deployment to Openshift 4.9. Added a debug line to Compliance:
  
  ```
  collection/nodeinventorizer: 2023/01/31 08:04:25.319831 nodeinventory.go:33: Debug: Content sets found under /host: [rhel-8-for-x86_64-baseos-eus-rpms__8_DOT_4 rhel-8-for-x86_64-appstream-eus-rpms__8_DOT_4 fast-datapath-for-rhel-8-x86_64-rpms rhocp-4.9-for-rhel-8-x86_64-rpms]
  ```
  
   Which maps correctly to the information found in `etc/os-release`:
   
   ```
   $ cat /host/etc/os-release 
  NAME="Red Hat Enterprise Linux CoreOS"
  VERSION="49.84.202212201621-0"
  ID="rhcos"
  ID_LIKE="rhel fedora"
  VERSION_ID="4.9"
  PLATFORM_ID="platform:el8"
  PRETTY_NAME="Red Hat Enterprise Linux CoreOS 49.84.202212201621-0 (Ootpa)"
  ANSI_COLOR="0;31"
  CPE_NAME="cpe:/o:redhat:enterprise_linux:8::coreos"
  HOME_URL="https://www.redhat.com/"
  DOCUMENTATION_URL="https://docs.openshift.com/container-platform/4.9/"
  BUG_REPORT_URL="https://bugzilla.redhat.com/"
  REDHAT_BUGZILLA_PRODUCT="OpenShift Container Platform"
  REDHAT_BUGZILLA_PRODUCT_VERSION="4.9"
  REDHAT_SUPPORT_PRODUCT="OpenShift Container Platform"
  REDHAT_SUPPORT_PRODUCT_VERSION="4.9"
  OPENSHIFT_VERSION="4.9"
  RHEL_VERSION="8.4"
  OSTREE_VERSION='49.84.202212201621-0'
  ```

### DIFF
--- a/pkg/analyzer/detection/detection.go
+++ b/pkg/analyzer/detection/detection.go
@@ -1,14 +1,20 @@
 package detection
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/stackrox/scanner/database"
 	"github.com/stackrox/scanner/ext/featurefmt"
 	"github.com/stackrox/scanner/ext/featurens"
+	"github.com/stackrox/scanner/ext/versionfmt"
+	versionfmtrpm "github.com/stackrox/scanner/ext/versionfmt/rpm"
 	"github.com/stackrox/scanner/pkg/analyzer"
 	"github.com/stackrox/scanner/pkg/component"
 	features2 "github.com/stackrox/scanner/pkg/features"
+	"github.com/stackrox/scanner/pkg/osrelease"
 	"github.com/stackrox/scanner/pkg/repo2cpe"
 	"github.com/stackrox/scanner/pkg/rhelv2/rpm"
 	"github.com/stackrox/scanner/pkg/wellknownnamespaces"
@@ -86,6 +92,17 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
 	if err != nil {
 		return nil, nil, err
 	}
+	// There is still a chance to set content sets for some RHCOS versions.
+	if len(contentSets) == 0 && wellknownnamespaces.IsRHCOSNamespace(namespace.Name) {
+		contentSets, err = getHardcodedRHCOSContentSets(files)
+		if err != nil {
+			logrus.WithError(err).Errorf("failed to get RHCOS content sets for %v: %v", namespace, err)
+		}
+	}
+	// Leave a warning for certified scans without content sets.
+	if len(contentSets) == 0 {
+		logrus.Warningf("empty content sets for RHEL certified namespace %v", namespace)
+	}
 	rhelfeatures := &database.RHELv2Components{
 		Dist:     namespace.Name,
 		Packages: packages,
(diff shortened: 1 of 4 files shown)


pkg/analyzer/detection/detection.go
```diff
@@ -99,6 +116,65 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
...
+		// https://access.redhat.com/articles/6907891
+		fmt.Sprintf("rhel-8-for-x86_64-baseos-eus-rpms__%s", rhelSuffix),
+		fmt.Sprintf("rhel-8-for-x86_64-appstream-eus-rpms__%s", rhelSuffix),
+		"fast-datapath-for-rhel-8-x86_64-rpms",
+		fmt.Sprintf("rhocp-%s-for-rhel-8-x86_64-rpms", metadata["OPENSHIFT_VERSION"]),
+	}, nil
+}
+
+// isRPMVersionInInterval checks if the provided rpm version is within the specified interval
+// `(minExcluded, maxIncluded]`.
```
vikin91, line 159: nit
```suggestion
// `[minIncluded, maxExcluded)`.
```

pkg/analyzer/detection/detection.go
```diff
@@ -99,6 +116,65 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
...
+		return false, nil
+	}
+	cmp, err = versionfmt.Compare(versionfmtrpm.ParserName, version, maxExcluded)
+	if err != nil {
+		return false, err
+	}
+	if cmp >= 0 {
+		return false, nil
+	}
+	return true, nil
```
vikin91, line 175: Optionally
```suggestion
	return cmp < 0, nil
```

pkg/osrelease/osrelease.go
```diff
@@ -15,19 +16,26 @@ var (
...
-			os = strings.Replace(strings.ToLower(r[1]), "\"", "", -1)
-		}
-
-		r = versionPattern.FindStringSubmatch(line)
-		if len(r) == 2 {
-			version = strings.Replace(strings.ToLower(r[1]), "\"", "", -1)
+		parts := strings.SplitN(line, "=", 2)
+		key := parts[0]
+		if len(fields) == 0 || fieldsSet.Contains(key) {
+			osReleaseMap[key] = strings.Replace(strings.ToLower(strings.TrimSpace(parts[1])), `"`, "", -1)
```
vikin91, line 37: `parts[1]` can panic if the string does not contain a `=`

vikin91 (approved):
LGTM! I found only minor things. You may want to check the comment about potential panic before merging

pkg/analyzer/detection/detection.go
```diff
@@ -99,6 +116,65 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
...
 
+// getHardcodedRHCOSContentSets returns hard coded content sets for supported
+// RHCOS versions. When the RHCOS version is not supported, it returns empty. The
+// function assumes we are working in a RHCOS namespace, and uses
+// `/etc/os-release` to gather versions and metadata. But, if the expected
+// metadata is not found, is inconsistent or cannot be retrieved, it returns an
+// error.
+func getHardcodedRHCOSContentSets(files analyzer.Files) ([]string, error) {
+	// Get metadata from etc/os-release, ensure it exists.
+	osReleaseData, exists := files.Get("etc/os-release")
```
Maddosaurus, line 127: `os-release` should be our source of truth for which RHCOS version we're running on, but thinking about long term impacts on the codebase:
We would have 2 places to determine what OS we're running on, `/ext/features/redhatrelease/` and this place here.
This might lead to a situation where we update one but not the other, and we would have a split-brain-situation.
Should/Could we use the code in `redhatrelease` to determine the RHCOS namespace here?

pkg/analyzer/detection/detection.go
```diff
@@ -99,6 +116,65 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
...
+	// Verify the RHCOS version is supported.
+	ok, err := isRPMVersionInInterval(metadata["VERSION_ID"], "4.7", "4.10")
+	if !ok {
+		return nil, err
+	}
+	// Format the content sets based on the metadata found.
+	rhelSuffix := strings.Replace(metadata["RHEL_VERSION"], ".", "_DOT_", 1)
+	return []string{
+		// The mapping from RHCOS version to RHEL minor releases can be found here:
+		// https://access.redhat.com/articles/6907891
```
Maddosaurus, line 150: Out of interest: This gives us the version mappings, but how would I get from version mappings to the content set names in this slice? Do we have a resource for that as well or is that tribal knowledge?

Maddosaurus (approved):
Changes look good, thank you for the PR. The comment about the RHCOS namespace is a thought that crossed my mind, totally optional.

pkg/analyzer/detection/detection.go
jvdm: Cool, but I leave it as is. The additional verbosity makes it a slightly cleaner narrative in the context of this function, IMO.

pkg/osrelease/osrelease.go
jvdm: Fix at d17c8fe8

pkg/analyzer/detection/detection.go
jvdm: >  This might lead to a situation where we update one but not the other, and we would have a split-brain-situation.

I think it would make more sense to update `redhatrelease` to use `/ec/os-release`, and enrich the namespace object with the additional metadata we need that only exist, to my knowledge, in `/etc/os-release`.

The code here doesn't want to determine the `namespace`. Instead, it is an extension of the namespace detection that goes to `/etc/os-release` to grab additional data. In fact, if we don't see the data we expect to see, in the format we understand, we ignore os-release and leave a warning behind.

Having this logic here, far from the actual namespace detection logic, might be a risk for maintainability because it is hard to connect the dots. But the future of namespace detection and other things is relatively unstable now that we are working with consolidation, so I think moving with what we have here is a good trade-off. What do you think?

pkg/analyzer/detection/detection.go
jvdm: Could you help me understand your question? Would you like to know how we formed the content set strings?

If yes, this was confirmed with RHCOS for 4.7, 4.8, and 4.9 -- The release suffix was confirmed by the table referenced and inspecting https://www.redhat.com/security/data/metrics/repository-to-cpe.json

pkg/analyzer/detection/detection.go
Maddosaurus: > Having this logic here, far from the actual namespace detection logic, might be a risk for maintainability because it is hard to connect the dots.

Yes, this was my primary concern, and I concur that we should move with the current state, but keep this in mind when/if we rework `redhatrelease`. Thanks!

pkg/analyzer/detection/detection.go
Maddosaurus: I was missing the JSON you have provided here in my mental model, thank you!

PR #1073, opened 2023-02-01
ROX-13796: benchmark node analyze

daynewlee:
The bash script is for creating RPM files and install RPM packages so we can benchmark node scanning `analyze` function on varieties of package list sizes.

The benchmark file is how to test the `analyze` function and collect data.


benchmarks/analyzeNode/analyze_node_test.go
```diff
@@ -0,0 +1,33 @@
...
+	b.ResetTimer()
+	runtime.GC()
+	runtime.ReadMemStats(&m1)
+
+	for i := 0; i < b.N; i++ {
+		var err error
+		_, err = node.Analyze("testNode", pathName, node.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: true})
+		require.NoError(b, err)
+	}
+	runtime.ReadMemStats(&m2)
```
vikin91, line 30: The way we gather memory stats may be inaccurate - we measure memory before the benchmark and after it. In between, there will be more that one execution of `Analyze`. I think, we should report memory allocation per single call to Analyze...

or skip doing this this way and rely on the numbers that go reports automatically with benchmem: ` go test -bench=foo -benchmem`.

I would be in favor of the second option.

benchmarks/analyzeNode/analyze_node_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+func runBenchmarkAnalyzeNode(b *testing.B, pathName string) {
+	var m1, m2 runtime.MemStats
+	b.ResetTimer()
+	runtime.GC()
+	runtime.ReadMemStats(&m1)
+
+	for i := 0; i < b.N; i++ {
+		var err error
+		_, err = node.Analyze("testNode", pathName, node.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: true})
```
vikin91, line 27: We can ignore the error completely here - checking the error costs us additional time that will be included in the stats. We want to have only `node.Analyze` inside the loop.

benchmarks/analyzeNode/analyze_node_test.go
```diff
@@ -0,0 +1,33 @@
...
+	_ "github.com/stackrox/scanner/ext/imagefmt/docker"
+)
+
+func BenchmarkAnalyzeNode(b *testing.B) {
+	runBenchmarkAnalyzeNode(b, "fake path")
+}
+
+func runBenchmarkAnalyzeNode(b *testing.B, pathName string) {
+	var m1, m2 runtime.MemStats
+	b.ResetTimer()
```
vikin91, line 21: This is not needed, as the timer is 0 at the begining.

vikin91 (commented):
This looks nice, but we are missing a manual how to reproduce the results. Could you provide a readme at the pkg level or a block-comment (`/* ... */`) in the benchmark code with instructions? Imagine someone without a prior knowledge would like to reproduce the same experiments - which command they would need to run?

scripts/generate.py
```diff
@@ -0,0 +1,16 @@
...
+        f.write("Summary: Summary here\n")
+        f.write("Name: package_name"+str(x)+"\n")
+        f.write("Version: 0\n")
+        f.write("Release: 0\n")
+        f.write("License: Public\n")
+        f.write("Group: Applications/System\n")
+        f.write("Requires: bash\n")
+        f.write("%description\n")
+        f.write("Package description here\n")
+        f.write("%files")
```
jvdm, line 16: Is it OK to remove the dependency on Python, especially if this is supposed to run in the node itself? 

(Note: Applying the suggestion means we should rename the file to `.sh` -- it is not required to execute the script, but it will definitely confuse people if you leave `*.py`).

```suggestion
#/usr/bin/env bash

# This script is to generate simple RPM spec files:
# spec => rpm files: run rpmbuild -bb *.spec
# install rpm: yum localinstall /root/rpmbuild/RPMS/x86_64/*.rpm

for n in {1..100}; do
    cat >pkg-$(printf "%03d" ${n}).spec <<EOF
Name: package_name_${n}
Summary: Package number ${n}
Version: 0
Release: 0
License: Public
Group: Applications/System
Requires: bash
%description
Package number ${n}
%files
EOF
done
```

jvdm (changes requested):
Also, do you know how this script is supposed to be used? Is it run in the node itself? Or do we build the `*.rpm`s locally and install them in the node another way? 

Should we automate the steps to reproduce the test in our CI? Not a blocker. More like a follow-up. :smile: 

Also, a description in the PR would be great. It gives context to the change, helping people without context to understand it.

jvdm:
> This looks nice, but we are missing a manual how to reproduce the results. Could you provide a readme at the pkg level or a block-comment (`/* ... */`) in the benchmark code with instructions? Imagine someone without a prior knowledge would like to reproduce the same experiments - which command they would need to run?

Can I suggest we time-box looking into automating this so it can be run as part of the "performance e2e" CI? If it's too much work, please, document how to run the test in a way that someone without context can.

daynewlee:
> Also, do you know how this script is supposed to be used? Is it run in the node itself? Or do we build the `*.rpm`s locally and install them in the node another way?
> 
> Should we automate the steps to reproduce the test in our CI? Not a blocker. More like a follow-up. 😄
> 
> Also, a description in the PR would be great. It gives context to the change, helping people without context to understand it.

It would be a great idea if we can install RPM packages in CI and test this function. Perhaps we can create a phase 2 ticket for that.

jvdm (changes requested):
1. Change `scripts/generate.py` to `scripts/generate-rpm-specs.sh` or `scripts/generate-rpm-specs`.
2. Are we following up on Piotr comment on using `-benchmem` instead of measuring memory manually?

jvdm (approved):
LGTM.

Maddosaurus:
@daynewlee Did you follow up on adding instructions how to reproduce your results?
I wasn't able to find them, and I wanted to try and reproduce your results.

PR #1074, opened 2023-02-01
ROX-13107: Add missing content sets to the hard coded list for RHCOS 4.7, 4.8 and 4.9.

jvdm:
## Description

Add missing content sets to the hard-coded list for RHCOS 4.7, 4.8, and 4.9.

The additional content sets were found after the following steps, which I would like reviewers to also perform for validity:

1. In the corp network, you can inspect the RHCOS source code to determine which RPM repositories were enabled during the build [on gitlab](https://gitlab.cee.redhat.com/coreos/redhat-coreos/).
2. For each of the branches: 4.7, 4.8 and 4.9:
   1. Open all the `*.repo` files.
   2. For each repository, confirm the content set:
      a. Open https://www.redhat.com/security/data/metrics/repository-to-cpe.json
      b. Replace the architecture with `x86_64`, and grep for the repository path.
      c. Copy the content set name for the entry found.

Notice that for 4.7, `advanced-virt-for-rhel-8-x86_64-eus-rpms` is not used, and the code acts accordingly.

## Tests

- UTs.

### DIFF
--- a/pkg/analyzer/detection/detection.go
+++ b/pkg/analyzer/detection/detection.go
@@ -145,14 +145,21 @@ func getHardcodedRHCOSContentSets(files analyzer.Files) ([]string, error) {
 	}
 	// Format the content sets based on the metadata found.
 	rhelSuffix := strings.Replace(metadata["RHEL_VERSION"], ".", "_DOT_", 1)
-	return []string{
-		// The mapping from RHCOS version to RHEL minor releases can be found here:
-		// https://access.redhat.com/articles/6907891
+	sets := []string{
+		// RHEL8.
 		fmt.Sprintf("rhel-8-for-x86_64-baseos-eus-rpms__%s", rhelSuffix),
 		fmt.Sprintf("rhel-8-for-x86_64-appstream-eus-rpms__%s", rhelSuffix),
+		fmt.Sprintf("rhel-8-for-x86_64-nfv-tus-rpms__%s", rhelSuffix),
+		// Fast datapath.
 		"fast-datapath-for-rhel-8-x86_64-rpms",
+		// Openshift RHOCP.
 		fmt.Sprintf("rhocp-%s-for-rhel-8-x86_64-rpms", metadata["OPENSHIFT_VERSION"]),
-	}, nil
+	}
+	if metadata["VERSION_ID"] != "4.7" {
+		// This is only specified for 4.8 and 4.9
+		sets = append(sets, "advanced-virt-for-rhel-8-x86_64-eus-rpms")
+	}
+	return sets, nil
 }
 
 // isRPMVersionInInterval checks if the provided rpm version is within the specified interval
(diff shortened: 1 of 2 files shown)


pkg/analyzer/detection/detection_test.go
```diff
@@ -88,8 +88,26 @@ OSTREE_VERSION='49.84.202212201621-0'
...
+				"advanced-virt-for-rhel-8-x86_64-eus-rpms",
+			},
+		},
+		{
+			name: "when ocp/4.7 then should not use advanced-virt",
+			osReleaseContents: `
+ID="rhcos"
+VERSION_ID="4.7"
+OPENSHIFT_VERSION="4.7"
+RHEL_VERSION="8.4"
```
RTann, line 103: can we add another test to ensure we can correctly get something like `8_DOT_3` or something other than 4 as well?

jvdm:
```suggestion
RHEL_VERSION="8.3"
```

pkg/analyzer/detection/detection_test.go
```diff
@@ -88,8 +88,26 @@ OSTREE_VERSION='49.84.202212201621-0'
...
+`,
+			want: []string{
+				"rhel-8-for-x86_64-baseos-eus-rpms__8_DOT_4",
+				"rhel-8-for-x86_64-appstream-eus-rpms__8_DOT_4",
+				"rhel-8-for-x86_64-nfv-tus-rpms__8_DOT_4",
```
jvdm, line 108:
```suggestion
				"rhel-8-for-x86_64-baseos-eus-rpms__8_DOT_3",
				"rhel-8-for-x86_64-appstream-eus-rpms__8_DOT_3",
				"rhel-8-for-x86_64-nfv-tus-rpms__8_DOT_3",
```

jvdm: @ross like this?

pkg/analyzer/detection/detection_test.go
jvdm: @RTann like this?

RTann: yeah as long as we can confirm that it doesn't just happen to work for 8.4 only

vikin91:
Pinging @daynewlee as this PR has influence on [ROX-13801](https://issues.redhat.com/browse/ROX-13801) - we should test the clusters after this is merged and used in stackrox.

Maddosaurus (approved):
Thanks for the clear instructions!
I was able to verify the path to discover the additional content sets 😄

pkg/analyzer/detection/detection.go
```diff
@@ -145,14 +145,21 @@ func getHardcodedRHCOSContentSets(files analyzer.Files) ([]string, error) {
 	}
 	// Format the content sets based on the metadata found.
 	rhelSuffix := strings.Replace(metadata["RHEL_VERSION"], ".", "_DOT_", 1)
-	return []string{
-		// The mapping from RHCOS version to RHEL minor releases can be found here:
-		// https://access.redhat.com/articles/6907891
```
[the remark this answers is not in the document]
jvdm: Great call. I added the instructions in the JIRA ticket. The JIRA is connected to the PR, which is connected to the git commit that introduced the changes. A README might be better, but we are at least covered.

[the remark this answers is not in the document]
jvdm: By the way, thank you very much for taking the time to double-checking the paths.

PR #1075, opened 2023-02-01
Use downstream API to allow use of go1.19 memory mgmt changes to GC

connorgorman:
These changes are mirrored in https://github.com/stackrox/stackrox/pull/4662 but should get this to run on scale env, e2e tests, etc

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -22,6 +22,14 @@ spec:
         image: {{ .Values.scannerImage}}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
         env:
+        - name: GOMEMLIMIT
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.memory
+        - name: GOMAXPROCS
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.cpu
         - name: POD_NAME
           valueFrom:
             fieldRef:


chart/templates/deployment.yaml
```diff
@@ -22,6 +22,14 @@ spec:
         image: {{ .Values.scannerImage}}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
         env:
+        - name: GOMEMLIMIT
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.memory
+        - name: GOMAXPROCS
```
RTann, line 29: This has to be an integer, right? I don't know if `limits.cpu` has to be an int. Also, I'm curious why this was added

chart/templates/deployment.yaml
```diff
@@ -22,6 +22,14 @@ spec:
         image: {{ .Values.scannerImage}}:{{ .Values.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
         env:
+        - name: GOMEMLIMIT
+          valueFrom:
+            resourceFieldRef:
+              resource: limits.memory
```
RTann, line 28: similar for this. I think this has to be an `int64` in Go, and I don't know if this is required in k8s. Also, the `Suggested Uses` here https://tip.golang.org/doc/gc-guide#Memory_limit mentions we should leave 5-10% of wiggle room

PR #1077, opened 2023-02-01
Update max known OpenShift version to 4.13

RTann:
OpenShift 4.13 is now the max known OpenShift version.

This actually is not needed at this time, as the CPEs in the OVAL v2 feeds specify this for us; however, we still have this to protect from potential future OVAL v2 changes.

Currently, the OVAL v2 feed for unfixable OpenShift vulns give us CPEs of the form:
`cpe:/a:redhat:openshift:4.13`, but there were talks of changing it to `cpe:/a:redhat:openshift:4`. We have this variable to capture what we know to be the latest version of OpenShift in case this change happens...

### DIFF
--- a/pkg/cpeutils/utils.go
+++ b/pkg/cpeutils/utils.go
@@ -10,7 +10,7 @@ import (
 )
 
 // Note: this must be updated with each new OpenShift release.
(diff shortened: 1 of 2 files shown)


jvdm (approved):
LGTM.

PR #1078, opened 2023-02-01
ROX-14288: Ensure Linux Kernel related packages are not skipped in RHCOS node scanning

jvdm:
Allowlist `linux` and `kernel` packages when matching vulnerability when the namespace is RHCOS.

Scanner explicitly skips packages prefixed with "linux" or "kernel" to ensure kernel-related vulnerability packages are not reported, as they are irrelevant for images. This is not the case for nodes. We should ensure these are noticed in RHCOS nodes.

## Tests

- UTs.

### DIFF
--- a/api/v1/models.go
+++ b/api/v1/models.go
@@ -95,9 +95,16 @@ func featureFromDatabaseModel(dbFeatureVersion database.FeatureVersion, uncertif
 	}
 }
 
-func hasKernelPrefix(name string) bool {
+// isPackageBlacklistedInScan returns true if the given package name in the
+// namespace should be skipped in vulnerability matching scans.
+func isPackageBlacklistedInScan(namespace, pkgName string) bool {
+	// Allow kernel packages in RHCOS.
+	if namespaces.IsRHCOSNamespace(namespace) {
+		return false
+	}
+	// Blacklist kernel prefixes everywhere else.
 	for _, prefix := range kernelPrefixes {
-		if strings.HasPrefix(name, prefix) {
+		if strings.HasPrefix(pkgName, prefix) {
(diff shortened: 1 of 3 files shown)


api/v1/models.go
```diff
@@ -95,9 +95,16 @@ func featureFromDatabaseModel(dbFeatureVersion database.FeatureVersion, uncertif
 	}
 }
 
-func hasKernelPrefix(name string) bool {
+// isPackageBlacklistedInScan returns true if the given package name in the
+// namespace should be skipped in vulnerability matching scans.
+func isPackageBlacklistedInScan(namespace, pkgName string) bool {
```
RTann, line 100: nit: let's change the name to more inclusive language. I think we have been using `blocklist`

api/v1/models_rhelv2_test.go
```diff
@@ -277,3 +278,68 @@ func TestComponentsFromDatabaseModelRHELv2(t *testing.T) {
 	assert.Equal(t, expectedRHELv2PkgEnvs, imgComponents.RHELv2PkgEnvs)
 	assert.ElementsMatch(t, expectedNotes, imgComponents.Notes)
 }
+
+func Test_getFullRHELv2Features(t *testing.T) {
```
RTann, line 282: can we just add a test to also ensure we skip it for non-rhcos?

RTann (approved):

api/v1/models_rhelv2_test.go
jvdm: Oh, I saw your comment post-merge. I will follow up on the UT in a separate PR.

PR #1079, opened 2023-02-02
support OpenShift 4 on RHEL 9 CPEs

RTann:
Happened to notice this in the logs. Currently, we just simply print a warning about this, and we are not notified.

Options:

1. Error out - I don't like this because this blocks everything from getting updated due to one feed's errors
2. Send out a notification - This may be best. Idea is to utilize the ideas in https://github.com/stackrox/scanner/blob/master/scripts/ci/logcheck/check-logs.sh to check the genesis dump CI job's logs and notify the team either via some kind of failure or slack message

Created ROX-14795 for this.


jvdm:
@RTann:

> Happened to notice this in the logs. Currently, we just simply print a warning about this, and we are not notified.

Which logs did you find the errors and what is the issue or symptoms?

RTann:
> @RTann:
> 
> > Happened to notice this in the logs. Currently, we just simply print a warning about this, and we are not notified.
> 
> Which logs did you find the errors and what is the issue or symptoms?

sorry just noticed this. I found it in the `generate-genesis-dumps` logs. You can still find it in any `genesis-dump` log in a CI step which actually runs it (master and PRs using the label). For example: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/branch-ci-stackrox-scanner-master-merge-push-images/1623430584304930816/artifacts/build-logs/genesis-dump.log

Not adding this will mean we may not correctly match vulnerabilities for OpenShift images running on RHEL 9. Luckily, this just affects OpenShift 4.13 (I think) at this time, and 4.13 is not yet released.

jvdm (approved):

PR #1081, opened 2023-02-06
Added new readme file

daynewlee:
Added a readme file so people can reproduce the steps of benchmark node analyze 

Added a parameter to specs generation script so people can decide how many rpm files they need

Added specs generation script to dockerfile

Uploaded test image to quay RHACS-eng/sandbox so it can be kept reusing


jvdm (changes requested):
To get a README file, I think you should:

1. Rename to `README.md`.
2. Reformat the content to be [github's flavored markdown](https://github.github.com/gfm/).

GitHub will render the contents in HTML when opening the directory in their UI, which is nice.

scripts/benchmark_node_analyze_readme.txt
```diff
@@ -0,0 +1,23 @@
+Steps to do the node scanning benchmark
```
jvdm, line 1: Maybe use a generic title, since this is a README for all the scripts.

scripts/benchmark_node_analyze_readme.txt
```diff
@@ -0,0 +1,23 @@
...
+
+Pre steps:
+
+    1.Make sure docker is installed on your local laptop
+    2.execute docker run -it --rm registry.access.redhat.com/ubi8/ubi
+    3.make sure Vim, RPM and RPM-build are installed in the running ubi8 bash
```
jvdm, line 7: Vim is not actually required.

scripts/benchmark_node_analyze_readme.txt
```diff
@@ -0,0 +1,23 @@
...
+        if not, do yum search <package name> and yum install <package name>
```
jvdm, line 8: Does it worth creating a `Dockerfile` to generate this test image and upload it to the public-facing quay repo? I am thinking of a test image. We wouldn't need to execute this setup to run the tests; one step towards automating it.

scripts/benchmark_node_analyze_readme.txt
```diff
@@ -0,0 +1,23 @@
...
+        Modify the for loop in the python code for the number of specs you want to create
+    2.execute the bash file: bash generate-rpm-specs.sh <number of the specs> e.g bash generate-rpm-specs.sh 100
+        You will be able to see the RPM spec files in the folder
+    3.execute rpmbuild -bb *.spec to create RPM files based on all generated spec files
+        You will see output like: Wrote: /root/rpmbuild/RPMS/x86_64/package_name9-0-0.x86_64.rpm Obviously this tells you where the rpm files are
+    4.install all packages using the rpm files just created: yum localinstall <path to rpm files>/.*rpm  e.g yum localinstall /root/rpmbuild/RPMS/x86_64/.*rpm
+    5.check the running ubi containers by executing: docker container ls and get the container id of the container where you are currently installing the rpm packages
+    6.execute docker export [container id]>[tar name].tar (e.g docker export c8c57bb7e926>demoV1.tar) and unzip the tar ball to get the file system
+    7.Go to github.com/stackrox/scanner/blob/master/benchmarks/analyzeNode folder and make sure to add local path to file system in line 13, then execute command: go test -bench=. -benchmem to run the test
```
jvdm, line 22: Can we have a `Makefile` target to do that?  It would be easier to re-run `go test` with the correct parameters.

scripts/generate-rpm-specs.sh
```diff
@@ -4,8 +4,8 @@
 # spec => rpm files: run rpmbuild -bb *.spec
 # install rpm: yum localinstall /root/rpmbuild/RPMS/x86_64/*.rpm
 
-for n in {1..100}; do
-    cat >pkg-$(printf "%03d" ${n}).spec <<EOF
+for ((n=1; n<=$1; n++)); do
```
jvdm, line 7: Why are we changing this?

jvdm (changes requested):

scripts/generate-rpm-specs.sh
jvdm, line 7: I suggest we fail if `$1` is not specified.

```suggestion
pkgcount=${1:?missing package count}
for n in $(seq 1 $pkgcount); do
```

scripts/benchmark_node_analyze_readme.txt
daynewlee: Yeah, I agree with the idea of creating a Dockerfile in quay repo.

benchmarks/analyzeNode/analyze_node_test.go
```diff
@@ -17,7 +17,7 @@ func runBenchmarkAnalyzeNode(b *testing.B, pathName string) {
 	runtime.GC()
 
 	for i := 0; i < b.N; i++ {
-		node.Analyze("testNode", pathName, node.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: true})
+		node.Analyze("testNode", pathName, node.AnalyzeOpts{UncertifiedRHEL: false, IsRHCOSRequired: false})
```
RTann, line 20: curious why this changed

scripts/benchmark_node_analyze_readme.txt
RTann: not 100% sure we have public-facing images. Perhaps we can start this? I believe there is a `quay.io/stackrox-io/qa` repo, but I don't think it's populated

benchmarks/analyzeNode/analyze_node_test.go
daynewlee: If IsRHCOSRequired is true, it won't do the analyze on non-core os node like ubi8 or 9

scripts/benchmark_node_analyze_readme.txt
Maddosaurus: Having a minimal test image with all prereqs would be a great idea!
That could also make it easier to automate. In terms of registries, let's start this.

@RTann are you maybe thinking of `https://quay.io/repository/rhacs-eng/qa`? That one isn't public, though.
But we could set up a `qa` or `testing-infrastructure` repo that could hold this kind of images.

scripts/benchmark_node_analyze_readme.txt
Maddosaurus: I would also like to pitch rephrasing these instructions a bit to make them more actionable.
They don't necessarily need to be copy-paste-able, but it would be really cool if they would be as close to that as possible.

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
+# Steps to benchmark node scanning 
+
+## Pre steps:
+
+1.Make sure docker is installed on your local laptop
```
Maddosaurus, line 5:
```suggestion
1. Make sure docker is installed on your local laptop
```
Thank you for converting to Markdown! Great to see the renderer picking it up.
Looking at [the rendered file](https://github.com/stackrox/scanner/blob/01bfda2267e65abe911dd949b49d0c93022279b0/scripts/node/README.md), it seems that the renderer didn't pick up the ordered list.
I suspect this is because there are missing whitespaces between the number and the text ([see also: Spec](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists)). See my change for an example. 

I recommend this to be done for all list items in this file.
Maybe we should also capitalize the first word of each list item.

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
...
+2.execute
+```
+docker run -it --rm registry.access.redhat.com/ubi8/ubi
```
Maddosaurus, line 8: A docker run alone might work now, but only because the image has a default entrypoint which is bash.
How about we make this more explicit?
```suggestion
docker run -it --rm --entrypoint /bin/bash registry.access.redhat.com/ubi8/ubi
```

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+3.make sure **Vim, RPM and RPM-build** are installed in the running ubi8 bash. If not, do yum search <package name> and yum install <package name>
+
+---------------------------------------------------------------------------------------------
+
+(optional) create a temporary/target folder in the running ubi8 bash for RPM specs. e.g mkdir temp-specs
+1.go to the target folder and create a python file to generate RPM specs (see the generate-junit-reports.sh in scanner/scripts folder)
```
Maddosaurus, line 15: How about we mount the host folder that contains this script instead of instructing the user to create it?
Also, is it a python file or are you talking about `generate-rpm-specs.sh` here?

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+2.execute the bash file: bash generate-rpm-specs.sh <number of the specs>  . You will be able to see the RPM spec files in the folder
+```
+bash generate-rpm-specs.sh 100
+```
+3.execute
+```
+rpmbuild -bb *.spec
+```
+You will see output like: Wrote: /root/rpmbuild/RPMS/x86_64/package_name9-0-0.x86_64.rpm Obviously this tells you where the rpm files are
```
Maddosaurus, line 27:
```suggestion
You will see output like: 
`Wrote: /root/rpmbuild/RPMS/x86_64/package_name9-0-0.x86_64.rpm`
This gives you the path where RPM files were generated to.
```

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
...
+4.install all packages using the rpm files just created: yum localinstall <path to rpm files>/.*rpm  
+```
+yum localinstall /root/rpmbuild/RPMS/x86_64/.*rpm
+```
+5.check the running ubi containers by executing: docker container ls and get the container id of the container where you are currently installing the rpm packages
+6.execute docker export [container id]>[tar name].tar (e.g docker export c8c57bb7e926>demoV1.tar) and unzip the tar ball to get the file system
```
Maddosaurus, line 33: What parts of the extracted filesystem are we interested in? Maybe we should tell the reader here.

scripts/node/README.md
```diff
@@ -0,0 +1,37 @@
...
+7.Go to **github.com/stackrox/scanner/blob/master/benchmarks/analyzeNode** folder and make sure to add local path to file system in line 13, then execute command:
```
Maddosaurus, line 34: If it is a single point in the code where we provide this path, how about we either hardcode it and tell the user where to put the required files, or we use an environment variable?

Maddosaurus (changes requested):
Thanks for the update!
I think we need another pass to improve clarity and readability in the readme.
Please find my comments below.

scripts/node/dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi
+
+RUN yum -y install rpm-build rpm-libs rpm-build-libs rpm
+RUN yum -y install vim-common vim-enhanced vim-filesystem
+
```
jvdm, line 5: Can we `COPY`  the `generate-rpm-specs.sh` into the image and get rid of Step 1. in the `README.md`?

scripts/node/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+4.install all packages using the rpm files just created: yum localinstall <path to rpm files>/.*rpm  
+```
+yum localinstall /root/rpmbuild/RPMS/x86_64/.*rpm
+```
+5.check the running ubi containers by executing: `docker container ls` and get the container id of the container where you are currently installing the rpm packages
+
+6.execute `docker export [container id]>[tar name].tar` (e.g `docker export c8c57bb7e926>demoV1.tar`) and unzip the tar ball to get the file system
+
+7.Go to **scanner/blob/master/benchmarks/analyzeNode** folder and make sure to add local path to file system in line 13, then execute command:
```
jvdm, line 39: I believe you need [a space between `.` and the beginning of the list content](https://github.github.com/gfm/#lists).

Also, for all list items, I think the list contents are English sentences, and AFAIK, it would be better style to Capitalize the first letter.

scripts/node/README.md
```diff
@@ -0,0 +1,42 @@
+# Steps to benchmark node scanning 
+
+## Pre steps:
+
+quay.io/repository/rhacs-eng/sandbox:benchmark-node-analyze-1.0 is an Ubi8 based image with vim, RPM and RPM-build installed. Those packages are necessary for benchmark node analyze function
+
+back-up image: `quay.io/rh_ee_yli3/nodes:benchmark-node-analyze-1.0`
```
jvdm, line 7: What is the purpose of the backup image?

Should we move it from the `yli3` namespace into something generic?

scripts/node/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+You will see output like: `Wrote: /root/rpmbuild/RPMS/x86_64/package_name9-0-0.x86_64.rpm` This gives you the path where RPM files were generated to.
+
+4.install all packages using the rpm files just created: yum localinstall <path to rpm files>/.*rpm  
+```
+yum localinstall /root/rpmbuild/RPMS/x86_64/.*rpm
+```
+5.check the running ubi containers by executing: `docker container ls` and get the container id of the container where you are currently installing the rpm packages
+
+6.execute `docker export [container id]>[tar name].tar` (e.g `docker export c8c57bb7e926>demoV1.tar`) and unzip the tar ball to get the file system
```
jvdm, line 37: Nitpick: Very unlikely to happen, but better to have a space between `[container id]` and `>`: if the ID is all numeric and small enough, the shell might believe you are referring to a file descriptor instead.

jvdm (approved):

Maddosaurus (approved):
Thank you for the changes!
Do we know why the e2e tests are failing?

daynewlee:
> Thank you for the changes! Do we know why the e2e tests are failing?

I will investigate that today.

PR #1082, opened 2023-02-06
Test_getFullRHELv2Features compare slices with assert.ElementsMatchf

RTann:
assert.Equalf works iff the order is the same. Use  assert.ElementsMatchf so we don't have to care about the ordering


RTann:
ignore the failing tests, they are unrelated

dcaravel (approved):
/LGTM

PR #1084, opened 2023-02-07
Reduce race condition likelihood, remove redundant diff download

dcaravel:
This change will reduce (not eliminate) the likelihood of a race condition that occurs when the `definitions.stackrox.io` GCS bucket is updated after the `sanity-check-vuln-updates` job has started.  It does this by using the `age` of the GCS `diff.zip` as close to the 1hr age check as possible (instead of the `age` determined when the job started)

The CloudFlare `age` header is now also logged for future troubleshooting / logic as necessary, it will help determine the actual cache age (for the particular CloudFlare node)
    
Additionally, each diff was being downloaded multiple times (back to back), once for extracting a manifest, and the other for extracting headers.  This PR removes the redundant download (`wget`) which should reduce bandwidth and potentially speed up the job

### DIFF
--- a/scripts/ci/jobs/sanity-check-vuln-updates.sh
+++ b/scripts/ci/jobs/sanity-check-vuln-updates.sh
@@ -88,39 +88,53 @@ function run_tests_for_diff_id {
   local diff_id gsutil_last_update
   local url_gsutil url_cloudflare url_gcs_https
   local metadata_cloudflare metadata_gcs_https
-  local cache_control_cloudflare cache_control_gcs_https
+  local cache_control_cloudflare cache_control_gcs_https cache_age_cloudflare_secs
   local md5sum_cloudflare md5sum_gcs_https
+  local gcs_object_age_seconds gcs_object_age_secs_now update_time_epoc_secs 
+  local curl_verbose_cloudflare
+  local diff_archive_cloudflare diff_archive_gcs
 
   diff_id=${1:-0133c2cf-8abe-4d79-9250-9b64b5b3e43e}
 
   url_cloudflare="https://definitions.stackrox.io/$diff_id/diff.zip"
   url_gcs_https="https://storage.googleapis.com/definitions.stackrox.io/$diff_id/diff.zip"
 
-  metadata_cloudflare=$(wget -q "$url_cloudflare" && unzip -q -c diff.zip manifest.json | jq -cr '.' && rm -f diff.zip) \
-    || bash_exit_failure "curl failed on $url_cloudflare"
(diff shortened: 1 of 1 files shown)


scripts/ci/jobs/sanity-check-vuln-updates.sh
```diff
@@ -88,39 +88,55 @@ function run_tests_for_diff_id {
...
-  metadata_cloudflare=$(wget -q "$url_cloudflare" && unzip -q -c diff.zip manifest.json | jq -cr '.' && rm -f diff.zip) \
-    || bash_exit_failure "curl failed on $url_cloudflare"
-  metadata_gcs_https=$(wget -q "$url_gcs_https" && unzip -q -c diff.zip manifest.json | jq -cr '.' && rm -f diff.zip) \
-    || bash_exit_failure "curl failed on $url_gcs_https"
+  diff_archive_cloudflare="$WORKING_DIR/diff1.zip"
+  diff_archive_gcs="$WORKING_DIR/diff2.zip"
 
-  cache_control_cloudflare=$(curl -s -H 'Accept-encoding: gzip' -o /tmp/diff1.zip \
-    -v "$url_cloudflare" 2>&1 | grep "cache-control" | sed -e "s#^< ##g; s#\r##g;") \
+  curl_verbose_cloudflare=$(curl -s -H 'Accept-encoding: gzip' -o "$diff_archive_cloudflare" \
```
RTann, line 105: nit: maybe just `verbose_cloudflare`?

scripts/ci/jobs/sanity-check-vuln-updates.sh
```diff
@@ -88,39 +88,55 @@ function run_tests_for_diff_id {
...
+    -v "$url_cloudflare" 2>&1 | sed -e "s#^< ##g; s#\r##g;") \
```
RTann, line 106: curious is this `sed` still useful?

RTann (approved):
LGTM with two optional comments

scripts/ci/jobs/sanity-check-vuln-updates.sh
dcaravel: Raw output from `curl -v ...` would look something like this:

```
...
* Connection state changed (MAX_CONCURRENT_STREAMS == 256)!
< HTTP/2 200 
< date: Thu, 09 Feb 2023 18:42:49 GMT
< cache-control: public, max-age=3600
< cf-cache-status: HIT
< age: 777
< accept-ranges: bytes
< vary: Accept-Encoding
< server: cloudflare
...
```
If that `sed` is removed we would have to change the header parsing lines to look for `< age` or`< cache-control` instead of `age` or `cache-control`.

Suppose one could argue if there was an body to this request the `< ` is meaningful to determine that the headers are headers and not part of the body, but thats not the case with these downloads.

PR #1090, opened 2023-02-08
Disable safe directory checks in git on CI

jvdm:
The checks were introduced in Git 2.28.0. This message appears when a repository's ownership is unclear and Git attempts to warn the user about this issue. After the latest build image upgrade to bump to go-1.9, CI started failing due to this check.

This change is a mirrored change of the [steps taken in stackrox/stackrox](https://github.com/stackrox/stackrox/commit/48fe67a44e1e064898e288bd0d82d9204bea701d).

## Follow up

Build targets in Makefile might also need an update. This will be addressed in a separate PR after some testing.

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -545,6 +545,13 @@ openshift_ci_mods() {
     info "Env A-Z dump:"
     env | sort | grep -E '^[A-Z]' || true
 
+    ensure_writable_home_dir
+
+    # Prevent "detected dubious ownership in repository" errors introduced in
+    # Git 2.28.0.  This message appears when the ownership of a repository is
+    # not clear.
+    git config --global --add safe.directory "$(pwd)"
+
     info "Git log:"
     git log --oneline --decorate -n 20 || true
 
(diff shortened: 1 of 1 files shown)


dcaravel (approved):
Changes /lgtm - since it worked [for stackrox repo](https://github.com/stackrox/stackrox/pull/4459) don't see why it wouldn't work here.

Assuming the plan is to let CI run to determine if `podman` / `docker --user` [related changes](https://github.com/stackrox/stackrox/pull/4622) made to `stackrox` also affect scanner? Looks like initially they were suspected to be related.

jvdm:
> Changes /lgtm - since it worked [for stackrox repo](https://github.com/stackrox/stackrox/pull/4459) don't see why it wouldn't work here.

Well, we got a green CI run in this PR; that's the confirmation we need, I hope.

> Assuming the plan is to let CI run to determine if `podman` / `docker --user` [related changes](https://github.com/stackrox/stackrox/pull/4622) made to `stackrox` also affect scanner? Look like, initially they were suspected to be related.

They are partially related. It might affect when building images locally, in your workstation. I decided to test it first and maybe follow up in a PR before holding CI on this orthogonal problem. I originally left a note in the PR description.

PR #1093, opened 2023-02-15
ROX-15044: Bump nvtools to fix NVD CPE version padded with zeros

jvdm:
This is a follow up on https://github.com/stackrox/nvdtools/pull/8

When numeric parts are compared, 0's are padded to the left to ensure they have the same length. When all components are the same, the code relies on the string length, but the padded zeros are ignored, resulting in incorrect results.

## Tests

CI


dcaravel (approved):
LGTM

PR #1106, opened 2023-02-28
Ignore "Red Hat Enterprise Linux must be installed" criterion in RHELv1

RTann:
The "genesis-dump" logs are riddled with this log everywhere. Let's just ignore this criterion, as it doesn't help us out, and it not an actual issue.


dcaravel (approved):
LGTM, noticed this can be tested locally by running the `./bin/updater generate-dump ...` per the Readme

PR #1110, opened 2023-03-01
Remove envisolator & update StackRox dep

Maddosaurus:
As we needed to update the stackrox/rox dep in Scanner, envisolator is not available anymore.
This PR removes the envisolator and updates stackrox to latest master.

This is a prerequisite to working on ROX-13934, for which we need the new `NodeInventory` storage types in the StackRox library (see draft PR https://github.com/stackrox/scanner/pull/1111)


RTann (approved):

Maddosaurus:
Looking at the failing e2e-tests, it seems that the `FixedBy` field changed.
I suspect this has not been caused by my changes, but is due to newer data, correct?

vikin91:
> Looking at the failing e2e-tests, it seems that the `FixedBy` field changed. I suspect this has not been caused by my changes, but is due to newer data, correct?

This happens in the Scanner repo sometimes. Good to merge for me, but let's wait for Scanner colleagues opinion.

Maddosaurus:
See also [this Slack discussion](https://redhat-internal.slack.com/archives/C033Z8KMZAM/p1677688373921949).

PR #1116, opened 2023-03-03
ROX-13934: Scanner running in Node Inventory mode

fredrb:
This PR introduces the first portion of [ROX-13934](https://issues.redhat.com/browse/ROX-13934): implement scanner to run in a "Node Inventory" mode.

This will enable the compliance pod to communicate with Scanner (in Node Inventory mode) via the pod local host network.

```
    SENSOR POD      |                      COLLECTOR POD                 |
sensor <------------|-------------> [compliance] <---> [node-scanner] <--|---> NODE HOST
```

Even though there are a lot of lines added, it's not that big. The code under `pkg/nodeinventory/inventorizer.go` and `pkg/nodeinventory/inventorizer_test.go` was copied (and the types changed accordignly) from [here](https://github.com/stackrox/stackrox/blob/master/compliance/collection/nodeinventorizer/nodeinventory.go). 

**Important parts:**

1. Communication is done via a gRPC API using `scannerV1` interfaces. These are known to ACS and they will be parsed and converted to the `storage.NodeInventory` structure in Compliance. 
2. This *doesn't introduce a new binary*. Although this was the preferred option in the beginning, we decided to use the same binary as Scanner. This means that scanner has a new scanner binary flag (`--nodeinventory`), and it will start in Node Inventory mode. And that no changes are needed downstream 🎉 .
3. We don't need TLS, because compliance will communicate with node inventory scanner through local host. This PR introduces a way to start the `gRPC` server in Scanner *without* the public facing TLS endpoint. I.e.: only the local server at `8444` is started. 
4. There is one new environment variable needed: `ROX_NODE_NAME`. It should be injected by Kubernetes when running in the Collector pod, like the following: 
```yaml
        env:
        - name: ROX_NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
```

---

**Note to reviewers:**

There are two things that I'm not sure about this change, and your feedback is welcome:

(1) **Unifying with Scanner boot process**

Right now, the simplest approach was to fork at the very beginning of `main()` if the flag is set. However, there might be some advantages to unifying a _little_ with Scanner boot process. For example, reusing `config.yaml` to set log level and port configuration. 

This comes with some drawbacks, like having to mount a `config.yaml` in the collector pod as well. 

I would go with the simplest option of keeping it completely separated, but maybe there is something important that I'm missing here.

(2) **Naming**

As usual, naming things isn't easy, and once this is commited, they will probably stick for a long time. 

Initially, I called this feature "Node Scanner". But it was only until I had to create the gRPC API that I realized that this name conflicts with the already existing `NodeScanService`. Having a `NodeScannerService` and `NodeScanService` sound very confusing. 

So my proposal is to name this `NodeInventoryService`. Let me know what do you think about it. 

---

## Testing

After the image is built, you can simply run it in docker like:

```
$ docker run -it --entrypoint /scanner scanner-slim:TAG --nodeinventory
INFO[0000] Starting Scanner in Node Inventory mode
INFO[0000] Launching backend GRPC listener on 127.0.0.1:8444
```


api/v1/nodeinventory/service.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Service defines the node scanning service.
+type Service interface {
+	apiGRPC.APIService
+
+	v1.NodeInventoryServiceServer
+}
+
+// NewService returns the service for node scanning
+func NewService(nodeName string) Service {
```
vikin91, line 23: I usually return the impl-type from the constructor to allow the impl type to potentially satisfy many interfaces (also the future ones). Was there a reason to bind this constructor to the interface?

```suggestion
func NewService(nodeName string) serviceImpl {
```

api/v1/nodeinventory/service.go
```diff
@@ -0,0 +1,59 @@
...
+type serviceImpl struct {
+	inventoryCollector *nodeinventory.Scanner
+	nodeName           string
+}
+
+func (s *serviceImpl) GetNodeInventory(ctx context.Context, req *v1.GetNodeInventoryRequest) (*v1.GetNodeInventoryResponse, error) {
+	inventoryScan, err := s.inventoryCollector.Scan(s.nodeName)
+	if err != nil {
+		log.Errorf("Error running inventoryCollector.Scan(%s): %v", s.nodeName, err)
+		return nil, errors.New("Internal scanner error: failed to scan node")
```
vikin91, line 39: I think the process in which we build an inventory is called "node analysis"  in the scanner team. We also call it inventorying but it is something introduced only recently.

proto/scanner/api/v1/node_inventory_service.proto
```diff
@@ -0,0 +1,30 @@
...
+    string                      node_name   = 1;
+    Components                  components  = 2;
+    repeated Note               notes       = 3;
+}
+
+message GetNodeInventoryRequest {}
+
+// NodeInventoryService is used in Secured Clusters to fetch information from Nodes and communicate with other
+// Secured Cluster components, like the compliance container in Collector.
+service NodeInventoryService {
```
vikin91, line 24: (Answering to Note to reviewers) I like this name

cmd/clair/main.go
```diff
@@ -201,12 +203,36 @@ func Boot(config *Config, slimMode bool) {
...
+	// Wait for interruption and shutdown gracefully.
+	waitForSignals(os.Interrupt, unix.SIGTERM)
+	log.Info("Received interruption, gracefully stopping ...")
+}
+
 func main() {
 	// Parse command-line arguments
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	flagConfigPath := flag.String("config", "/etc/scanner/config.yaml", "Load configuration from the specified file.")
+	flagNodeInventoryMode := flag.Bool("nodeinventory", false, "Run Scanner binary in Node Inventory mode (this should only be used in Secured Clusters)")
```
vikin91, line 227: As we are ignoring the config while `nodeinventory` is set to true: maybe we should show a warning when the user tries to provide both flags nodeinventory and config?

vikin91: On skipping config: I see that we will use unconfigured logger when we skip reading the config file here. This is not a big issue as the behavior will be consistent within the nodeInventory mode, but maybe we want to do the configuration, as scanner uses structured logs?

Makefile
```diff
@@ -47,7 +47,7 @@ LOCAL_VOLUME_ARGS := -v$(CURDIR):/src:delegated -v $(GOPATH):/go:delegated
 GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS := -e CGO_ENABLED=0 -e GOOS=linux -e GOARCH=${GOARCH}
 BUILD_FLAGS := CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH}
-BUILD_CMD := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -o image/scanner/bin/scanner ./cmd/clair
+BUILD_CMD := go build -buildvcs=false -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -o image/scanner/bin/scanner ./cmd/clair
```
vikin91, line 50: I couldn't find any references to `buildvcs` in the scanner repo. Why is it here?

vikin91 (commented):
Sharing my observations

Makefile
fredrb: Sorry -- this was just a local fix in my environment. Commited accidentally

api/v1/nodeinventory/service.go
fredrb: That's how all other gRPC services are defined. E.g. `api/v1/imagescan/service.go` or `api/v1/orchestratorscan/service.go`. 

I think the purpose of this abstraction is to have a single struct impl of the registered gRPC API. I'd prefer to keep it like this to comply with current abstraction.

api/v1/nodeinventory/service.go
fredrb: Is there a suggestion here? Should it say `Internal scanner error: failed to analyze node` instead?

cmd/clair/main.go
fredrb: Yea, this is my point (2) for reviewers on the PR description. I think it does make sense that some startup routines are shared between the original scanner mode and node inventory. But I didn't want to refactor too much without Scanner's team feedback on this. 

I agree it might make sense to _at least_ start the logger for both. However, the log level is defined in the `config.yaml`. Which means that in node inventory mode we'd also rely on the config file, even if most of the properties aren't used.

proto/scanner/api/v1/node_inventory_service.proto
Maddosaurus: +1 - I like the name as well

api/v1/nodeinventory/service.go
Maddosaurus: I think either would work. 
Changing to `analyze` would make it more consistent with the language the scanner team uses.
But as this error would eventually make its way back to Compliance as a return, it is likely to be shown to users. 
Therefore, maybe it would be the right call to stick to `scan`, as we call this feature `node scanning` in almost all user facing log messages.
I would tend to lean towards keeping it as `scan`, but that's non-blocking for me.

cmd/clair/main.go
Maddosaurus: Out of curiosity: 
Could we initialize the logger with defaults and overwrite them if there is a config found (later on)?
That way, we wouldn't be depending on scanners `config.yaml` but still act on it, e.g. when the log level is set to debug by someone.

Maddosaurus (commented):
All comments from my side are non-blocking. I'd be fine with accepting this, but I'd be interested in hearing the opinion of the scanner team beforehand 😄 

With regards to the failing e2e test: The `FixedBy` field differs in these tests. That's something we see from time to time, that is no reason to worry for this PR.

To the questions:
1. I agree that unifying this a little could be beneficial. Could we introduce defaults for all settings that we share/take from `config.yaml` so that Node Scanning could still run even when the yaml is not provided? Could that be a good compromise?
2. I am in favour of keeping it `NodeInventoryService`, that's the most precise naming we have so far.

cmd/clair/main.go
vikin91: The logger has its defaults I think, so there is no danger here. Maybe we can configure the logger additionally (when `flagNodeInventoryMode` is true) with the default values from the config?

I do not like that starting the nodeInventoryMode would require a config file... I think we should avoid that.

api/v1/nodeinventory/service.go
vikin91: Options like:
- failed to analyze node
- failed to build node inventory

would be okay for me. 

I think Scanner team uses additionally terms like: analyzing (scanning disk to search for pkgs), enriching (adding vulnerabilities data to the inventory). We never fully synced our dictionaries, so I guess we can also keep it the way it is.

Maddosaurus:
After the latest iteration of changes we removed that custom entrypoint script, correct?
I just tried running the image and ended on `docker run -it --entrypoint /scanner scanner-slim:TAG --nodeinventory` - is that the intended invocation of the new code path?

cmd/clair/main.go
```diff
@@ -201,12 +203,36 @@ func Boot(config *Config, slimMode bool) {
 	serv.Close()
 }
 
+func bootNodeInventoryScanner() {
+	// TODO: Check: Unify with Scanner boot and run as a different configuration rather than a different function.
```
jvdm, line 207: Nit: Not clear what this TODO means. If we use the same config for both the regular scanner and `NodeInventory=true` scanner, will this still be necessary?

Maybe better to create a follow-up task instead of a TODO in the code.

jvdm (changes requested):
I am only suggesting to re-use the configuration. I believe that the changes to the deployment templates to support the config are straightforward.  I acknowledge that they might be annoying. But still, having the config here allows us to fork to `nodeinventory` mode later in `main()`, and re-use some checks that a worthwhile IMO.

I also think the config has some reasonable defaults, by the way. So I think we can deploy without it, test, run demos, etc. The config entries we use for `nodeinventory` should not impact tests and demos. It does impact production because it would not allow us to change the log level. Not sure if it's a good idea to release without this toggle.

My other comments are just reminders to check things and clarify some conversations. A nit on the TODO that you can ignore if you want to.

cmd/clair/main.go
fredrb: That makes sense. Moved the inventory mode startup. Thanks!

cmd/clair/main.go
```diff
@@ -180,10 +182,10 @@ func Boot(config *Config, slimMode bool) {
...
 	go api.RunClairify(serv)
 
-	grpcAPI := grpc.NewAPI(grpc.Config{
-		Port:         config.API.GRPCPort,
-		CustomRoutes: debugRoutes,
-	})
+	grpcAPI := grpc.NewAPI(
+		grpc.WithTLSEndpoint(config.API.GRPCPort),
+		grpc.WithDefaultInterceptors(),
+		grpc.WithCustomRoutes(debugRoutes))
```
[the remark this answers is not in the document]
fredrb: `grpc.NewAPI` is being called from both Node Inventory Mode and Scanner. However, they need different configurations of the GRPC server. 

1. Scanner requires TLS, Node inventory doesn't
2. Scanner requires a peer certificate interceptors, Node inventory doesn't

I thought it would be more idiomatic to use name opts function (e.g. `WithTLSEndpoint`) rather than passing multiple parameters or a struct with many properties.

Since I changed the `NewAPI` signature and pulled some dependencies outside (e.g. interceptors), I had to change how the grpc server is started on scanner path as well.

cmd/clair/main.go
fredrb: Removed the TODO. Your comment above about starting node inventory mode after setting up default config already solves this question.

pkg/env/list.go
```diff
@@ -19,4 +19,8 @@ var (
 	// This is ignored if SkipPeerValidation is enabled.
 	// This variable was copied over from the stackrox repo.
 	OpenshiftAPI = RegisterBooleanSetting("ROX_OPENSHIFT_API", false)
+
+	// NodeName is used when running Scanner in Node Inventory mode. This should be set by
+	// Kubernetes in the Secured Cluster.
+	NodeName = RegisterSetting("ROX_NODE_NAME", WithDefault("unset"))
 )
```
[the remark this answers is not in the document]
fredrb: I agree with that. Had a chat with Piotr this morning and we think it's better to crash it with an error message. Because if the node name isn't set, or it's set to a dummy value, Sensor will not be able to link the inventory message with the v1.Node message that comes from K8s. RHCOS will *silently fail*, and users will have to look through the logs to figure out that an env was not set. 

For that reason, it makes more sense to fail fast and crash node inventory with an error message saying that we need `ROX_NODE_NAME` to be set to an unique node value, likely coming from Kubernetes.

jvdm (approved):
LGTM.

Please ensure Collector will gracefully handle the scenario where `node-inventory` is not offering the gRPC endpoint, maybe because `ROX_NODE_NAME` is not present or for any other reason.

PR #1118, opened 2023-03-13
Add context to `nodes.Analyze` to allow quitting early

vikin91:


pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,45 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
...
+	errC := make(chan error)
+	go func(ctx context.Context) {
+		defer close(errC)
+		errC <- filepath.WalkDir(dir, func(path string, info fs.DirEntry, err error) error {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+				return fn(path, info, err)
+			}
```
jvdm, line 125: You can avoid one more WalkDir() iteration and potentially one more file system call if you run `fn()` first, and check the context after. Most of the time, the context will be canceled while `fn()` is running. Good to avoid another call because if we are timing out, that call might hang for a while and you would have this goroutine running for longer unnecessarily.

```suggestion
			if fErr := fn(path, info, err); fErr != nil {
				return fErr
			}
			select {
			case <-ctx.Done():
				return ctx.Err()
			default:
				return nil
			}
```

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,45 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
...
+		})
+	}(ctx)
+
+	// Wait for the WalkDir to complete or for the context to be cancelled.
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
```
jvdm, line 132: It might be worth revisiting the [error log triggered by this](https://github.com/stackrox/scanner/pull/1118/files#diff-8757df52309cf15678c26e87d2de2e18ba065a1826da17959e169e4a323b9b0aR108). It mentions the files map could not be matched, but for timeouts, the timeout cancellation next to that message might be confusing.

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,45 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 	m := metrics.FileExtractionMetrics{}
 	// TODO(ROX-13771): Use `range matcher.GetCommonPrefixDirs()` again after fixing.
 	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm", "usr/share/buildinfo"} {
-		if err := n.addFiles(filepath.FromSlash(dir), matcher, &m); err != nil {
+		if err := n.addFiles(ctx, filepath.FromSlash(dir), matcher, &m); err != nil {
 			return nil, errors.Wrapf(err, "failed to match filesMap at %q (at %q)", dir, n.root)
```
jvdm, line 108:
```suggestion
			return nil, errors.Wrapf(err, "failed to extract files from %q at root %q", dir, n.root)
```

pkg/analyzer/nodes/node.go
vikin91: Got that covered! One more change was needed to prevent a goroutine leak.

pkg/analyzer/nodes/node.go
vikin91: Would sth like this be okay ?

```
node_test.go:469: Error: failed to extract files from "etc/" at root "/var/folders/.../001": operation took longer than expected and was interrupted: context deadline exceeded
```

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,57 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
...
+	}(ctx)
+
+	// Wait for the WalkDir to complete or for the context to be cancelled.
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-errC:
+		return err
+	}
+}
```
jvdm, line 148: Oh, good catch on the goroutine leak.

But now, all this code has become a little bit convoluted. Do you think we can work it out by buffering the channel?

I also removed passing the context to the anonymous goroutine and using the closure because I think it's easier to read it like that, but let me know if it won't work for any reason.

```suggestion
func walkDirWithContext(ctx context.Context, dir string, fn fs.WalkDirFunc) error {
	// Buffering so we can give up on reading the channel and not block the
	// child goroutine.
	errC := make(chan error, 1)

	go func() {
		defer close(errC)
		errC <- filepath.WalkDir(dir, func(path string, info fs.DirEntry, err error) error {
			if err = fn(path, info, err); err != nil {
				return err
			}
			select {
			case <-ctx.Done():
				return ctx.Err()
			default:
				return nil
			}
		})
	}()

	select {
	case <-ctx.Done():
		return ctx.Err()
	case err := <-errC:
		return err
	}
}
```

jvdm (changes requested):
Suggesting we should make the channel buffered instead of adding one additional `select` which made the code harder to read.

pkg/analyzer/nodes/node.go
vikin91: Yes, that looks good! Added in [9f1a280](https://github.com/stackrox/scanner/pull/1118/commits/9f1a28099d6e1318fd66e3edf21c6894f6f6fe29)

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,51 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
 	m := metrics.FileExtractionMetrics{}
 	// TODO(ROX-13771): Use `range matcher.GetCommonPrefixDirs()` again after fixing.
 	for _, dir := range []string{"etc/", "usr/share/rpm", "var/lib/rpm", "usr/share/buildinfo"} {
-		if err := n.addFiles(filepath.FromSlash(dir), matcher, &m); err != nil {
-			return nil, errors.Wrapf(err, "failed to match filesMap at %q (at %q)", dir, n.root)
+		if err := n.addFiles(ctx, filepath.FromSlash(dir), matcher, &m); err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				err = errors.Wrapf(err, "operation took longer than expected and was interrupted")
```
RTann, line 109: I'm not sure this really adds anything. The error message is already "context deadline exceeded" https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/context/context.go;l=171

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,51 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
...
 	}
 	m.Emit()
 	return n, nil
 }
 
+func walkDirWithContext(ctx context.Context, dir string, fn fs.WalkDirFunc) error {
+	// Buffering so we can give up on reading the channel and not block the
+	// child goroutine.
+	errC := make(chan error, 1)
+	go func() {
```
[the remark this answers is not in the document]
jvdm: > Why do we need this goroutine?

Isn't it because it allows the calling goroutine to give up early and not wait? For example, does your proposal handles the situation where `fn()` is blocked due to filesystem I/O, maybe over the network with potential high latency and timeouts?

pkg/analyzer/nodes/node.go
jvdm: Doesn't it give readers additional context on what precisely happened? In particular, that the file extraction was interrupted because it took too long to extract the files from a specific directory. Does "context deadline exceeded" convey the same information?

jvdm (approved):
LGTM. Thanks for the time iterating over it.

pkg/analyzer/nodes/node.go
RTann: I think "context deadline exceeded" already says the same thing as "operation took longer than expected and was interrupted". If this is to be used for logging purposes, then perhaps the caller can say something like: "Unable to complete analysis: <error>" or even customize the log if `errors.Is(err context.DeadlineExceeded)` since the caller knows this is a context with a deadline and this is a possible error

pkg/analyzer/nodes/node.go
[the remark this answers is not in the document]
RTann: ah I see what you mean

pkg/analyzer/nodes/node.go
```diff
@@ -103,22 +104,51 @@ func extractFilesFromDirectory(root string, matcher matcher.PrefixMatcher) (*fil
...
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+				return nil
+			}
+		})
+	}()
+	select {
+	case <-ctx.Done():
```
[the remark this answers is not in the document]
RTann: never mind I see why this is here now

pkg/analyzer/nodes/node_test.go
```diff
@@ -431,6 +434,47 @@ func Test_extractFilesFromDirectory(t *testing.T) {
 	}
 }
 
+func Test_extractFilesFromDirectory_Context(t *testing.T) {
+	defer goleak.VerifyNone(t)
```
RTann, line 438: let's say we did not buffer the channel. Would this check fail?

vikin91: Yes, it would in some cases. We had a code that was using unbuffered channel for this: see https://github.com/stackrox/scanner/pull/1118#discussion_r1136321437
We decided to apply buffer of 1 message to avoid one additional `select` and make the code more compact and easier to read.

Maddosaurus (approved):
LGTM.
The only thing I'm thinking about is if we can (or should) also take reading the RPM DB into consideration when cancelling the context. 
IIRC, it is a pretty atomic operation, so I'm not sure if it is possible.

FInally, interesting discussion around the channel usage, thanks everyone!

PR #1122, opened 2023-03-17
remove warning when content sets aren't detected

RTann:
Missing content_sets is logged for every image layer which does not contain content_sets. This PR removes the warning, as it is very normal for image layers to not contain content_sets.


pkg/analyzer/detection/detection.go
```diff
@@ -99,10 +99,7 @@ func detectAndAnnotateCertifiedRHELComponents(name string, files analyzer.Files,
 			logrus.WithError(err).Errorf("failed to get RHCOS content sets for %v: %v", namespace, err)
```
jvdm, line 99: Can you update this error line to communicate we will continue with empty content sets?

Alternatively, move the `Warningf()` into this `if` block.

jvdm (changes requested):

jvdm (approved):
LGTM.

PR #1123, opened 2023-03-21
ROX-13835: Define user experience when CPEs not found

jschnath:
Adds ~`CONTENT_SETS_UNAVAILABLE`~ `NODE_CERTIFIED_RHEL_CVES_UNAVAILABLE` note to Scanner API. We want to return this note when we ~discover no content sets~ cannot get CPEs from the inventory sent from compliance, so that the user is informed that the vulnerabilities are missing.

This PR effects API changes in stackrox/scanner required for a stackrox/stackrox PR [found here](https://github.com/stackrox/stackrox/pull/5336).


vikin91 (approved):
Looks good, but I was involved in writing this, so I may be biased.

@jvdm could you take a quick look?

jvdm (approved):
Folks, I and @vikin91 had an offline discussion. He did some modifications in the PR, I went ahead and fixed the note name and included a check for certified RHEL scans because this should only affect certified scans. The behavior now is:

1. If RHCOS and there are no content sets: Add a note that the node is unsupported. All supported RHCOS nodes should have content sets.
2. If certified RHEL and we could not get CPEs: We add a note to mark the node scan CVEs as unavailable.

I think the above covers https://issues.redhat.com/browse/ROX-13835

jschnath:
Thanks for having a look at the PR @jvdm @vikin91, if there are no objections I will be merging it as is soon.

PR #1132, opened 2023-03-23
dependabot update actions

RTann:
We can use dependabot to keep our actions up-to-date, so let's do it.

See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot for more information about it.


jvdm (approved):
LGTM.

PR #1133, opened 2023-03-23
remove unused import from node_inventory_service.proto

RTann:
I noticed the following would be printed upon generating protos:

```
scanner/api/v1/node_inventory_service.proto:7:1: warning: Import google/api/annotations.proto is unused.
```

Might as well remove the import since it is unused.


jvdm (approved):
LGTM, but I haven't checked if there is an actual reason for the `import weak` to be there. Curious to know if @Maddosaurus or @fredrb have something to say.

fredrb (approved):

fredrb:
Sorry, I missed this PR. 

> LGTM, but I haven't checked if there is an actual reason for the import weak to be there. 

This was copied from the other service definitions. E.g. `node_scan_service` also uses it to define the `option (google.api.http)`. 

I originally imported it because we had an HTTP endpoint definition for the API, but was removed in one of the later commit in the PR.

PR #1140, opened 2023-03-29
chore: log requests upon retrieval

RTann:
This came up when looking into potential network issues between Central and Scanner. These logs will let us know about all incoming traffic to the HTTP and gRPC endpoints which are exposed by Scanner.


api/grpc/logging_interceptor.go
```diff
@@ -0,0 +1,39 @@
...
+func loggingUnaryServerInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		peerInfo, exists := peer.FromContext(ctx)
+		if !exists {
+			return nil, status.Error(codes.InvalidArgument, "unable to parse peer information from request context")
+		}
+
+		logrus.WithFields(map[string]interface{}{
+			"Method": info.FullMethod,
+		}).Infof("Received gRPC request from %v", peerInfo.Addr)
```
dcaravel, line 26: I'm not an expert on grpc-gateway - but looks [like it supports](https://github.com/grpc-ecosystem/grpc-gateway/blob/f238f782038e2f05e1c238868b07db477fdf3782/runtime/context_test.go#L116) adding `X-Forwarded-For` header to the context 

Guessing should be able to extract the `Metadata` from the context, something like:
```
addrs := metadata.ValueFromIncomingContext(ctx, "X-Forwarded-For")
```
And then append the `addrs` to this log

dcaravel (approved):
I see the logs in the artifacts of the associated CI, made a few general comments.

Certainly an improvement.

api/v1/imagescan/service.go
```diff
@@ -203,11 +203,19 @@ func (s *serviceImpl) getImageComponents(_ context.Context, req *v1.GetImageComp
 }
 
 func (s *serviceImpl) GetImageVulnerabilities(_ context.Context, req *v1.GetImageVulnerabilitiesRequest) (*v1.GetImageVulnerabilitiesResponse, error) {
+	imageName := "unknown"
+	if req.GetImage() != "" {
+		imageName = req.GetImage()
+	}
+	logrus.Infof("Fetching vulnerabilities for %s", imageName)
```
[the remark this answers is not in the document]
RTann: https://github.com/stackrox/stackrox/pull/7575

api/grpc/logging_interceptor.go
RTann: doesn't seem like we can rely on this:

```
{"Event":"Received gRPC request from unknown","HEADERS":{":authority":["localhost:8443"],"content-type":["application/grpc"],"user-agent":["grpc-go/1.57.0"]},"Level":"info","Location":"logging_interceptor.go:28","Method":"/scannerV1.ImageScanService/GetImageComponents","Time":"2023-08-23 22:25:36.923706"}
```

Going to revert back to the original but leave out the peer's address because it added no value

PR #1141, opened 2023-03-30
remove grpc-middleware direct dependency

RTann:
Dependabot made a PR #1138 to update that `grpc-middlware` but CI was failing because the new version deprecated `grpcmiddleware .WithUnaryServerChain` in favor of `grpc.ChainUnaryInterceptor`. This PR makes this replacement.

See https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware#WithUnaryServerChain for more information.

### DIFF
--- a/api/grpc/grpc.go
+++ b/api/grpc/grpc.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 
 	"github.com/NYTimes/gziphandler"
-	grpcmiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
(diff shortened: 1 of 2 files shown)


RTann:
> LGTM.
> 
> I checked that dependabot will notify us on later releases and only skip this one.
> 
> I also slightly added some context to the description to make clear how this was found.

Just a heads up I updated the description again because it said this PR updates the dependency's version. This PR actually removes it as a direct dependency, but it is re-added as an indirect dependency with the same version

PR #1143, opened 2023-04-03
[DO NOT MERGE] ROX-16170: gracefully shutdown node inventory when not on openshift or rhcos

jschnath:
This PR is to gracefully shut down the `node-inventory` container when we discover that it runs on something different than Openshift or RHCOS. This is to save resources and not run a container that does nothing.

## Tested
- On colima (non-openshift) - the `node-inventory` quits gracefully, scanner deployment stays (even after force-rescanning image).
- On openshift - both `node-inventory` and scanner deployments continue working. 

See below for screenshots from the tests.

### DIFF
--- a/api/v1/nodeinventory/service.go
+++ b/api/v1/nodeinventory/service.go
@@ -51,7 +51,7 @@ func (s *serviceImpl) GetNodeInventory(_ context.Context, _ *v1.GetNodeInventory
 		log.Errorf("error analyzing node %q: %v", s.nodeName, err)
 		switch {
 		case errors.Is(err, detection.ErrNodeScanningUnavailable):
-			log.Infof("Replying and gracefully shutting down the container because it is not RHCOS")
+			log.Infof("Gracefully shutting down the container because it is not RHCOS")
 			defer s.shutdownContainer()
 			return nil, err
 		default:
--- a/api/v1/nodeinventory/service.go
+++ b/api/v1/nodeinventory/service.go
@@ -23,7 +23,7 @@ type Service interface {
 }
 
 // NewService returns the service for node scanning
-func NewService(nodeName string) Service {
+func NewService(nodeName string, shutdownContainer context.CancelFunc) Service {
 	cachedCollector := nodeinventory.NewCachingScanner(
 		&nodeinventory.Scanner{},
 		"/cache/inventory-cache",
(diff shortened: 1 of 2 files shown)


jschnath:
Manual test locally, still need to test on openshift
image

vikin91:
Tested on scanner deployment on openshift (to check whether image scanning is healthy):

Screenshot 2023-04-04 at 18 46 24

Here are the logs from the `node-inventory` container - as expected, it stays alive on OpenShift

Screenshot 2023-04-04 at 18 47 37

RTann:
The idea is to run this as a container in a pod in a DaemonSet, right? Would the DaemonSet not just try to restart it upon shutdown?

vikin91:
> The idea is to run this as a container in a pod in a DaemonSet, right? Would the DaemonSet not just try to restart it upon shutdown?

We exit with status 0, so the container shall not be recreated. We will test that in a moment...

jvdm:
How do we guarantee the container won't be restarted? I am unfamiliar with the mechanism we are using here to ensure the container will be permanently shut down.  Can you clarify?

RTann:
Also, if not done already, we can try to make sure this is only deployed in an OpenShift environment (need to update the Helm charts I'm sure), so only OpenShift users have this potentially deployed

jvdm:
> We exit with status 0, so the container shall not be recreated. We will test that in a moment...

Oh, just noticed that. Disregard my comment.

vikin91:
It is not trivial - currently it tries to restart it. We are trying now this advice ;) 

Screenshot 2023-04-04 at 19 07 52

vikin91:
> Also, if not done already, we can try to make sure this is only deployed in an OpenShift environment (need to update the Helm charts I'm sure), so only OpenShift users have this potentially deployed

We have it in place already, this container will be created only on openshift 4.
In this PR, we want to kill it if it runs on openshift 4 but on something different than RHCOS.

vikin91:
Sorry, I think we will not merge this (it is not a blocker for the release).

Applying the trick with `terminationMessagePolicy: File` works in a way that it does stop the container `status: Completed` but then it restarts it after a while. I guess we have too little time to find a working solution.

Let's release 4.0 scanner without this PR, unless you know a way to make it work the way we intend?

vikin91 (changes requested):
Reviewing :x: to block the merge.

Unfortunately, the idea did not work as we expected and there is too little time before the release to dig deeper. Let's release without this PR.

PR #1145, opened 2023-04-03
ROX-13935: Enable RHCOS FF

vikin91:
This PR enables the feature flag for RHCOS node scanning.


jvdm (approved):
LGTM.

I also checked that as long as we don't send any `Components` to `GetNodeVulnerabilities()` [we should go through the previous scanning path](https://github.com/stackrox/scanner/blob/757cca5fc5995d041793b50a83ede1e85414dd0b/api/v1/nodescan/service.go#L245-L244).

PR #1154, opened 2023-04-10
ROX-16421: support OCI image index media type

RTann:
Scanner does not currently support images with manifest media type `application/vnd.oci.image.index.v1+json`. This PR adds this support by bumping the docker-registry-client library version which adds this support as well as adding a handler for this manifest type.

See https://github.com/stackrox/scanner/pull/1155 for proof this was not supported prior to this PR.

### DIFF
--- a/pkg/scan/scan.go
+++ b/pkg/scan/scan.go
@@ -5,13 +5,13 @@ import (
 	"fmt"
 	"io"
 
-	v5Manifest "github.com/containers/image/v5/manifest"
+	imageManifest "github.com/containers/image/v5/manifest"
 	"github.com/docker/distribution"
-	"github.com/docker/distribution/manifest/schema1"
-	"github.com/docker/distribution/manifest/schema2"
+	"github.com/docker/distribution/manifest/manifestlist"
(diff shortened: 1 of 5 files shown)


pkg/scan/scan.go
```diff
@@ -179,7 +188,7 @@ func handleManifest(reg types.Registry, manifestType, remote, ref string) (diges
 		}
 		layers := parseLayers(manifest.Layers)
 		return dig, layers, nil
-	case ociSpec.MediaTypeImageManifest:
+	case registry.MediaTypeImageManifest:
```
dcaravel, line 191: Thoughts on using the types from [opencontainers/image-spec](https://github.com/opencontainers/image-spec/blob/48fd0740a26631ae0f59f514cbf017bcf841d00b/specs-go/v1/mediatype.go#L24-L28) for `MediaTypeImageManifest` and `MediaTypeImageIndex`?

dcaravel (changes requested):
/LGTM

dcaravel (approved):

PR #1164, opened 2023-04-19
ROX-13770: Introduce local Node Scanner

Maddosaurus:
To make local experimentation and debugging with Node Scanning code easier, this PR introduces a CLI binary that calls said code.
For the scan to work, the binary requires the path to a folder containing a filesystem.
As the Node Scanning code requires the `rpmdb` binary to be present in the PATH, additionally to the CLI binary, also a self-contained Docker image is introduced to make running the binary possible on common dev platforms.

After discussions in a Maple Demo Session, I have created a follow-up ticket with more feature ideas for this local node scanner:
[ROX-16621](https://issues.redhat.com/browse/ROX-16621).


tools/local-nodescanner/main.go
```diff
@@ -0,0 +1,70 @@
...
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/stackrox/scanner/pkg/analyzer/nodes"
+)
+
+func main() {
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fPath := flag.String("fspath", "/host", "Path to the root folder of a filesystem")
+	fRHCOSrequired := flag.Bool("rhcosRequired", true, "Fails scan if filesystem is not RHCOS. Node Scanning only works on RHCOS.")
```
vikin91, line 16: `filesystem is not RHCOS` it is actually the OS, not the filesystem. Maybe this could be rephrased?

vikin91 (commented):
Looks good, but I think we should mention the content sets in the readme as it is an important element of this all

tools/local-nodescanner/README.md
```diff
@@ -0,0 +1,52 @@
...
+Be warned that RPM installed via `brew` on OSX *will not work correctly*.
+
+The only required flag is the path to a filesystem. 
+This can be a RO-mount of a running system (e.g. `/host` in the Compliance or Node-Scanner images),
+or an unpacked filesystem, e.g. from a Docker image or ISO.
+
+As of writing this readme, only RHCOS RPM components are supported by the Node Scan.
+Therefore, not a full filesystem is needed. The minimal folder/file structure needed for a scan to succeed is:
+- `/etc/redhat-release` with contents denoting an RHCOS system (e.g. `Red Hat Enterprise Linux CoreOS release 4.12`)
+- `/usr/share/rpm` containing the RPM database
```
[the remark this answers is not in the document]
Maddosaurus: For the `os-release`, I grabbed an RHCOS 4.9 version, which has hardcoded CPEs. Hence I didn't catch that. Thanks!
I have now corrected the os-release to 4.12 and added the content_sets.json to required files.

Maddosaurus:
After discussions in a Maple Demo Session, I have created a follow-up ticket with more feature ideas for this local node scanner:
[ROX-16621](https://issues.redhat.com/browse/ROX-16621).

tools/local-nodescanner/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Building
+A `makefile` target named `local-nodescanner` is available in the main makefile.
+It will create binaries in the projects' `bin` folder.
+For ease of use, a Docker image is also available as target `local-nodescanner-image`.
+
+## Running the Docker image
+The Docker image only requires the path to a target filesystem to be scanned.
+As the default for `fspath` is set to `/host`, one can run the image without changes when mounting the target fs to the right path:
+`docker run -it -v /path/to/rhcos/fs:/host local-nodescanner:2.29.x-5-g7a3b50ef72-dirty`
```
vikin91, line 13: Would that work here?

```suggestion
`docker run -it -v /path/to/rhcos/fs:/host local-nodescanner:$(make tag)`
```

tools/local-nodescanner/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+## Requirements
+The scanning code requires an `rpmdb` binary to be available in the executing systems `PATH`.
+Be warned that RPM installed via `brew` on OSX *will not work correctly*, as it will produce an empty RPM database.
+
+The only required flag is the path to a filesystem. 
+This can be a RO-mount of a running system (e.g. `/host` in the Compliance or Node-Scanner images),
+or an unpacked filesystem, e.g. from a Docker image or ISO.
+Refer to `testdata/NodeScanning/rhcos4.12-minimal.tar.gz` for an archive containing a minimal example.
```
vikin91, line 24: I was hoping that we get a bit more instructions how to handle that 17MB file - could we maybe get a command where we extract it to some tmp location and then do `docker run` with mounting it?

tools/local-nodescanner/main.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Infof("Number of installed RPM packages: %v", len(components.CertifiedRHELComponents.Packages))
+		for _, c := range components.CertifiedRHELComponents.Packages {
+			log.Debugf("Component: %v", c)
+		}
+	}
+
+	if components.CertifiedRHELComponents.ContentSets != nil {
+		log.Infof("Number of discovered ContentSets: %v", len(components.CertifiedRHELComponents.ContentSets))
+		for _, cs := range components.CertifiedRHELComponents.ContentSets {
+			log.Debugf("CPE: %v", cs)
```
vikin91, line 66:
```suggestion
			log.Debugf("Content set: %v", cs)
```

vikin91 (approved):
Il looks good, but I found some minor issues. Giving it green, but please check them

tools/local-nodescanner/README.md
Maddosaurus: That works, great idea!

tools/local-nodescanner/README.md
```diff
@@ -0,0 +1,29 @@
...
+For ease of use, a Docker image is also available as target `local-nodescanner-image`.
+
+## Running the Docker image
+The Docker image only requires the path to a target filesystem to be scanned.
+As the default for `fspath` is set to `/host`, one can run the image without changes when mounting the target fs to the right path:
+`docker run -it -v /path/to/rhcos/fs:/host local-nodescanner:2.29.x-5-g7a3b50ef72-dirty`
+
+
+
+## Requirements
```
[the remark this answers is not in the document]
Maddosaurus: Good catch! Done

tools/local-nodescanner/README.md
Maddosaurus: I have added a complete working example that should get anyone up and running

Makefile
```diff
@@ -460,3 +465,19 @@ else
...
 endif
+
+
+###########
+## Tools ##
+###########
+
+# Local Node Scanner
+.PHONY: local-nodescanner
+local-nodescanner:
```
RTann, line 476: is the idea that this should only be used inside a container? Otherwise, Mac users won't be able to use the binary directly :(

tools/local-nodescanner/main.go
```diff
@@ -0,0 +1,70 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/stackrox/scanner/pkg/analyzer/nodes"
+)
+
+func main() {
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fPath := flag.String("fspath", "/host", "Path to the root folder of a filesystem")
+	fRHCOSrequired := flag.Bool("rhcosRequired", true, "Fails scan if path does not contain a filesystem generated by RHCOS. Node Scanning only works on RHCOS.")
+	fUncertifiedRHEL := flag.Bool("uncertifiedRhel", false, "Set true for CentOS and false for RHEL.")
```
RTann, line 17: nit: `uncertifiedRHEL` (not sure if that breaks flag name conventions)

testdata/NodeScanning/rhcos4.12-minimal.tar.gz
RTann, line 1: feeling a bit uneasy about pushing a 17MB file to the repo. How was this generated? Can we just recreate it on-demand? Maybe add some make targets to pull and save the image or something?

tools/local-nodescanner/main.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Fatalf("Encountered error while formatting path: %v", err)
+	}
+
+	log.Infof("Analyzing rootfs in %v", fspath)
+	components, err := nodes.Analyze(context.Background(), "nodename", fspath, nodes.AnalyzeOpts{UncertifiedRHEL: *fUncertifiedRHEL, IsRHCOSRequired: *fRHCOSrequired})
+	if err != nil {
+		log.Errorf("Encountered error while scanning: %v", err)
+	}
+
+	printResultsToTTY(components)
```
jvdm, line 34: Nit: You are simply writing to stdout, which might not be a TTY and can be some other file type. I would replace `TTY` with `Stdout` or just `printResults()`.

tools/local-nodescanner/main.go
jvdm: That would make it harder to use in the command line, assuming it's one of the intended use-cases of the tool? I would use `rhcos-required` or `rchosrequired`, honestly.

Makefile
```diff
@@ -460,3 +465,19 @@ else
...
+	@echo "+ $@"
+	GOOS=linux GOARCH=amd64 go build -trimpath -o ./tools/bin/local-nodescanner ./tools/local-nodescanner
```
jvdm, line 478: Should we build this inside our `$(BUILD_IMAGE)`?

Makefile
Maddosaurus: The idea is to have maximum flexibility. I have added a UBI 9 container target in this PR to make it runnable anywhere, but we could add more, e.g. UBI 8 or even Fedora latest for a bleeding edge RPM version.
The binary can either be used for tests and/or CI, or as a debug entry point for local debugging. That local debugging would not work on bare metal OSX w.r.t. RPM packages, but still might be useful, or could be done in the container.

testdata/NodeScanning/rhcos4.12-minimal.tar.gz
Maddosaurus: Most of the 17MB is actually the RPM package DB of an RHCOS 4.12 installation.
AFAIK there is no Docker image for RHCOS available, so I generated this by spinning a cluster and then manually pulling the RPM DB plus the additional text/json files that Node Scanning relies on.

I don't know whether we can automate that process or where to get a pre-populated RPM DB. 
I remember that we found the download link of an RHCOS live image, though we would need to check whether that comes with a populated DB at rest or if it needs to be started.

Makefile
Maddosaurus: I wanted to keep the tool rather isolated, but if you think we should use the build image, I'm happy to move that over into it.

jvdm: By isolation, do you mean minimizing dependencies? If yes, I think it's a very good goal.

If the primary goal of the tool is to aid development and local testing, I think it is OK not to use the build image to build it. Plan to have this evolve into something we can leverage in CI. Using the builder image might be necessary to ensure the built environment doesn't affect the final binary.

tools/local-nodescanner/main.go
Maddosaurus: I have renamed both flags, see https://github.com/stackrox/scanner/pull/1164/commits/07e33824d9936c620f60d2a017297ec5a06fd3cc

Makefile
Maddosaurus: Thanks for the feedback! I've introduced a new dockerized build that is based on the builder image. Let me know what you think! 
See https://github.com/stackrox/scanner/pull/1164/commits/07e33824d9936c620f60d2a017297ec5a06fd3cc

testdata/NodeScanning/rhcos4.12-minimal.tar.gz
Maddosaurus: @RTann I have checked and there is one way I can think of:

1. Download the raw live rootfs from OpenShift (e.g. [here](https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.11/latest/rhcos-live-rootfs.x86_64.img))
2. Unpack the img with `cpio`
3. Unpack the contained squashfs with `unsquashfs`
4. Navigate the resulting raw ostree to find the root volume
5. Copy all required files into our `/testdata` folder

If I can figure out a reliable way of navigating the ostree layers (all hashes), I could build this as a makefile target and parametrize the script to download a specific RHCOS version. 
I know that 17MB is quite a bit, but I don't know whether we can build automation robust enough to not break with the next RHCOS update. 
What do you think?

jvdm (approved):
Curious to know what will be the outcome of https://github.com/stackrox/scanner/pull/1164#discussion_r1190109211, but otherwise, LGTM.

Maddosaurus:
@jvdm / @RTann Regarding a makefile target: I have hit my timebox trying to make this work. It works on Linux, but fails on OSX when trying to use `unsquashfs`. You can see the state here: https://github.com/stackrox/scanner/commit/db691c680b71725d4118a292e9d7c2052ddd90cb.
I am merging this as is, but have noted this in a follow-up ticket, ROX-16621.

PR #1212, opened 2023-07-10
return err when version conversion fails

daynewlee:
Return error instead of nil when version conversion fails

### DIFF
--- a/api/v1/orchestratorscan/service.go
+++ b/api/v1/orchestratorscan/service.go
@@ -117,7 +117,7 @@ func (s *serviceImpl) GetIstioVulnerabilities(_ context.Context, req *v1.GetIsti
 	}
 
 	if req.GetIstioVersion() == "" {
-		return nil, errors.New("Can't get vulnerabilities for empty v.")
+		return nil, errors.New("Can't get vulnerabilities for empty version.")
 	}
 	v, err := convert.TruncateVersion(req.GetIstioVersion())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


api/v1/orchestratorscan/service.go
```diff
@@ -122,7 +122,7 @@ func (s *serviceImpl) GetIstioVulnerabilities(_ context.Context, req *v1.GetIsti
 	version, err := convert.TruncateVersion(req.GetIstioVersion())
 	if err != nil {
 		log.Warnf("Unable to convert Istio version of %s - %v. Skipping...", version, err)
-		return nil, nil
+		return nil, errors.Wrap(err, "Failed to convert Istio version")
```
RTann, line 125: nit: `just return `err` here

api/v1/orchestratorscan/service.go
```diff
@@ -122,7 +122,7 @@ func (s *serviceImpl) GetIstioVulnerabilities(_ context.Context, req *v1.GetIsti
 	version, err := convert.TruncateVersion(req.GetIstioVersion())
```
RTann, line 122: nit: `v` instead of `version`, as `version` is also the name of a package this file requires.

api/v1/orchestratorscan/service.go
```diff
@@ -122,7 +122,7 @@ func (s *serviceImpl) GetIstioVulnerabilities(_ context.Context, req *v1.GetIsti
...
 	if err != nil {
 		log.Warnf("Unable to convert Istio version of %s - %v. Skipping...", version, err)
```
RTann, line 124: `log.Warnf("Unable to convert Istio version: %v. Skipping...", err)`

connorgorman:
This is a good fix, but maybe in a follow up lets just wrap this in a panic handler

api/v1/orchestratorscan/service.go
```diff
@@ -117,16 +117,16 @@ func (s *serviceImpl) GetIstioVulnerabilities(_ context.Context, req *v1.GetIsti
 	}
 
 	if req.GetIstioVersion() == "" {
-		return nil, errors.New("Can't get vulnerabilities for empty version.")
+		return nil, errors.New("Can't get vulnerabilities for empty v.")
```
RTann, line 120: v -> version

RTann (approved):
The one nit then merge

PR #1223, opened 2023-07-22
ROX-18202: Fix language layer error due to unique constraint

connorgorman:
The issue was laid out in the doc by David. Basically what's happening is that the sample images have layers that are identical so they have the same "layer_name" which is the digest, but the lineage is different. This means that if you have identical layers with vulns that have different lineages (or parent layers) then the second insertion is due to the conflict on the layer_name (insert does on conflict do nothing). This change makes it so that table is keyed on layer_name and lineage, effectively only deduping layers with the exact same parents (which is good because we don't need to reinsert the layers). The unit test should test this case

~Note: for some reason, this does not work on scratch, but I find that less pressing because it seems to be when the language vuln is simply in the base layer~

~This is fixed by always evaluating language vulns even if there are no db features~

### DIFF
--- a/database/pgsql/migrations/00020_languagelayer_lineage_pk.go
+++ b/database/pgsql/migrations/00020_languagelayer_lineage_pk.go
@@ -1,0 +1,14 @@
+package migrations
+
+import "github.com/remind101/migrate"
+
+func init() {
+	RegisterMigration(migrate.Migration{
+		ID: 20,
+		Up: migrate.Queries([]string{
+			`ALTER TABLE LanguageLayer DROP CONSTRAINT IF EXISTS languagelayer_pkey`,
+			`ALTER TABLE LanguageLayer DROP CONSTRAINT IF EXISTS languagelayer_layer_name_key`,
+			`ALTER TABLE LanguageLayer ADD PRIMARY KEY (layer_name, lineage)`,
+		}),
+	})
+}
--- a/e2etests/vuln_test.go
+++ b/e2etests/vuln_test.go
@@ -334,6 +334,24 @@ func TestDistrolessVulnImages(t *testing.T) {
 				{"base-files", "9.9+deb9u13", []expectedVuln{}},
 			},
 		},
+		{
+			image: "quay.io/rhacs-eng/sandbox:drools-debian",
+			expectedFeatures: []feature{
+				{"drools", "6.4.0.final", []expectedVuln{
+					{name: "CVE-2021-41411", fixedBy: ""},
+				},
+				},
+			},
+		},
+		{
+			image: "quay.io/rhacs-eng/qa:drools-ubi-minimal",
+			expectedFeatures: []feature{
+				{"drools", "6.4.0.final", []expectedVuln{
+					{name: "CVE-2021-41411", fixedBy: ""},
+				},
+				},
+			},
+		},
 	} {
 		t.Run(testCase.image, func(t *testing.T) {
 			testSingleVulnImage(testCase, t)
--- a/e2etests/vuln_test.go
+++ b/e2etests/vuln_test.go
@@ -242,6 +242,24 @@ func TestStackroxVulnImages(t *testing.T) {
 				},
 			},
 		},
+		{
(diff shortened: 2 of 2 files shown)


connorgorman:
I'll look into the failure

database/pgsql/migrations/00020_languagelayer_lineage_pk.go
```diff
@@ -0,0 +1,14 @@
+package migrations
+
+import "github.com/remind101/migrate"
+
+func init() {
+	RegisterMigration(migrate.Migration{
+		ID: 20,
+		Up: migrate.Queries([]string{
+			`ALTER TABLE LanguageLayer DROP CONSTRAINT IF EXISTS languagelayer_pkey`,
```
RTann, line 9: is this a generic, default constraint? I don't see it in my deployment:

```
postgres=# \d languagelayer
                         Table "public.languagelayer"
       Column       |          Type          | Collation | Nullable | Default 
--------------------+------------------------+-----------+----------+---------
 layer_id           | integer                |           | not null | 
 layer_name         | character varying(128) |           | not null | 
 component_data     | bytea                  |           | not null | 
 removed_components | text[]                 |           |          | 
 lineage            | character varying      |           |          | 
Indexes:
    "languagelayer_layer_name_key" UNIQUE CONSTRAINT, btree (layer_name)
```

database/pgsql/migrations/00020_languagelayer_lineage_pk.go
```diff
@@ -0,0 +1,14 @@
...
+			`ALTER TABLE LanguageLayer DROP CONSTRAINT IF EXISTS languagelayer_layer_name_key`,
```
RTann, line 10: so dropping this constraint removes the `UNIQUE` part of `layer_name VARCHAR(128) NOT NULL UNIQUE`?

api/v1/models.go
```diff
@@ -131,32 +131,33 @@ func LayerFromDatabaseModel(db database.Datastore, dbLayer database.Layer, linea
 
 	notes := getNotes(layer.NamespaceName, uncertifiedRHEL)
 
-	if (withFeatures || withVulnerabilities) && (dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName)) {
-		for _, dbFeatureVersion := range dbLayer.Features {
-			feature := featureFromDatabaseModel(dbFeatureVersion, opts.GetUncertifiedRHEL(), depMap)
+	if withFeatures || withVulnerabilities {
+		if dbLayer.Features != nil || namespaces.IsRHELNamespace(layer.NamespaceName) {
```
RTann, line 135: would need to see if our scanning notes accounts for us still finding language vulns without OS-level vulns. Don't think we considered this possibility before

dcaravel: In the [GetComponents](https://github.com/stackrox/scanner/blob/7d161d3089a7c3477cfd137ce6cd8aa180bf5a03/api/v1/models.go#L225-L227) flow used via GRPC looks like the OS-level vulns vs. language vulns are separated - FWIW

database/pgsql/migrations/00020_languagelayer_lineage_pk.go
dcaravel: `languagelayer_pkey` is the name of the constraint after `ALTER TABLE LanguageLayer ADD PRIMARY KEY (layer_name, lineage)` is applied, the default name of a primary key looks to be `<table name>_pkey`

Possibly added to the migration so that it would not cause errors if re-ran.

dcaravel (dismissed):
LGTM

Cloned and retested per the original findings and this indeed fixes the issue.  Thanks @connorgorman !

dcaravel (commented):
Retesting, GRPC looks good, HTTP is better except for if `scratch` is the base image detection fails.  The changes to `api/v1/models.go` seemed to fix that, were those reverted due to other failing tests?

connorgorman:
Yeah i separated them because it was causing other issues and is a larger change

dcaravel (approved):

PR #1224, opened 2023-07-24
ROX-17114: Fix misattribution of Java versions when there are multiple matching lines

connorgorman:
So we don't stop early if we have matched all the possible key lines in the manifest.mf file. This file can contain many entries and this will cause us to attribute the data to a wrong version

### DIFF
--- a/pkg/analyzer/java/parse_manifest_test.go
+++ b/pkg/analyzer/java/parse_manifest_test.go
@@ -7,8 +7,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const testTomcat10ManifestMF = `
-Manifest-Version: 1.0
+const testTomcat10ManifestMF = `Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: tomcat-embed-core
 Bundle-SymbolicName: org.apache.tomcat-embed-core
@@ -39,8 +38,7 @@ Specification-Vendor: Eclipse Foundation
 Specification-Version: 3.0
 `
 
-const testTomcat9ManifestMF = `
-Manifest-Version: 1.0
+const testTomcat9ManifestMF = `Manifest-Version: 1.0
 Automatic-Module-Name: org.apache.tomcat.embed.core
 Bnd-LastModified: 1593547891735
 Bundle-ManifestVersion: 2
--- a/pkg/analyzer/java/parse_manifest_test.go
+++ b/pkg/analyzer/java/parse_manifest_test.go
@@ -2,11 +2,11 @@ package java
 
 import (
 	"bytes"
-	"fmt"
+	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
-const testManifestMF = `
+const testTomcat10ManifestMF = `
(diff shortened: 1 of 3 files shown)


pkg/analyzer/java/parse_manifest_test.go
```diff
@@ -0,0 +1,49 @@
+package java
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+)
+
+const testManifestMF = `
```
daynewlee, line 9: @RTann 
Maybe we should add this specific example to the Claircore to manifest parsing. What do you think?

pkg/analyzer/java/parse_manifest_test.go
```diff
@@ -0,0 +1,49 @@
...
+Implementation-Title: jakarta.security.auth.message
+Implementation-Vendor: Apache Software Foundation
+Implementation-Version: 3.0
+Specification-Title: Jakarta Authentication SPI for Containers
+Specification-Vendor: Eclipse Foundation
+Specification-Version: 3.0
+`
+
+func TestStopOnceFirstKeyIsFound(t *testing.T) {
+	manifestBytes := bytes.NewBuffer([]byte(testManifestMF))
```
connorgorman, line 42: Whoops, I'll added assertions here

daynewlee (approved):
LGTM

pkg/analyzer/java/parse_manifest_test.go
```diff
@@ -0,0 +1,173 @@
...
+Implementation-Title: javax.servlet
+Implementation-Vendor: Apache Software Foundation
+Implementation-Version: 4.0.FR
+Specification-Title: Java API for Servlets
+Specification-Vendor: Sun Microsystems, Inc.
+Specification-Version: 4.0
+`
+
+func TestStopOnceFirstKeyIsFound(t *testing.T) {
+	manifest, err := parseManifestMFFromReader("", bytes.NewBufferString(testTomcat9ManifestMF))
```
RTann, line 153: can you check `Bundle-Version` too for completeness?

pkg/analyzer/java/parse_manifests.go
```diff
@@ -83,7 +76,7 @@ func parseManifestMF(locationSoFar string, f *zip.File) (parsedManifestMF, error
 			isTemplatedManifestFile = true
 			continue
 		}
-
+		valueFromLine = strings.TrimSpace(valueFromLine)
```
RTann, line 79: should we do this before the `if valueFromLine == ""` check?

connorgorman: Ah yeah, good call

RTann:
I will be taking over this PR at this point

daynewlee (approved):

pkg/analyzer/java/parse_manifest_test.go
RTann: ClairCore doesn't have this problem, as they only read the "main" section and also stop at `Name`: https://github.com/quay/claircore/blob/main/java/jar/jar.go#L526

PR #1236, opened 2023-08-15
fix: do not skip NVD vulns with just CVSSv3

RTann:
Before updating the condition:

```
INFO[0130] Obtained metadata for 180078 NVD vulns (skipped 42616)
```

After:

```
INFO[0132] Obtained metadata for 209238 NVD vulns (skipped 13309)
```

We were ignoring a lot of vulnerabilities solely because they lacked CVSSv2 scores


ext/vulnsrc/manual/manual.go
```diff
@@ -13,168 +12,7 @@ type updater struct {
 }
 
 // Vulnerabilities lists vulnerabilities which may not already exist in the feeds for other distros.
-var Vulnerabilities = []database.Vulnerability{
```
RTann, line 16: These entries are no longer required, as we can fetch their entries from NVD properly now

daynewlee (approved):
LGTM

PR #1237, opened 2023-08-16
Take runtime.GOARCH into account for list manifests

connorgorman:


pkg/scan/scan.go
```diff
@@ -153,6 +154,34 @@ func renderDigest(manifest payloadGetter) (digest.Digest, error) {
...
+		}
+		if m.Platform.Architecture == "amd64" {
+			foundAMD = true
+			amdManifest = m
+		}
+	}
+	if foundAMD {
+		return handleManifest(reg, amdManifest.MediaType, remote, amdManifest.Digest.String())
+	}
+	return "", nil, errors.Errorf("no manifest in list matched linux and amd64 or %s architectures: %q", runtime.GOARCH, ref)
```
RTann, line 182: nit: can you add `remote`, too?

pkg/scan/scan.go
```diff
@@ -153,6 +154,34 @@ func renderDigest(manifest payloadGetter) (digest.Digest, error) {
 	return dig, nil
 }
 
+func handleManifestLists(reg types.Registry, remote, ref string, manifests []manifestlist.ManifestDescriptor) (digest.Digest, []string, error) {
+	if len(manifests) == 0 {
+		return "", nil, errors.Errorf("no valid manifests found for %s:%s", remote, ref)
+	}
+	if len(manifests) == 1 {
```
RTann, line 161: is this case needed? should be covered by the loop, right? plus, I think this allows non-linux platforms, too

pkg/scan/scan.go
connorgorman: will do

pkg/scan/scan.go
connorgorman: So this actually deviates from what central does today:
```
	if len(manifestList.Manifests) == 1 {
		return handleManifests(r, manifestList.Manifests[0].MediaType, remote, manifestList.Manifests[0].Digest.String())
	}
	```

RTann (approved):

PR #1258, opened 2023-09-20
ROX-18363: Fix RHEL CVSS equals 0 even after update

connorgorman:
Here's what happening:

1. A CVE comes in via update that has CVSS = 0 because it is currently being investigated by RH prod sec
2. The CVE is updated with a valid CVSS score, but the insertion logic writes a new CVE entry into the database so you have something like this where the first 1 has cvss that's empty, but the second is valid:

```
postgres=# select name, updated,cvss3, package_name, cpe from vuln_v2 where name = 'CVE-2022-40090';
      name      |        updated         |                      cvss3                       |  package_name   |                   cpe                    
----------------+------------------------+--------------------------------------------------+-------+-----------------+------------------------------------------
<CVE-2023-WXYZ> | 2023-09-01 00:00:00+00 |                                                  | <package>       | cpe:/o:redhat:enterprise_linux:8::baseos
<CVE-2023-WXYZ> | 2023-09-03 00:00:00+00 | 6.5/CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H | <package>       | cpe:/o:redhat:enterprise_linux:8::baseos
<CVE-2023-WXYZ> | 2023-09-05 00:00:00+00 | 6.5/CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H | <package>       | cpe:/o:redhat:enterprise_linux:8::baseos
 ```

3. When we're evaluating for vulns we run a query to get all vulns that affect the image. This list has some matching logic, but it's largely in insertion order meaning that the CVSS = 0 vuln is first. Then we dedupe the second vuln in the logic that is modified.

Modifying the insertion logic would be non-trivial as we'd effectively need to match all components of the CVE anyways and the hash value is not useful to us in this circumstance as other values may have changed. Note that this will NOT fix the circumstance where say a CPE, package_name or package_module is corrected because the search query is `vuln.package_name = $1 AND vuln.package_module = $2 AND vuln.cpe = $3;`. Thus, instead of modifying the insertion, if we sort by updated time, we'll always reconcile the latest and most up to date first

### CODE REVIEW

**Review by jvdm — APPROVED** [review]
LGTM. I was wondering if we could sort within the query, but it seems like the performance gain is negligible.

File: database/pgsql/rhelv2_vulnerability.go [hunk 1]
```diff
@@ -216,6 +217,11 @@ func (pgSQL *pgSQL) GetRHELv2Vulnerabilities(records []*database.RHELv2Record) (
 
 			addedVulns := uniqueVulns[recordID]
 
+			// Order the vulns in terms of updated first so that if we have multiple of the same vulns
+			// the ones with the latest update time will be used
+			sort.SliceStable(matchedVulns, func(i, j int) bool {
+				return matchedVulns[i].Updated.After(matchedVulns[j].Updated)
```
RTann, line 223: is this safe when updated time may possibly be blank? (if it's possible?)

RTann:
Makes sense. ClairCore doesn't have this issue (or if it does, it's not forever), as it garbage collects and (maybe?) only obtains the results of the latest modification operations

database/pgsql/rhelv2_vulnerability.go
connorgorman: Yes, because these are not pointers and are just time.Time which is zero value safe

RTann (approved):

jvdm:
@connorgorman This should get your E2E clean: 95dc0fb2

connorgorman:
I was able to verify this by taking a 4.2.0 which has cvss = 0 for CVE-2023-39975 and then an update adds a new entry. The scan even after the update is 0, but then when I switch from 4.2.0 scanner to this PRs scanner, it will return the correct value from the updated entry

PR #1263, opened 2023-10-03
ROX-19942: support usr/lib/redhat-release and RHEL Atomic Host release

RTann:
It is possible an official Red Hat container image has `etc/redhat-release` as a symlink to `usr/lib/redhat-release` instead of being a regular file. This PR adds `usr/lib/redhat-release` as a file to track.

It also updates the `etc/os-release` support for RHEL/CentOS in case there may be some other case we miss. The update ensures only the major version of the release is returned when reading a RHEL or CentOS namespace from `etc/os-release`.

### DIFF
--- a/ext/featurens/osrelease/osrelease.go
+++ b/ext/featurens/osrelease/osrelease.go
@@ -32,11 +32,12 @@ var (
 	blocklistFilenames = []string{
 		"etc/alpine-release",
 		"etc/centos-release",
+		"usr/lib/centos-release",
 		"etc/fedora-release",
 		"etc/oracle-release",
 		"etc/redhat-release",
+		"usr/lib/redhat-release",
 		"etc/rocky-release",
-		"usr/lib/centos-release",
 	}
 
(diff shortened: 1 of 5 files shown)


ext/featurens/osrelease/osrelease.go
```diff
@@ -32,11 +32,12 @@ var (
 	blocklistFilenames = []string{
 		"etc/alpine-release",
 		"etc/centos-release",
+		"usr/lib/centos-release",
 		"etc/fedora-release",
```
jvdm, line 36: I am wondering about fedora, too, since it's upstream.

jvdm (approved):

ext/featurens/osrelease/osrelease.go
RTann: We don't support Fedora in Scanner v2, so I considered removing this, but I decided against it because I didn't want to touch anything unrelated to the goal of this PR

PR #1273, opened 2023-10-11
ROX-20122: bump golang.org/x/net from 0.15.0 to 0.17.0

stehessel:
This resolves [CVE-2023-39325](aka https://github.com/advisories/GHSA-qppj-fm5r-hxr3).


RTann:
I think it'd be preferred to update the source of these deps, which is the stackrox repo: https://github.com/stackrox/scanner/pull/1274

vladbologa:
> I think it'd be preferred to update the source of these deps, which is the stackrox repo: #1274

I tried to do that (in the `release-2.30` branch, by pointing stackrox to the `release-4.1` branch tip), then ran `go mod tidy` and it didn't update the version of x/net.

vladbologa:
Merged in https://github.com/stackrox/scanner/pull/1279

PR #1287, opened 2023-10-12
ROX-20122: set gRPC max concurrent streams to 100

vladbologa:
As part of the mitigation for [cve-2023-44487](https://access.redhat.com/security/cve/cve-2023-44487), some projects have decided to reduce MaxConcurrentStreams to 100 (this is the minimum recommended value in the HTTP/2 standard - default is 250).

This PR applies this change to the Scanner gRPC server.

### DIFF
--- a/api/grpc/grpc.go
+++ b/api/grpc/grpc.go
@@ -13,17 +13,31 @@ import (
 	grpcprometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	log "github.com/sirupsen/logrus"
+	"github.com/stackrox/scanner/pkg/env"
 	"github.com/stackrox/scanner/pkg/mtls"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
 const (
-	localEndpoint = "127.0.0.1:8444"
+	localEndpoint                   = "127.0.0.1:8444"
+	defaultGrpcMaxConcurrentStreams = 100 // HTTP/2 spec recommendation for minimum value
 )
 
 func init() {
 	grpcprometheus.EnableHandlingTimeHistogram()
+}
+
+var (
+	maxGrpcConcurrentStreamsSetting = env.RegisterIntegerSetting("ROX_GRPC_MAX_CONCURRENT_STREAMS", defaultGrpcMaxConcurrentStreams)
+)
+
+func maxGrpcConcurrentStreams() uint32 {
+	if maxGrpcConcurrentStreamsSetting.Int() < 0 {
+		return defaultGrpcMaxConcurrentStreams
+	}
+
+	return uint32(maxGrpcConcurrentStreamsSetting.Int())
 }
 
 // NewAPI creates a new gRPC API instantiation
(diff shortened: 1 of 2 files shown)


vladbologa:
> Can we make the `100` configurable via an env variable? So that we have a workaround in the event this causes issues.

@dcaravel Yes we could make it configurable. BTW since I'm not familiar with Scanner, do you know what the e2e failures are about? Can it be something that this PR caused?

dcaravel:
> BTW since I'm not familiar with Scanner, do you know what the e2e failures are about? Can it be something that this PR caused?
No not caused by this PR. TL;DR; our tests rely on external data / vuln feeds, when those feeds change our tests may break (like what your seeing) - I'll fix this particular failure in a different PR.

dcaravel:
FYI, https://github.com/stackrox/scanner/pull/1288 has been merged, if bring in those changes here e2e tests should pass.

RTann:
Created https://github.com/stackrox/scanner/pull/1289 so we can accept integer env vars

RTann:
also, I'm thinking we call it ROX_MAX_CONCURRENT_STREAMS. We should coordinate with whatever may be done for the stackrox repo, though

vladbologa:
> also, I'm thinking we call it ROX_MAX_CONCURRENT_STREAMS. We should coordinate with whatever may be done for the stackrox repo, though

@RTann I called it `ROX_GRPC_MAX_CONCURRENT_STREAMS`. It's the same in `stackrox/stackrox`, but there we have an additional setting `ROX_HTTP2_MAX_CONCURRENT_STREAMS`

api/grpc/grpc.go
```diff
@@ -13,19 +13,33 @@ import (
...
 func init() {
 	grpcprometheus.EnableHandlingTimeHistogram()
 }
 
+var (
+	maxGrpcConcurrentStreamsSetting = env.RegisterIntegerSetting("ROX_GRPC_MAX_CONCURRENT_STREAMS", defaultGrpcMaxConcurrentStreams)
+)
+
+func maxGrpcConcurrentStreams() uint32 {
+	if maxGrpcConcurrentStreamsSetting.Int() < 0 {
```
RTann, line 36: Though the gRPC library defaults this to `math.MaxUint` when 0, the HTTP/2 library defaults it to at least 100 when 0. Can we also default it to 100?

api/grpc/grpc.go
```diff
@@ -13,19 +13,33 @@ import (
...
+	localEndpoint                   = "127.0.0.1:8444"
+	defaultGrpcMaxConcurrentStreams = 100 // HTTP/2 spec recommendation for minimum value
 )
 
 func init() {
 	grpcprometheus.EnableHandlingTimeHistogram()
 }
 
+var (
+	maxGrpcConcurrentStreamsSetting = env.RegisterIntegerSetting("ROX_GRPC_MAX_CONCURRENT_STREAMS", defaultGrpcMaxConcurrentStreams)
```
RTann, line 32: let's put this in `pkg/env/list.go`

api/grpc/grpc.go
vladbologa: Are you suggesting that I change `< 0` to`<= 0`?

RTann: yep

vladbologa: Yeah I missed that. I should do it also in stackrox.

api/grpc/grpc.go
vladbologa: Done.

RTann (approved):

PR #1297, opened 2023-10-13
http(metrics): disallow auto HTTP/2 enablement

RTann:
disallow HTTP/2 for the metrics server, as it's not required.


daynewlee (approved):
LGTM, and it might be better to get one more approval

PR #1310, opened 2023-10-26
ROX-19980 pre-built scanner-db image

ludydoo:


image/db/rhel/scripts/custom-entrypoint.sh
```diff
@@ -0,0 +1,66 @@
...
+  mkdir -p /var/lib/postgresql/data/pgdata
+
+  echo "Moving archive to target directory..."
+  mv /tmp/data.tar.gz /var/lib/postgresql/data/pgdata/data.tar.gz
+
+  echo "Uncompressing into /var/lib/pgsql/data/pgdata..."
+  tar -xzf /var/lib/postgresql/data/pgdata/data.tar.gz -C /var/lib/postgresql/data/pgdata
+
+  echo "Removing archive..."
+  rm /var/lib/postgresql/data/pgdata/data.tar.gz
```
SimonBaeumer, line 21: @dcaravel @RTann Could we move these steps to build time instead of start-up time?
@ludydoo proposes this to save IO in the cloud service which prevents us from cluster upgrades and increases ACS upgrade complexity.
Every restart of ACS instances causes to write ~1.8GB per instance.
Additionally the manifest must be updated that the scanner-db pod can read the Postgres Password.

What do you think about this approach? It seems simple and scaleable to us.
cc @ludydoo

PR #1313, opened 2023-10-31
chore: account for OpenShift 4.14

RTann:
OpenShift 4.14 went GA, so we need to make sure we can find all vulns for it. This codepath is not used at this time, but it's possible it would be needed were the OpenShift CPE changed. It is here for that safety.


dcaravel (approved):
LGTM

PR #1327, opened 2023-11-17
revert NVD API changes

RTann:
Builds are failing at the moment. Let's revert this change until we use the API key.


jvdm (approved):
LGTM.

PR #1318, opened 2023-11-02
chore(nvd): use NVD API v2

RTann:
NVD [plans to retire the 1.1 JSON feeds on December 15, 2023](https://nvd.nist.gov/General/News/change-timeline). This PR swaps our usage of those feeds with the v2 API.

For backwards compatibility purposes with previous Scanner versions, which may still be deployed somewhere, we convert the v2 API JSON responses to the 1.1 JSON formatting. This allows us to continue serving NVD data to live Scanners without any downtime nor required updates.

See https://github.com/stackrox/nvdtools/pull/9 for the related changes to the nvdtools library

PR #1329, opened 2023-11-17
chore(nvd): use NVD API v2

RTann:
See #1318 for the original implementation. It was reverted due to rate limits without an API key which were undetected during testing. This PR cherry-picks the original commit, slows down the frequency of the calls, and adds a simply retry mechanism.

I opted to rate limit it one call every 10 seconds to play it safe. This means the estimated time to read all the files will be ~18 minutes instead of ~11 minutes.

This PR has three commits. The first a cherry-pick of the original implementation. The second has the updates related to using the API key and using a simple backoff procedure. The third removes the API key because it's inaccessible in CI at this time and slows down the the query frequency to once every 10 seconds.


pkg/vulnloader/nvdloader/loader.go
```diff
@@ -1,139 +1,249 @@
...
+			}
 		}
-		attr, err := wfn.UnbindFmtString(cpeMatch.Cpe23Uri)
+
+		// Rudimentary rate-limiting.
+		// NVD limits users without an API key to roughly one call every 6 seconds.
+		// With an API key, it is roughly one call every 0.6 seconds.
+		// At this time, the API key is inaccessible in CI, so we opt to play it safe and make a call every 10 seconds.
+		// As of writing there are ~216,000 vulnerabilities, so this whole process should take ~18 minutes.
+		time.Sleep(10 * time.Second)
```
daynewlee, line 98: Just curious, how long does it approximately take to get all data?

RTann: about 18 minutes. Says it on the line directly above this 😄

daynewlee: It's odd that it was taking longer in scanner v4 workflow with not API key

daynewlee (approved):
LGTM

PR #1334, opened 2023-11-28
ROX-20753: Add `scanner` RHTAP build pipeline

kylape:
A few differences from the standard RHTAP build pipeline:

* Use of 6GB buildah image to avoid OOMs during container build task
* There is an extra step to fetch the vuln feed data.  This is done outside of the build step to make hermetic builds easier.

The vuln feed data script pulls the definitions from the Google storage location and writes them to the "source" folder of the "source" workspace which is shared by the build-container task.  This task will build using `image/scanner/rhtap/Dockerfile`, which expects the vuln files to be in the buildah working directory.

I decided to create a separate dockerfile for RHTAP to avoid any regressions with modifying the existing ones.  The dockerfile was created using a combination of the upstream and downstream dockerfiles.

Current RHTAP build trigger config:

* PRs will only build in RHTAP when "rhtap" is in the branch name
* Pushes to master will trigger an RHTAP build


.tekton/scanner-push.yaml
```diff
@@ -0,0 +1,422 @@
...
+              #!/usr/bin/env bash
+              mkdir -p "$(workspaces.source.path)/source"
+              blobs=(
+                nvd-definitions.zip
+                k8s-definitions.zip
+                repo2cpe.zip
+                genesis_manifests.json
+              )
+              for blob in "${blobs[@]}"; do
+                curl --fail --show-error \
```
tommartensen, line 224: You could add retry and timeout flags to the `curl` command to prevent retry failures.

image/scanner/rhtap/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+ENV CI=1
+
+COPY . /src
+WORKDIR /src
+
+RUN unzip -j blob-repo2cpe.zip -d image/scanner/dump/repo2cpe && \
+    unzip -j blob-k8s-definitions.zip -d image/scanner/dump/k8s_definitions && \
+    unzip -j blob-nvd-definitions.zip -d image/scanner/dump/nvd_definitions
+
+RUN go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(git describe --tags --abbrev=10 --dirty --long)" -o image/scanner/bin/scanner ./cmd/clair
```
tommartensen, line 18: Can you check that the version is set inside the image?
https://github.com/stackrox/scanner/blob/master/.dockerignore#L1 -> There should not be a git context, ie git commands should fail. Because you're using a subshell, this failure might be silent.

.containerignore
```diff
@@ -0,0 +1,9 @@
+# This file is derived of the .dockerignore
+# It is required to pass `.git` to the builder containers on RHTAP so we can bake build information into the binary.
+# It is required to pass the `.gitignore` files in `image/bin/` and `image/rhel/bin` and the tests in `qa-tests-backend` to avoid a dirty version.
```
tommartensen, line 3: These comments should be updated.

scripts/ci/jobs/check-generated.sh
```diff
@@ -0,0 +1,31 @@
...
+
+FAIL_FLAG="/tmp/fail"
+
+info 'Check .containerignore file is in sync with .dockerignore (If this fails, follow instructions in .containerignore to update it.)'
+function check-containerignore-is-in-sync() {
+    diff \
+        --unified \
+        --ignore-blank-lines \
+        <(grep -v -e '^#' .containerignore) \
+        <(grep -vF -e '/.git/' -e '/image/' -e '/qa-tests-backend/' .dockerignore) \
```
tommartensen, line 17: This should be updated to what you changed between .dockerignore and .containerignore

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,430 @@
...
+    - name: clone-repository
+      params:
+      - name: url
+        value: $(params.git-url)
+      - name: revision
+        value: $(params.revision)
+      - name: fetchTags
+        value: true
+      - name: depth
+        value: 1000
```
tommartensen, line 169:
```suggestion
        value: 0
```

0 will fetch the whole history.

.github/workflows/style.yaml
```diff
@@ -0,0 +1,33 @@
...
+    container:
+      image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.61
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - uses: ./.github/actions/job-preamble
```
BradLugo, line 27: Is there a way to get this from [stackrox/stackrox](https://github.com/stackrox/stackrox)? If so, would it make to do that?

kylape: I imagine we could publish the Action just like any other reusable GH Action.

.tekton/scanner-push.yaml
```diff
@@ -0,0 +1,426 @@
...
+                nvd-definitions.zip
+                k8s-definitions.zip
+                repo2cpe.zip
+                genesis_manifests.json
+              )
+              for blob in "${blobs[@]}"; do
+                curl --fail --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
+                  --output "$(workspaces.source.path)/source/blob-${blob}" \
+                  "https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob}"
+              done
```
BradLugo, line 231: Should this be the same as `gen-vuln-feed-data` from the pull request config?

kylape: yes, I need to sync the changes from the other tekton file to this one.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,449 @@
...
+      taskSpec:
+        steps:
+          - name: gen-vuln-feed-data
+            computeResources:
+              requests:
+                memory: 6Gi
+              limits:
+                memory: 6Gi
+            image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.61
+            script: |
```
RTann, line 227: Just trying to understand the workflow: how come we extracted this out of the script?

This will generate the genesis dump in each commit for a PR, right? This is different behavior from before

.tekton/scanner-push.yaml
```diff
@@ -0,0 +1,446 @@
...
+      - clone-repository
+      taskSpec:
+        steps:
+          - name: gen-vuln-feed-data
+            computeResources:
+              requests:
+                memory: 6Gi
+              limits:
+                memory: 6Gi
+            image: quay.io/stackrox-io/apollo-ci:stackrox-test-0.3.61
```
RTann, line 223: scanner-test

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,449 @@
...
+    - name: show-sbom
+      params:
+      - name: IMAGE_URL
+        value: $(tasks.build-container.results.IMAGE_URL)
+      taskRef:
+        params:
+        - name: name
+          value: show-sbom
+        - name: bundle
+          value: quay.io/redhat-appstudio-tekton-catalog/task-show-sbom:0.1@sha256:4b8a47b3cd3f80d0a9ec46ec43fd2ebe0487979f351a8eeebe5aa04854d3861c
```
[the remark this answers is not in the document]
kylape: RHTAP opens a PR to update them, like this: https://github.com/stackrox/scanner/pull/1344

.tekton/scanner-pull-request.yaml
kylape: I had it pulling the dumps from GCS before.  Would you rather it do that instead?

> how come we extracted this out of the script?

I assume you're referring to the script `.openshift-ci/build/generate-genesis-dump.sh`.  That script had calls to things like `openshift_ci_mods` and `gate_job` that didn't seem like something I wanted to execute in the RHTAP build environment.

.tekton/scanner-push.yaml
tommartensen: @kylape you could consider adding a custom Tekton Task to avoid duplication. I tried that a while ago in https://github.com/stackrox/infra/tree/tm/rhtap-onboarding/.rhtap/tasks. 

In a nutshell, you create the Task yaml and push the definition to a registry with `tkn bundle push`. 
Then you can use it like https://github.com/stackrox/infra/blob/tm/rhtap-onboarding/.tekton/infra-pull-request.yaml#L395-L409. 

IIRC there is no other step needed.

kylape:
I removed all the commits from this PR branch since I switch from downloading the vuln dump to generating it.  In other words, I just switch back to downloading the vuln dump from GCS.

image/scanner/rhtap/Dockerfile
kylape: Version should be set properly now.  I kept a `git describe` in the dockerfile to validate that it's being set properly.

.tekton/scanner-pull-request.yaml
kylape: I decided to pull the dumps from GCS for now.  We can decide if/when to generate the dumps in a later PR.

.containerignore
```diff
@@ -0,0 +1,9 @@
+# This file is derived of the .dockerignore
```
tommartensen, line 1: If this file is empty, you don't need it.

.containerignore
```diff
@@ -0,0 +1,9 @@
...
+# It is required to pass `.git` to the builder containers on RHTAP so we can bake build information into the binary.
+# It is required to pass the `.gitignore` files in `image/bin/` and `bin` to avoid a dirty version.
+# RHTAP uses Buildah which first looks up for `.containerignore`, then, if absent, `.dockerignore`.
+# Having this file enables us to have different exclusion list for RHTAP builds than in GHA/OSCI.
+# To regenerate, replace the below contents with the output of the following command (please keep these comments).
+#  grep -vF -e '/.git/' -e '/bin/' -e '/image/bin/' .dockerignore
+
+# currently empty!
```
tommartensen, line 9: Could you add that it is empty and must be kept to prevent RHTAP from using `.dockerignore` file?

tommartensen (approved):
all makes sense, please update the comment in the `.containerignore`.
Formatting of the Tekton Pipeline files is different from the ones we already have, but we have a reconcile planned already.

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# files of the dump and the manifest.
+COPY ./blob-genesis_manifests.json image/scanner/dump/genesis_manifests.json
+
+FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG}
+
+LABEL name="scanner" \
+      vendor="StackRox" \
+      maintainer="support@stackrox.com" \
+      summary="Image scanner for the StackRox Kubernetes Security Platform" \
+      description="This image supports image scanning in the StackRox Kubernetes Security Platform."
```
msugakov, line 31: Please borrow labels from other rhtap Dockerfiles for consistency and so that pipeline checks are happier. E.g. here

https://github.com/stackrox/stackrox/blob/8510071fb1dfa95026f4bc4d871e29afec6a3bda/image/roxctl/rhtap.Dockerfile#L36-L50

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+
+COPY --from=builder /src/image/scanner/scripts /
+COPY --from=builder /src/image/scanner/bin/scanner ./
+COPY --chown=65534:65534 --from=builder "/src/image/scanner/dump${NVD_DEFINITIONS_DIR}/" ".${NVD_DEFINITIONS_DIR}/"
+COPY --chown=65534:65534 --from=builder "/src/image/scanner/dump${K8S_DEFINITIONS_DIR}/" ".${K8S_DEFINITIONS_DIR}/"
+COPY --chown=65534:65534 --from=builder "/src/image/scanner/dump${REPO_TO_CPE_DIR}/" ".${REPO_TO_CPE_DIR}/"
+COPY --chown=65534:65534 --from=builder /src/image/scanner/dump/genesis_manifests.json ./
+
+RUN microdnf upgrade --nobest && \
+    microdnf install xz && \
```
msugakov, line 45: I don't know why the downstream dockerfile installs `findutils` as I can't spot any calls to that in `*.sh` code. Maybe some code existed but was removed. Maybe this was added or copied without much consideration. Maybe this was copied for manual debuggability (I, for instance, like it much more when container where I `exec` has `find` as opposed when it doesn't).

For the last point, I want to establish a baseline of utils to be present in every our image so that they are useful when poking them manually. But that's certainly outside of scope of our RHTAP onboarding.

If you don't see a reason to have `findutils`, let's proceed without them and this comment is no-op.

It's fine to not try to install `ca-certificates` because they are already present in recent ubi-minimal versions.

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    # We don't uninstall rpm because scanner uses it to get packages installed in scanned images.
+    rpm -e --nodeps $(rpm -qa curl '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
+    rm -rf /var/cache/dnf /var/cache/yum && \
+    chown -R 65534:65534 /tmp && \
+    # The contents of paths mounted as emptyDir volumes in Kubernetes are saved
+    # by the script `save-dir-contents` during the image build. The directory
+    # contents are then restored by the script `restore-all-dir-contents`
+    # during the container start.
+    chown -R 65534:65534 /etc/pki /etc/ssl && /save-dir-contents /etc/pki/ca-trust /etc/ssl && \
+    chmod +rx /scanner
```
msugakov, line 57: [nit] I can't spot `chmod +rx /scanner` in the downstream Dockerfile. Are you sure it's needed?

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    # (Optional) Remove line below to keep package management utilities
+    # We don't uninstall rpm because scanner uses it to get packages installed in scanned images.
+    rpm -e --nodeps $(rpm -qa curl '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
+    rm -rf /var/cache/dnf /var/cache/yum && \
+    chown -R 65534:65534 /tmp && \
+    # The contents of paths mounted as emptyDir volumes in Kubernetes are saved
+    # by the script `save-dir-contents` during the image build. The directory
+    # contents are then restored by the script `restore-all-dir-contents`
+    # during the container start.
+    chown -R 65534:65534 /etc/pki /etc/ssl && /save-dir-contents /etc/pki/ca-trust /etc/ssl && \
```
msugakov, line 56: [nit] This might be better in two separate lines
```suggestion
    chown -R 65534:65534 /etc/pki /etc/ssl && \
    /save-dir-contents /etc/pki/ca-trust /etc/ssl && \
```

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+COPY --chown=65534:65534 --from=builder /src/image/scanner/dump/genesis_manifests.json ./
+
+RUN microdnf upgrade --nobest && \
+    microdnf install xz && \
+    microdnf clean all && \
+    # (Optional) Remove line below to keep package management utilities
+    # We don't uninstall rpm because scanner uses it to get packages installed in scanned images.
+    rpm -e --nodeps $(rpm -qa curl '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
+    rm -rf /var/cache/dnf /var/cache/yum && \
+    chown -R 65534:65534 /tmp && \
```
msugakov, line 51: In addition to `/tmp`, the downstream dockerfile seems to change ownership of a few other things.

```bash
    chown -R 65534:65534 \
        genesis_manifests.json \
        "${NVD_DEFINITIONS_DIR}" \
        "${K8S_DEFINITIONS_DIR}" \
        "${REPO_TO_CPE_DIR}" \
        /tmp
```

I like none of that and I also found it difficult to find out why this is done for some similar things in the past.

This comment is not blocking for your PR but if you want to cargo-cult additional locations in the `chown` command, this _might_ save us from some issues during later testing.

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    unzip -j blob-nvd-definitions.zip -d image/scanner/dump/nvd_definitions
+
+RUN echo "version: $(git describe --tags --abbrev=10 --dirty --long)"; git status; go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(git describe --tags --abbrev=10 --dirty --long)" -o image/scanner/bin/scanner ./cmd/clair
+
+# Replace genesis manifests file in the source code with the one generated at
+# the point when the dump was taken.  This is to avoid discrepancy between other
+# files of the dump and the manifest.
+COPY ./blob-genesis_manifests.json image/scanner/dump/genesis_manifests.json
+
+FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG}
```
msugakov, line 25: [not about this line]
There seems to be `SHELL ["/bin/sh", "-o", "pipefail", "-c"]` defined in both upstream and downstream dockerfiles.
I don't know if there's a reason to define `pipefail` at the container level: I don't see any use of pipes below and I see that `entrypoint.sh` with its dependencies does call `set -...o pipefail` explicitly.

If you omit `SHELL ["/bin/sh", "-o", "pipefail", "-c"]` here, could you also please kill it from upstream dockerfiles for consistency? Later on, someone may take care of the downstream ones.

Alternatively let's add `SHELL ...` here (again: for consistency) and have someone eventually deal with it later.

msugakov (commented):
Haven't checked `.tekton` files. Will go through them next.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && source_branch.contains("rhtap")
+  creationTimestamp: null
+  labels:
+    appstudio.openshift.io/application: acs
+    appstudio.openshift.io/component: scanner
+    pipelines.appstudio.openshift.io/type: build
+  name: scanner-on-pull-request
+  namespace: rh-acs-tenant
+spec:
+  params:
```
msugakov, line 19: [not only about these lines]
Could you please reformat these `.tekton` files as done in stackrox/stacrkox, [roxctl](https://github.com/stackrox/stackrox/blob/8510071fb1dfa95026f4bc4d871e29afec6a3bda/.tekton/roxctl-pull-request.yaml#L29-L29) for example?

When there are blank lines between blocks and certain blocks are re-ordered, reading and reviewing the pipeline becomes much easier.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+      - name: rebuild
+        value: $(params.rebuild)
+      - name: skip-checks
+        value: $(params.skip-checks)
+      - name: skip-optional
+        value: $(params.skip-optional)
+      - name: pipelinerun-name
+        value: $(context.pipelineRun.name)
+      - name: pipelinerun-uid
+        value: $(context.pipelineRun.uid)
```
msugakov, line 150: All these can be removed for this `0.2` version of `task-init`. https://github.com/redhat-appstudio/build-definitions/blob/main/task/init/0.2/MIGRATION.md
```suggestion
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+      name: rebuild
+      type: string
+    - default: "false"
+      description: Skip checks against built image
+      name: skip-checks
+      type: string
+    - default: "true"
+      description: Skip optional checks, set false if you want to run optional checks
+      name: skip-optional
+      type: string
```
msugakov, line 98: Since `skip-optional` was used only in `task-init` and should be gone after it got updated, this can be removed too.
```suggestion
```

.tekton/scanner-pull-request.yaml
msugakov, line 19: 1. I once looked at `false` default value of `rebuild` param and concluded it's a premature optimization which prevents us from overwriting images with the same tags should we need to rebuild something.
Therefore, I suggest adding this to `params`

```
  - name: rebuild
    value: 'true'
```

2. Do you plan to enable gomod prefetching (`prefetch-input`) and `build-source-image` as part of this PR or subsequent PR(s)?

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+    - name: clone-repository
+      params:
+      - name: url
+        value: $(params.git-url)
+      - name: revision
+        value: $(params.revision)
+      - name: fetchTags
+        value: true
+      - name: depth
+        value: 0
```
msugakov, line 169: I'm not solid on YAML implicit type conversion rules and how the parser interprets them, but these params are declared as strings and it might be more canonical to make values be explicit strings

```suggestion
      - name: fetchTags
        value: "true"
      - name: depth
        value: "0"
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+        - name: bundle
+          value: quay.io/redhat-appstudio-tekton-catalog/task-prefetch-dependencies:0.1@sha256:0b7bec23b6c08f37138a86e569835842763b3aa42f4455fd70ba3986350e07c7
+        - name: kind
+          value: task
+        resolver: bundles
+      when:
+      - input: $(params.hermetic)
+        operator: in
+        values:
+        - "true"
```
msugakov, line 210: I had a long thread about this `when` block in `prefetch-dependencies`, and, in short, it's redundant and should be removed. :-)

```suggestion
```

This means: prefetching dependencies can be done (once you set `prefetch-input`) even without enabling hermetic builds.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+      - name: source
+        workspace: workspace
+    - name: fetch-vuln-feed-data
+      runAfter:
+      - init
+      - clone-repository
+      taskSpec:
+        steps:
+          - name: fetch-vuln-feed-data
+            image: registry.access.redhat.com/ubi8/ubi
```
msugakov, line 221: [minor] Let's tell it's `latest` for explicitness.
```suggestion
            image: registry.access.redhat.com/ubi8/ubi:latest
```

Also, `ubi8/ubi-minimal` should do just as well since it has `curl` OOTB. Up to you.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+            script: |
+              #!/usr/bin/env bash
```
msugakov, line 223: 1. Let's add the usual mantra for consistent scripting style and also to make bash behave more like a runtime with exceptions which fail the program when anything unexpected happened:

```suggestion
              #!/usr/bin/env bash
              set -euo pipefail
```

2. Since this script is used in two places, it will be best to create a `.sh` file and call that to simplify maintenance. Perhaps then it could auto-detect `$(workspaces.source.path)/source` location from its own path (with a bonus that the script can be called locally).

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+              for blob in "${blobs[@]}"; do
+                echo "https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob} > $(workspaces.source.path)/source/blob-${blob}"
+                curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
+                  --output "$(workspaces.source.path)/source/blob-${blob}" \
+                  "https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob}"
+
+                if [ "$?" != "0" ]; then
+                    echo "Failed to download"
+                    exit 1
+                fi
```
msugakov, line 240: With `set -euo pipefail`, this block can be removed and I think it's fine because you have `echo` command before `curl` and it will be clear what was happening before the `curl` and the script failed on a download problem.
```suggestion
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+              #!/usr/bin/env bash
+              mkdir -p "$(workspaces.source.path)/source"
+              blobs=(
+                nvd-definitions.zip
+                k8s-definitions.zip
+                repo2cpe.zip
+                genesis_manifests.json
+              )
+              for blob in "${blobs[@]}"; do
+                echo "https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob} > $(workspaces.source.path)/source/blob-${blob}"
```
msugakov, line 232: Minor optional suggestion:
```suggestion
                echo "Downloading https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob} > $(workspaces.source.path)/source/blob-${blob}"
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+                curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
+                  --output "$(workspaces.source.path)/source/blob-${blob}" \
+                  "https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob}"
```
msugakov, line 235: The `/latest/` part, unfortunately, is not always the case, see the [downstream script](https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-scanner/pre-build-script?ref_type=heads#L31).

Do you want to tackle this in the same PR or in a follow-up?

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+        value: $(tasks.clone-repository.results.commit)
+      runAfter:
+      - prefetch-dependencies
+      - fetch-vuln-feed-data
+      taskRef:
+        params:
+        - name: name
+          value: buildah-6gb
+        - name: bundle
+          value: quay.io/redhat-appstudio-tekton-catalog/task-buildah-6gb:0.1
```
msugakov, line 271: I think that the system may require to pin a digest for this image. I usually find those in the output of `docker pull quay.io/redhat-appstudio-tekton-catalog/task-buildah-6gb:0.1`.

```suggestion
          value: quay.io/redhat-appstudio-tekton-catalog/task-buildah-6gb:0.1@sha256:9c9b4081d873a217f0e20bcc7be3cd2ff96887f4c989439544640e3d2bc0e8b9
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+        resolver: bundles
+      when:
+      - input: $(params.skip-checks)
+        operator: in
+        values:
+        - "false"
+      workspaces:
+      - name: workspace
+        workspace: workspace
+    # - name: clamav-scan
```
msugakov, line 374: Regarding the disabled ClamAV scan, I quickly scrolled over [your thread](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1704986484024359) and got an impression that his is lack of CPU `requests` sticking out its ears again, right? We may need to come up with a plan how to solve this because I asked `1` CPU to be given to ClamAV but noticed it being rolled back.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+        resolver: bundles
+      when:
+      - input: $(params.skip-checks)
+        operator: in
+        values:
+        - "false"
+    workspaces:
+    - name: workspace
+    - name: git-auth
+      optional: true
```
msugakov, line 425: I suggest to make `git-auth` workspace/volume not optional so that the pipeline fails early when misconfigured rather than sporadically in cases GitHub decides to throttle requests.
```suggestion
```

msugakov (changes requested):
Marking as "request for changes" because of the amount of comments.

.containerignore
kylape: Thanks, I followed your suggestion.

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+ENV CI=1
+
+COPY . /src
+WORKDIR /src
+
+RUN unzip -j blob-repo2cpe.zip -d image/scanner/dump/repo2cpe && \
+    unzip -j blob-k8s-definitions.zip -d image/scanner/dump/k8s_definitions && \
+    unzip -j blob-nvd-definitions.zip -d image/scanner/dump/nvd_definitions
+
+RUN echo "version: $(git describe --tags --abbrev=10 --dirty --long)"; git status; go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(git describe --tags --abbrev=10 --dirty --long)" -o image/scanner/bin/scanner ./cmd/clair
```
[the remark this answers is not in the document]
kylape: https://github.com/stackrox/scanner/pull/1334/commits/1371932458919648a5fd4d34091b8a5bdc3e76ae

image/scanner/konflux/Dockerfile
kylape: You added `findutils` to the downstream dockerfile in commit 93cf4b3c7b356 (in the rhacs-midstream repo), but it looks like Ross fixed it in scanner by removing the calls to `find` in 28fbde2c5304640da6dd1ff08371ff102b605513.  So it should be safe to omit it from the dockerfile now.

image/scanner/konflux/Dockerfile
kylape: af81ebe

image/scanner/konflux/Dockerfile
kylape: Those extra files are `chown`'d in their [respective `COPY` lines](https://github.com/stackrox/scanner/blob/60685202bfdde41fb73d2d2fd90b06d75a005f28/image/scanner/konflux/Dockerfile#L53-L56).  According to https://issues.redhat.com/browse/RS-260, [scanner runs in the container as the nobody user](https://github.com/stackrox/stackrox/blob/e0911d2e77f8c3f0868cf45f267f3f0cead93a8c/image/templates/helm/shared/templates/02-scanner-06-deployment.yaml.htpl#L88).

image/scanner/konflux/Dockerfile
kylape: I tested without this line, and scanner was successfully invoked in the container as pid 1.  I checked the permissions of the `/scanner` file, and they were `755`.  So I'll remove this line from the konflux dockerfile.

image/scanner/konflux/Dockerfile
kylape: I'd rather avoid touching the upstream dockerfiles if I can help it, so I'll add the `SHELL` line to the konflux dockerfile.

kylape: d77527a

image/scanner/konflux/Dockerfile
kylape: d77527a

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,446 @@
...
+  name: scanner-on-pull-request
+  namespace: rh-acs-tenant
+spec:
+  params:
+  - name: dockerfile
+    value: image/scanner/konflux/Dockerfile
+  - name: git-url
+    value: '{{repo_url}}'
+  - name: image-expires-after
+    value: 5d
```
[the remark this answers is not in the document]
kylape: * Adjusted expire value for both pull requests and pushes
* Adjusted `max-keep-runs` to match stackrox images as well
* Copied over comments for TODOs
* I'll keep this conversation open until the end as a reminder to sync the two tekton pipeline files.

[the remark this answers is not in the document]
kylape: 67dd3cfe

.tekton/scanner-pull-request.yaml
kylape: 1. d5c70398
2. May as well do it in this PR.  I'll keep this conversation open until it's successfully added.

.tekton/scanner-pull-request.yaml
kylape: 225640345618f

.tekton/scanner-pull-request.yaml
kylape: https://issues.redhat.com/browse/ROX-22130

.tekton/scanner-pull-request.yaml
kylape: 2256403

.tekton/scanner-pull-request.yaml
kylape: I can't remember why I switched to the 6gb image, but I vaguely remember identifying something that was taking up memory and somehow removing it.  So I just tried to switch back to the standard buildah image, and it worked just fine :shrug: (ffe1f37)

.tekton/scanner-pull-request.yaml
kylape: 8580b71

.tekton/scanner-pull-request.yaml
kylape: c6d3016

.tekton/scanner-pull-request.yaml
kylape: 6e1c579

.tekton/scanner-pull-request.yaml
kylape: 6e1c579

.tekton/scanner-pull-request.yaml
kylape: 6e1c579

.tekton/scanner-pull-request.yaml
kylape: Fix here: 6e1c579

I didn't have the script auto-detect the target path, but I required it as a parameter to the script.

.tekton/scanner-pull-request.yaml
kylape: 2256403

.tekton/scanner-pull-request.yaml
kylape: a3091e84fbc77a13e7db781

.tekton/scanner-pull-request.yaml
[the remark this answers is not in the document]
kylape: pipeline sync: a8d2ec41

scripts/konflux/version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# This script is used by the Konflux dockerfile to get the correct version to
+# compile in to the binary.  If HEAD points to a git tag, use that; otherwise
+# use `git describe ...`.
+
```
msugakov, line 6: [nit]
```suggestion

set -euo pipefail

```

mainly for consistency with other scripts.

scripts/konflux/version.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
msugakov, line 1: [also consistency nit]
```suggestion
#!/usr/bin/env bash
```

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+COPY . /src
+WORKDIR /src
+
+RUN scripts/konflux/fail-build-if-git-is-dirty.sh
+
+RUN unzip -j blob-repo2cpe.zip -d image/scanner/dump/repo2cpe && \
+    unzip -j blob-k8s-definitions.zip -d image/scanner/dump/k8s_definitions && \
+    unzip -j blob-nvd-definitions.zip -d image/scanner/dump/nvd_definitions
+
+RUN echo "version: $(scripts/konflux/version.sh)" && \
```
msugakov, line 20: It could be
```suggestion
RUN echo -n "version: " && scripts/konflux/version.sh && \
```
but really up to you. I just being pedantic.

image/scanner/konflux/Dockerfile
msugakov: Thanks for investigating this. It has been a long time and I forgot.

image/scanner/konflux/Dockerfile
msugakov: Ah, that makes sense: I did not pay attention that there's `COPY` with `--chown`.

What bugs me about the ownership change: the container is supposed to run on the cluster with read-only root filesystem so these files are not supposed to be modified. What that means to me, the desire is to open the files for _reading_ by the executing user (uid=65534). In such case, giving ownership is overly permissive, it could be sufficient to do `chown`. It could be tightened up with some regression testing but it's certainly not in scope of this work here, we could just make sure we don't make it worse.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+kind: PipelineRun
+metadata:
+  annotations:
+    build.appstudio.openshift.io/repo: https://github.com/stackrox/scanner?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: '{{revision}}'
+    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
+    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    pipelinesascode.tekton.dev/max-keep-runs: "500"
+    # TODO(ROX-21073): re-enable for all PR branches
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && source_branch.contains("rhtap")
```
msugakov, line 11: I spotted that you used `konflux` in branch name for `scanner-db`. Maybe it's convenient to allow either of these. I'd optionally suggest (untested):
```suggestion
    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("konflux"))
```

scripts/konflux/fetch-vuln-feed-data.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ "$#" < 1 ]]; then
+  echo "Please pass target directory."
+  exit 1
+fi
+
+TARGET_DIR=$1
```
msugakov, line 10: It's good to quote substitutions for consistent style and to avoid issues when paths contain spaces or other funny stuff.

```suggestion
TARGET_DIR="$1"
```

scripts/konflux/fetch-vuln-feed-data.sh
```diff
@@ -0,0 +1,29 @@
...
+blobs=(
+  nvd-definitions.zip
+  k8s-definitions.zip
+  repo2cpe.zip
+  genesis_manifests.json
+)
+
+for blob in "${blobs[@]}"; do
+  echo "Downloading https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob} > $TARGET_DIR/blob-${blob}"
+  curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
```
msugakov, line 26: In order to fully close on the comment [https://github.com/stackrox/scanner/pull/1334#discussion_r1469457260], I suggest adding a TODO comment here as another reminder.

```suggestion
  # TODO(ROX-22130): Assign proper suffix for tagged commits instead of /latest/.
  curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
```

.tekton/scanner-pull-request.yaml
msugakov: I suspect that absent ClamAV check will fail EC. Could you please check? If that's the case, we shouldn't merge the PR until we find some solution in order not to make the EC more red.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+    - name: fetch-vuln-feed-data
+      runAfter:
+      - init
+      - clone-repository
+      taskSpec:
+        steps:
+          - name: fetch-vuln-feed-data
+            image: registry.access.redhat.com/ubi8-minimal:latest
+            script: |
+              #!/usr/bin/env bash
```
msugakov, line 245: Maybe shebang is even redundant in this case, see <https://tekton.dev/docs/pipelines/tasks/#running-scripts-within-steps>. Up to you.
```suggestion
```

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+              $(workspaces.source.path)/source/scripts/konflux/fetch-vuln-feed-data.sh "$(workspaces.source.path)/source"
```
msugakov, line 246: I think, the executable can be quoted just as well as its argument. Untested.

```suggestion
              "$(workspaces.source.path)/source/scripts/konflux/fetch-vuln-feed-data.sh" "$(workspaces.source.path)/source"
```

msugakov (commented):
Everything from this round is minor.
ClamAV is something that can block the merge (semantically, not technically) if it fails EC (I left a separate comment about it).

scripts/konflux/fetch-vuln-feed-data.sh
msugakov: FYI: I sort-of forked the script in <https://github.com/stackrox/scanner/pull/1387>. Maybe you'd be interested to use that as a base.

.tekton/scanner-pull-request.yaml
kylape: I chose to leave it in.

.tekton/scanner-pull-request.yaml
kylape: I looked at the logs from https://console.redhat.com/preview/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/acs-wqkrn-sslkb/logs, and I don't see anything referencing `scanner` specifically, only `scanner-db`.  That said, [this PR](https://github.com/redhat-appstudio/build-definitions/pull/774) looks promising to resolve the ClamAV issue.  Perhaps it would be simpler to just wait until that gets merged.

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+    #     value: $(tasks.build-container.results.IMAGE_URL)
+    #   runAfter:
+    #   - build-container
+    #   timeout: 3h
+    #   taskRef:
+    #     params:
+    #     - name: name
+    #       value: clamav-scan
+    #     - name: bundle
+    #       value: quay.io/redhat-appstudio-tekton-catalog/task-clamav-scan:0.1@sha256:353fa2cda9855217cfcec3303973b666a10f384795630cf0eb13b874c24b0f7a
```
msugakov, line 397: Re-enabling ClamAV seems to be the only remaining thing. As I wrote in Slack, the recent task fixed issues for `scanner-db` and I'm hopeful it can fix things for scanner.

```suggestion
    #       value: quay.io/redhat-appstudio-tekton-catalog/task-clamav-scan:0.1@sha256:3a546fe7c07fc24ecf3bddef88a7a9a66b525c4dd53acc8b5ea13069f60c086d
```

kylape: Looks like the update to the clamav image wokred :tada:

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+
+    - name: clamav-scan
+      params:
+      - name: image-digest
+        value: $(tasks.build-container.results.IMAGE_DIGEST)
+      - name: image-url
+        value: $(tasks.build-container.results.IMAGE_URL)
+      runAfter:
+      - build-container
+      timeout: 3h
```
msugakov, line 391: I suggest to leave it out and rely on the default timeout here. The default seems to be 1 hour.
```suggestion
```
Up to you.

.tekton/scanner-push.yaml
```diff
@@ -0,0 +1,434 @@
...
+
+    - name: clamav-scan
+      params:
+      - name: image-digest
+        value: $(tasks.build-container.results.IMAGE_DIGEST)
+      - name: image-url
+        value: $(tasks.build-container.results.IMAGE_URL)
+      runAfter:
+      - build-container
+      timeout: 3h
```
msugakov, line 390: Similarly
```suggestion
```

msugakov (approved):

.tekton/scanner-pull-request.yaml
```diff
@@ -0,0 +1,435 @@
...
+        resolver: bundles
+      when:
+      - input: $(params.skip-checks)
+        operator: in
+        values:
+        - "false"
+  timeouts:
+    pipeline: "4h0m0s"
+    tasks: "3h0m0s"
+    finally: "1h0m0s"
```
msugakov, line 432: Actually, unless you have an **intent** behind these, I suggest removing them too. I find it difficult find reasoning for code that is just there without comments or tests. Is it there on purpose? What would break if it's changed or removed? The latter bothers me a lot when nothing breaks immediately but weird edge cases come over time.

```suggestion
```

.tekton/scanner-push.yaml
```diff
@@ -0,0 +1,434 @@
...
+        resolver: bundles
+      when:
+      - input: $(params.skip-checks)
+        operator: in
+        values:
+        - "false"
+  timeouts:
+    pipeline: "4h0m0s"
+    tasks: "3h0m0s"
+    finally: "1h0m0s"
```
msugakov, line 431: Similarly,
```suggestion
```

msugakov:
@kylape Here's what I meant by unifying the script - <https://github.com/stackrox/scanner/pull/1395>. Note that I had to rebase on `master` therefore all commits show up as new. Only the last two commits are significant.

I suggest you merge this <https://github.com/stackrox/scanner/pull/1334> as-is, and then either of us can refresh <https://github.com/stackrox/scanner/pull/1395> and publish for review. This way we don't hold #1334 from merging.

image/scanner/konflux/Dockerfile
```diff
@@ -0,0 +1,77 @@
+ARG BASE_REGISTRY=registry.access.redhat.com
```
msugakov, line 1: [about the file path]
I realized there's now inconsistency. This file is
`image/scanner/konflux/Dockerfile`
whereas `scanner-db`-s is
`image/db/rhel/konflux.Dockerfile`

I suggest moving this file to `image/scanner/rhel/konflux.Dockerfile` for consistency. WDYT?

kylape: I agree.

msugakov (approved):

PR #1347, opened 2023-12-11
ROX-21124: Move non-e2e testing OSCI jobs to GitHub Actions

BradLugo:
# Description
These changes move the OpenShift CI jobs into GitHub Actions and change the updater to use an API token for grabbing NVD data.

Some interesting side-effects from this work:
- We're able to download artifacts such as the genesis-dump directly from CI (i.e., without needing to pull the image)
- We can run scanner code related CI (excluding e2e tests) without needing to wait for the data processing

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,5 +1,7 @@
 name: CI
 on:
+  schedule:
+    - cron: '0 */3 * * *'
   push:
     branches:
       - master
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,4 +1,4 @@
-name: Build
+name: CI
 on:
   push:
     branches:
@@ -401,3 +401,91 @@ jobs:
 
       - name: diff-dumps
         run: ./scripts/ci/jobs/diff-dumps.sh
+
+  upload-db-dump:
+    # Only run on master branch
+    if: github.ref == 'refs/heads/master'
+    env:
+      GOOGLE_SA_CIRCLECI_SCANNER: ${{ secrets.GOOGLE_SA_CIRCLECI_SCANNER }}
+    runs-on: ubuntu-latest
+    needs:
+      - generate-db-dump
+    container:
+      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: db-dump
+          path: /tmp/postgres
+
+      - name: upload-db-dump
+        run: ./scripts/ci/jobs/upload-db-dump.sh
+
+  upload-dumps-for-downstream:
+    # Only run on master branch
+    if: github.ref == 'refs/heads/master'
+    env:
+      GOOGLE_SA_STACKROX_HUB_VULN_DUMP_UPLOADER: ${{ secrets.GOOGLE_SA_STACKROX_HUB_VULN_DUMP_UPLOADER }}
+    runs-on: ubuntu-latest
+    needs:
+      - generate-db-dump
+    container:
+      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: vuln-dump
+          path: /tmp/vuln-dump
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: db-dump
+          path: /tmp/postgres
+
+      - name: upload-dumps-for-downstream
+        run: ./scripts/ci/jobs/upload-dumps-for-downstream.sh
+
+  upload-dumps-for-embedding:
+    # Only run on master branch
+    if: github.ref == 'refs/heads/master'
+    env:
+      GOOGLE_SA_CIRCLECI_SCANNER: ${{ secrets.GOOGLE_SA_CIRCLECI_SCANNER }}
+    runs-on: ubuntu-latest
+    needs:
+      - generate-db-dump
+    container:
+      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: vuln-dump
+          path: /tmp/vuln-dump
+
+      - name: upload-dumps-for-downstream
+        run: ./scripts/ci/jobs/upload-dumps-for-embedding.sh
+
+
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -65,8 +65,6 @@ jobs:
           path: scanner-build.tgz
 
   style-check:
-    env:
-      ARTIFACT_DIR: /artifacts
     runs-on: ubuntu-latest
     container:
       image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
@@ -82,15 +80,10 @@ jobs:
       - name: Cache Go dependencies
         uses: ./.github/actions/cache-go-dependencies
 
-      - name: Create artifacts dir
-        run: mkdir -p "$ARTIFACT_DIR"
-
       - name: Run style checks
         run: ./scripts/ci/jobs/style-checks.sh
 
   unit-tests:
-    env:
-      ARTIFACT_DIR: /artifacts
     runs-on: ubuntu-latest
     needs:
       - pre-build-scanner
@@ -107,9 +100,6 @@ jobs:
 
       - name: Cache Go dependencies
         uses: ./.github/actions/cache-go-dependencies
-
-      - name: Create artifacts dir
-        run: mkdir -p "$ARTIFACT_DIR"
 
       - uses: actions/download-artifact@v3
         with:
@@ -123,8 +113,6 @@ jobs:
         run: ./scripts/ci/jobs/unit-tests.sh
 
   db-integration-tests:
-    env:
-      ARTIFACT_DIR: /artifacts
     runs-on: ubuntu-latest
     needs:
       - pre-build-scanner
@@ -141,9 +129,6 @@ jobs:
 
       - name: Cache Go dependencies
         uses: ./.github/actions/cache-go-dependencies
-
-      - name: Create artifacts dir
-        run: mkdir -p "$ARTIFACT_DIR"
 
       - uses: actions/download-artifact@v3
         with:
@@ -416,62 +401,3 @@ jobs:
 
       - name: diff-dumps
         run: ./scripts/ci/jobs/diff-dumps.sh
-
-  # TODO(DO NOT MERGE): These store functions don't seem quite right
-  store-genesis-dump:
-    env:
-      ARTIFACT_DIR: /artifacts
-    runs-on: ubuntu-latest
-    needs:
-      - generate-genesis-dump
-    container:
-      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
-
-      - uses: ./.github/actions/job-preamble
-
-      - name: Create artifacts dir
-        run: mkdir -p "$ARTIFACT_DIR"
-
-      - uses: actions/download-artifact@v3
-        with:
-          name: genesis-dump
-          path: /tmp/genesis-dump
-
-      - name: store-genesis-dump
-        run: |
-          ./scripts/ci/jobs/store-genesis-dump.sh
-
-  store-db-dump:
-    env:
-      ARTIFACT_DIR: /artifacts
-    runs-on: ubuntu-latest
-    needs:
-      - generate-db-dump
-    container:
-      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
-
-      - uses: ./.github/actions/job-preamble
-
-      - name: Create artifacts dir
-        run: mkdir -p "$ARTIFACT_DIR"
-
-      - uses: actions/download-artifact@v3
-        with:
-          name: db-dump
-          path: /tmp/postgres
-
-      - name: store-db-dump
-        run: |
-          ./scripts/ci/jobs/store-db-dump.sh
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,7 +29,7 @@ jobs:
       - name: Build updater
         run: make build-updater
 
-      - name: Bundle the build to preserve permissions
+      - name: Archive the build to preserve permissions
         run: tar -cvzf updater-build.tgz bin/updater
 
       - uses: actions/upload-artifact@v3
@@ -56,7 +56,7 @@ jobs:
       - name: Build Scanner
         run: make scanner-build-nodeps
 
-      - name: Bundle the build to preserve permissions
+      - name: Archive the build to preserve permissions
         run: tar -cvzf scanner-build.tgz image/scanner/bin/scanner
 
       - uses: actions/upload-artifact@v3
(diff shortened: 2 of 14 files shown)


scripts/ci/lib.sh
```diff
@@ -819,6 +819,74 @@ send_slack_notice_for_vuln_check_failure() {
...
+
+    ls -lrt /tmp/postgres || info "No local DB dump"
+
+    if is_in_PR_context && ! pr_has_label "generate-dumps-on-pr"; then
+        info "Label generate-dumps-on-pr not set. Pulling dumps from GCS bucket"
+        gsutil cp gs://stackrox-scanner-ci-vuln-dump/pg-definitions.sql.gz image/db/dump/definitions.sql.gz
+    else
+        cp /tmp/postgres/pg-definitions.sql.gz image/db/dump/definitions.sql.gz
+    fi
+}
```
BradLugo, line 888: I debated putting these in `scripts/ci/jobs/<relevant-name>`

.github/workflows/scheduled-build.yaml
BradLugo, line 1: I was planning on basically copying `.github/workflows/build.yaml`, then adding the `upload-dumps-for-downstream` and `upload-dumps-for-embedding`, but that seems a bit naive. Do y'all have any suggestions? One thought I had was splitting up `.github/workflows/build.yaml` into a bunch of actions and reusing those, but we'd need every job as an action, which would also be a bit messy.

BradLugo, line 1: I was planning on basically copying `.github/workflows/build.yaml`, then adding the `upload-dumps-for-downstream` and `upload-dumps-for-embedding`, but that seems a bit naive. Do y'all have any suggestions? One thought I had was splitting up `.github/workflows/build.yaml` into a bunch of actions and reusing those, but we'd need every job as an action, which would also be a bit messy.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+
+      - name: Get genesis dump
+        run: |
+          source ./scripts/ci/lib.sh
+          get_genesis_dump
+
+      - name: Make bundle
+        run: image/scanner/rhel/create-bundle.sh image/scanner image/scanner/rhel
+
+      - name: Bundle the bundle to preserve permissions # TODO(DO NOT MERGE): word this better?
```
RTann, line 277: lol it's accurate. it's honestly ok with me. you can also switch the first bundle with archive

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+
+      - name: Build scanner-db image
+        run: |
+          docker build -t scanner-db:"$(make --quiet tag)" -f image/db/rhel/Dockerfile image/db/rhel
+
+      - name: Build scanner-db-slim image
+        run: |
+          docker build -t scanner-db-slim:"$(make --quiet tag)" -f image/db/rhel/Dockerfile.slim image/db/rhel
+
+        # needed for docs ensure_image.sh initial pull with RHACS_BRANDING
```
RTann, line 370: we don't have branding. is this still required?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: genesis-dump
+          path: /tmp/genesis-dump
+
+      - name: diff-dumps
+        run: ./scripts/ci/jobs/diff-dumps.sh
+
+  # TODO(DO NOT MERGE): These store functions don't seem quite right
```
RTann, line 421: what's wrong with them?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -100,7 +100,7 @@ diff_dumps() {
 
     # These are not needed until later, but no reason to continue if these do not exist.
     require_environment "GOOGLE_SA_STACKROX_HUB_VULN_DUMP_UPLOADER"
-    require_environment "GCP_SERVICE_ACCOUNT_CREDS"
+    require_environment "SCANNER_GCP_SERVICE_ACCOUNT_CREDS"
```
RTann, line 103: assuming you added this to the repo?

scripts/ci/lib.sh
```diff
@@ -819,6 +819,74 @@ send_slack_notice_for_vuln_check_failure() {
     curl -XPOST -d @- -H 'Content-Type: application/json' "$webhook_url"
 }
 
+generate_genesis_dump() {
```
RTann, line 822: it seems like we do this unconditionally. This should use the same conditionals as the others below. No need to run this in PRs which don't ask for it

scripts/ci/lib.sh
RTann: eh this is fine

RTann (commented):
THANK YOU for doing this. This is greatly appreciated

RTann:
Build / generate-db-dump (pull_request) Successful in 44m

I missed this

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - name: Cache Go dependencies
+        uses: ./.github/actions/cache-go-dependencies
+
+      - name: Create artifacts dir
+        run: mkdir -p "$ARTIFACT_DIR"
```
gavin-stackrox, line 86: I took a deeper look at this `ARTIFACT_DIR`. I think you can remove it from this PR. It is only used in `store_test_results` and that returns for non openshift environments without doing anything. 

The only reason to save those Junit test results would be to display them with something like https://github.com/test-summary/action which could be done in a follow on PR if warranted (I don't think there is enough ongoing dev in stackrox/scanner to warrant it IMHO).

gavin-stackrox (commented):
I have some questions about testing which I'll bring up elsewhere.

scripts/ci/postgres.sh
```diff
@@ -8,15 +8,33 @@ source "$SCRIPTS_ROOT/scripts/lib.sh"
...
+
+    if [[ $(id -u) == 0 ]]; then
+        info "This function should not be run as root."
+        if is_CI; then
+            info "Running in CI. Creating a non-root user."
+            groupadd -g 1001 pg
+            adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
+
+            # The PATH is not completely preserved, so set the PATH here to ensure postgres-related commands can be found.
+            runuser -l pg -c "PATH=$PATH $SCRIPTS_ROOT/scripts/ci/postgres.sh _start_postgres" # TODO(DO NOT MERGE): this is a mess
```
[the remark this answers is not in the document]
BradLugo: I don't love calling `_start_postgres` like this (I don't really even like the name tbh), but I don't fully understand idiomatic bash so :shrug:. I also see some scripts use `su <user> ...` so I wasn't sure which should be used here. I'm fine to keep this as-is if y'all think it's fine

scripts/ci/jobs/diff-dumps.sh
BradLugo: That's correct - added at the organization level, then configured for this repo. The environment variables with the corresponding secrets are set in the workflow.yaml.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - name: Cache Go dependencies
+        uses: ./.github/actions/cache-go-dependencies
+
+      - name: Build Scanner
+        run: make scanner-build-nodeps
```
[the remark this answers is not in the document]
BradLugo: Honestly, I think `scanner-build-nodeps` is a bit of a misnomer. AFAICT, it still grabs the dependencies anyway. That being said, now that I look at the `deps` target, it might be worth adding that step first (since it has some other dependent targets). The main reason I used `scanner-build-nodeps` (without `deps`) was because that's how we do it in `stackrox/stackrox`

.github/workflows/build.yaml
```diff
@@ -0,0 +1,478 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: ./.github/actions/job-preamble
+
+      - name: Cache Go dependencies
```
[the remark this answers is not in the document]
BradLugo: In a sense, both. If the jobs run with new or changed dependencies, it would create a cache, but it would typically fetch from the cache. `.github/actions/cache-go-dependencies/action.yaml` should have the exact details on how that process works but the reason we need the Go cache in this job is because of the `golangci-lint` make target (actual name is `$(GOLANGCILINT_BIN)`)

.github/workflows/build.yaml
BradLugo: I changed the implementation of `store_test_results` in this branch, but yeah, that's actually why I had comment at https://github.com/stackrox/scanner/pull/1347/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R370. Would it make sense to remove all the usage of `ARTIFACT_DIR` _and_ all the jobs that only call `store_test_results`?

.github/workflows/build.yaml
BradLugo: Basically what Gavin brought up here: https://github.com/stackrox/scanner/pull/1347/files#r1433028812
I'm not sure if these store functions are actually needed in GHA

.github/workflows/build.yaml
gavin-stackrox: Ah I see. I missed that. I'd say remove `store_test_results()` from any test/job that is moved to GHA. The jobs that remain in OSCI will still benefit from it.

.github/workflows/build.yaml
gavin-stackrox: They should be removed. The artifacts are stored earlier (and available through the GH UI) via upload-artifact.

.github/workflows/scheduled-build.yaml
gavin-stackrox, line 1: Can you remove this empty file, it is triggering an action failure.

scripts/ci/lib.sh
BradLugo: The rest of the workflow depends on the artifact this function produces. If we make this conditional, I think we'd need to supply some sort of `genesis-dump.zip` and `vuln-dump`. Would you want to grab those from a previous run or from the latest `master` run?

fwiw, it looks like we used to do this unconditionally in OSCI

BradLugo:
fysa, `upload-db-dump`, `upload-dumps-for-downstream`, and `upload-dumps-for-embedding` are untested. Do y'all have any suggestions on how to test them?

scripts/ci/jobs/diff-dumps.sh
BradLugo: We could also not change the name here and instead bind the secret `secrets.SCANNER_GCP_SERVICE_ACCOUNT_CREDS` to the environment variable `GCP_SERVICE_ACCOUNT_CREDS`

.github/workflows/build.yaml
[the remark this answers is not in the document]
RTann: let's keep it as-is to stay aligned with stackrox/stackrox

.github/workflows/CI.yaml
RTann, line 1: nit: can we keep file names lowercase?

scripts/ci/lib.sh
RTann: In OSCI we halt certain jobs based on this gating JSON: https://github.com/stackrox/scanner/blob/master/scripts/ci/gate-jobs-config.json

BradLugo: Oh, I see. Would this mean that images aren't generated/pushed unless they have the `generate-dumps-on-pr` label? Is that ok?

RTann: this step can take a while, so in PRs we don't generate the genesis dump (unless a specific label is specified). instead, we pull the dumps from GCP. I think this can be seen in get_genesis_dump()

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,513 @@
...
+          name: scanner-build
+
+      - name: Unpack scanner build
+        run: |
+          tar xvzf scanner-build.tgz
+
+      - uses: actions/download-artifact@v3
+        if: |
+          github.event_name != 'pull_request' ||
+          contains(github.event.pull_request.labels.*.name, 'generate-dumps-on-pr')
```
BradLugo, line 260: Could also be `needs.generate-genesis-dump.result == 'success'` maybe? Probably doesn't matter that much

BradLugo:
Needs to be rebased once https://github.com/stackrox/scanner/pull/1364 is merged.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,514 @@
...
+          path: /tmp/postgres/pg-definitions.sql.gz
+
+  generate-scanner-bundle:
+    runs-on: ubuntu-latest
+    needs:
+      - pre-build-scanner
+      - generate-genesis-dump
+    if: |
+      always() &&
+      (needs.generate-genesis-dump.result == 'success' || needs.generate-genesis-dump.result == 'skipped')
```
RTann, line 237: when is `always()` false?

BradLugo: `always()` will never be false, but I think I know what you're getting at. Here's what happens when you don't have `always()`: https://github.com/stackrox/scanner/actions/runs/7468175561

My understanding from the [documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#always) is the condition won't even execute unless `always()` is present

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,514 @@
...
+          tar xvzf bundle.tgz
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: scanner-db-bundle
+          path: image/db/rhel
+
+      - name: Build scanner image
+        run: |
+          docker build -t scanner:"$(make --quiet tag)" -f image/scanner/rhel/Dockerfile image/scanner/rhel
```
RTann, line 362: want to throw a `--no-print-directory` in all these calls to `make tag`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,514 @@
...
+      - name: Build scanner-db-slim image
+        run: |
+          docker build -t scanner-db-slim:"$(make --quiet tag)" -f image/db/rhel/Dockerfile.slim image/db/rhel
+
+      - name: Docker login
+        # Skip for external contributions.
+        if: |
+          github.event_name == 'push' || !github.event.pull_request.head.repo.fork
+        run: |
+          docker login -u "${QUAY_RHACS_ENG_RO_USERNAME}" --password-stdin quay.io <<<"${QUAY_RHACS_ENG_RO_PASSWORD}"
```
RTann, line 381: what's the purpose of this step? Do we need the `RO` creds?

RTann (approved):
Assuming this only covers master and PRs, then looks good to me! We still need to support release builds and nightlies (nightlies not as important at this time)

.github/workflows/ci.yaml
BradLugo: To perform the `docker push` in the next step. I'm not sure why it's `RO`, but these steps were copied from `stackrox/stackrox`

.github/workflows/ci.yaml
BradLugo: I'm down, though I'm not entirely sure if adding that flag will do anything since we aren't switching directories

scripts/ci/lib.sh
```diff
@@ -819,6 +819,74 @@ send_slack_notice_for_vuln_check_failure() {
...
+    unzip -d image/scanner/dump /tmp/vuln-dump/repo2cpe.zip
+}
+
+generate_db_dump() {
+    info "Generating DB dump"
+
+    groupadd -g 1001 pg
+    adduser pg -u 1001 -g 1001 -d /var/lib/postgresql -s /bin/sh
+
+    # The PATH is not completely preserved, so set the PATH here to ensure postgres-related commands can be found.
```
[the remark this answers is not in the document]
BradLugo: It could be, but I am not entirely sure. I'm also down to leave it be

RTann (commented):
@BradLugo is the genesis dump accessible to a user who would want to download it? When cutting a new release, we will need access to the latest one generated from the master branch. Just remembered this

RTann:
> @BradLugo is the genesis dump accessible to a user who would want to download it? When cutting a new release, we will need access to the latest one generated from the master branch. Just remembered this

can be done in a followup PR, if not already done here

BradLugo:
> @BradLugo is the genesis dump accessible to a user who would want to download it? When cutting a new release, we will need access to the latest one generated from the master branch. Just remembered this

@RTann Would this suffice? https://github.com/stackrox/scanner/actions/runs/7292017884

RTann:
> > @BradLugo is the genesis dump accessible to a user who would want to download it? When cutting a new release, we will need access to the latest one generated from the master branch. Just remembered this
> 
> @RTann Would this suffice? https://github.com/stackrox/scanner/actions/runs/7292017884

yep!

BradLugo:
/override ci/prow/images ci/prow/style-checks ci/prow/unit-tests

BradLugo:
All `stackrox/stackrox` tests past: https://github.com/stackrox/stackrox/pull/9146

BradLugo:
e2e tests are fixed here: https://github.com/stackrox/scanner/pull/1368

PR #1348, opened 2023-12-12
chore(nvd): add more retries

RTann:
This PR bumps the number of retries from 3 to 5, and it also retries upon JSON decode error.


dcaravel (approved):
LGTM

PR #1371, opened 2024-01-16
ROX-21124: Fix upload-dumps-for-downstream GHA job

BradLugo:


BradLugo:
Let me know if there's a good way to test this, but it looks like this job is only expected to run on master. I'm not sure what the consequences are for overriding that behavior.

scripts/ci/lib.sh
```diff
@@ -282,6 +282,13 @@ get_base_ref() {
         else
             die "Expect PULL_BASE_REF or CLONEREFS_OPTIONS"
         fi
+    elif is_GITHUB_ACTIONS; then
+        # GITHUB_BASE_REF is only set for PRs. Use GITHUB_REF_NAME for Prow's PULL_BASE_REF equivalent.
```
RTann, line 286: prow?

RTann: oh never mind i misread

RTann (approved):
🚢

PR #1376, opened 2024-01-17
ROX-21124: Move release branch jobs from OSCI to GHA

BradLugo:
## Description

These changes enable GitHub Action workflows for jobs that were previously running in OSCI. Once merged, any new tags containing these changes will have OSCI configs that look more like master's than the other release branch OSCI configs.

Note for reviewers: here's the OSCI config that gates certain jobs from running in certain contexts: https://github.com/stackrox/scanner/blob/master/scripts/ci/gate-jobs-config.json. We should generally try to follow this for now unless we see a compelling reason not to.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -537,7 +537,6 @@ jobs:
   upload-dumps-for-downstream:
     # Only run this step on the master branch or any tags
     # Note that our scheduled jobs run on the master branch
-    # Run on master branch or tags
     if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/')
     env:
       GOOGLE_SA_STACKROX_HUB_VULN_DUMP_UPLOADER: ${{ secrets.GOOGLE_SA_STACKROX_HUB_VULN_DUMP_UPLOADER }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,9 @@ on:
   push:
     branches:
       - master
+      - 'release-*'
+    tags:
+      - '**'
   pull_request:
     types:
       - opened
@@ -149,6 +152,7 @@ jobs:
         run: ./scripts/ci/jobs/db-integration-tests.sh
 
   generate-genesis-dump:
+    # Run this job if it's not a PR or the PR contains the `generate-dumps-on-pr` label
     if: |
       github.event_name != 'pull_request' ||
       contains(github.event.pull_request.labels.*.name, 'generate-dumps-on-pr')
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - 'release-*'
```
RTann, line 8: I think we may need to run these things upon tag creation, too? Not sure how to convert the OSCI release process into GHA release

RTann (commented):
I don't think we want to run `diff-dumps` in `release-*` branches. Just master and PR with the `generate-dumps-on-pr` label.

We also want to run `upload-dumps-for-downstream` upon tag (would that be in a separate PR?)

.github/workflows/ci.yaml
```diff
@@ -264,6 +271,8 @@ jobs:
           tar xvzf scanner-build-${{ matrix.goos }}-${{ matrix.goarch }}.tgz
 
       - uses: actions/download-artifact@v4
+        # Run this step if it's not a PR or the PR containers the `generate-dumps-on-pr` label
```
dcaravel, line 274:
```suggestion
        # Run this step if it's not a PR or the PR contains the `generate-dumps-on-pr` label
```

.github/workflows/ci.yaml
```diff
@@ -440,15 +453,21 @@ jobs:
         run: |
           source ./scripts/ci/lib.sh
 
+          # If this is updated, be sure to update goarch in define-scanner-job-matrix above.
```
dcaravel, line 456:
```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -476,7 +495,6 @@ jobs:
           tar xvzf updater-build.tgz
 
       - uses: actions/download-artifact@v4
-        if: ${{ ! startsWith(github.ref, 'refs/tags/') }}
```
dcaravel, line 479: Did this condition cause the subsequent steps to fail if it was run for a tag?

.github/workflows/ci.yaml
```diff
@@ -503,22 +522,23 @@ jobs:
...
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: master
 
       - uses: ./.github/actions/job-preamble
 
       - uses: actions/download-artifact@v4
         with:
-          name: db-dump
+          name: db-dump`
```
dcaravel, line 531: \` a typo?

.github/workflows/ci.yaml
```diff
@@ -503,22 +522,23 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: master
```
dcaravel, line 525: Should we keep `fetch-depth: 0` here?

.github/workflows/ci.yaml
BradLugo: I don't recall specifically why I removed this, but it looks like this job (`diff-dumps`) won't/shouldn't run on tags anyway.

PR #1396, opened 2024-02-06
(chore): Setup `CODEOWNERS` for `.tekton` directory

msugakov:
Copied content from `stackrox` https://github.com/stackrox/stackrox/blob/master/.github/CODEOWNERS and hope it would just work.


tommartensen (approved):
> This CODEOWNERS file is valid.

LGTM!

PR #1400, opened 2024-02-07
Include latest v4 vulnerability

daynewlee:
Generate scanner-vuln-update including v4 vulnerability of latest version
 Currently I can only verify scanner v2 work with new **_scanner-vuln-updates.zip_**, more V4 related tests will be done in ROX-21981
## Test
1. Make sure the zip is containing scanner v4 stuff:
```
ls -lh scanner-vuln-updates.zip 
-rw-r--r-- 1 yili staff 272M Feb 14 15:04 scanner-vuln-updates.zip
```
```
unzip -l scanner-vuln-updates.zip
Archive:  scanner-vuln-updates.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
140154404  02-14-2024 11:41   scanner-defs.zip
   248088  02-14-2024 11:41   k8s-istio.zip
144211510  02-14-2024 11:41   scanner-v4-defs-4.3.zip
---------                     -------
284614002                     3 files
```

2. Upload to Central:
```
yli3-mac:stackrox yili$ curl --insecure -u admin:[PWD] -F "file=@scanner-vuln-updates.zip" https://localhost:8000/api/extensions/scannerdefinitions -v
*   Trying [::1]:8000...
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Request CERT (13):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Certificate (11):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-AES128-GCM-SHA256
* ALPN: server accepted h2
* Server certificate:
*  subject: OU=CENTRAL_SERVICE; CN=CENTRAL_SERVICE: Central; serialNumber=16561514245731368659
*  start date: Feb 14 20:08:00 2024 GMT
*  expire date: Feb 13 21:08:00 2025 GMT
*  issuer: CN=StackRox Certificate Authority; serialNumber=1018846594209314016
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* using HTTP/2
* Server auth using Basic with user 'admin'
* [HTTP/2] [1] OPENED stream for https://localhost:8000/api/extensions/scannerdefinitions
* [HTTP/2] [1] [:method: POST]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: localhost:8000]
* [HTTP/2] [1] [:path: /api/extensions/scannerdefinitions]
* [HTTP/2] [1] [authorization: Basic YWRtaW46SzNEWGJmZ3NlNVR0UzQ3c01ibG5PZTBqVQ==]
* [HTTP/2] [1] [user-agent: curl/8.4.0]
* [HTTP/2] [1] [accept: */*]
* [HTTP/2] [1] [content-length: 284614740]
* [HTTP/2] [1] [content-type: multipart/form-data; boundary=------------------------GlBMII0WAr7HTBb57GfFwh]
> POST /api/extensions/scannerdefinitions HTTP/2
> Host: localhost:8000
> Authorization: Basic YWRtaW46SzNEWGJmZ3NlNVR0UzQ3c01ibG5PZTBqVQ==
> User-Agent: curl/8.4.0
> Accept: */*
> Content-Length: 284614740
> Content-Type: multipart/form-data; boundary=------------------------GlBMII0WAr7HTBb57GfFwh
> 
* We are completely uploaded and fine
< HTTP/2 200 
< vary: Accept-Encoding
< content-type: text/plain; charset=utf-8
< content-length: 57
< date: Wed, 14 Feb 2024 21:19:38 GMT
< 
* Connection #0 to host localhost left intact
Successfully stored the offline vulnerability definitions
```

### DIFF
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -92,7 +92,43 @@ upload_offline_dump() {
     if is_in_PR_context; then
         cmd+=(echo "Would do")
     fi
+    curl --silent --show-error --max-time 60 --retry 3 --create-dirs -o out/RELEASE_VERSION.txt https://raw.githubusercontent.com/stackrox/stackrox/master/scanner/updater/version/RELEASE_VERSION
+    version_file="out/RELEASE_VERSION.txt"
+    # Use grep to extract X.Y versions, sort them, and get the last one as the latest
+    latest_version=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | tail -n 1)
+
+    file_to_check="scanner-v4-defs-${latest_version}.zip"
+    
+    curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check
+
+    if [ -f "$file_to_check" ]; then
+        # If the file exists, add it to scanner-vuln-updates.zip
+        zip scanner-vuln-updates.zip "$file_to_check"
+        echo "$file_to_check added to scanner-vuln-updates.zip"
+    else
+        echo "$file_to_check does not exist."
+    fi
     "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://scanner-support-public/offline/v1/scanner-vuln-updates.zip
+}
+
+upload_v4_versioned_vuln() {
+    info "Uploading offline dump"
+    cmd=()
+    if is_in_PR_context; then
+        cmd+=(echo "Would do")
+    fi
+    cd /tmp/offline-dump
+    version_file="out/RELEASE_VERSION.txt"
+    versions=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V)
+    while IFS= read -r version; do
+        echo "$version"
+        if curl --silent --show-error --max-time 60 --retry 3 -o "scanner-v4-defs-${version}.zip" "https://storage.googleapis.com/scanner-v4-test/offline-bundles/scanner-v4-defs-${version}.zip"; then
+            zip scanner-vulns-${version}.zip scanner-defs.zip k8s-istio.zip scanner-v4-defs-${version}.zip
+            "${cmd[@]}" gsutil cp scanner-vulns-${version}.zip gs://scanner-support-public/offline/v1/${version}/scanner-vulns-${version}.zip
+        else
+            echo "Failed to download scanner-v4-defs-${version}.zip, skipping..."
+        fi
+    done <<< "$versions"
 }
 
 diff_dumps() {
@@ -114,6 +150,8 @@ diff_dumps() {
     # Upload offline dump
     setup_gcp "${SCANNER_GCP_SERVICE_ACCOUNT_CREDS}"
     upload_offline_dump
+
+    upload_v4_versioned_vuln
 }
 
 diff_dumps "$*"
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -98,10 +98,8 @@ upload_offline_dump() {
     latest_version=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | tail -n 1)
 
     file_to_check="scanner-v4-defs-${latest_version}.zip"
-    
-    curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check
 
-    if [ -f "$file_to_check" ]; then
+    if curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check; then
         # If the file exists, add it to scanner-vuln-updates.zip
         zip scanner-vuln-updates.zip "$file_to_check"
         echo "$file_to_check added to scanner-vuln-updates.zip"
@@ -119,9 +117,12 @@ upload_v4_versioned_vuln() {
         cmd+=(echo "Would do")
     fi
     cd /tmp/offline-dump
-    version_file="out/RELEASE_VERSION.txt"
-    versions=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | uniq)
-    while IFS= read -r version; do
+
+    cat out/RELEASE_VERSION.txt |
+        grep -oE '^[0-9]+\.[0-9]+' |
+        sort -V |
+        uniq |
+    while read -r version; do
         echo "$version"
         if curl --silent --show-error --max-time 60 --retry 3 -o "scanner-v4-defs-${version}.zip" "https://storage.googleapis.com/scanner-v4-test/offline-bundles/scanner-v4-defs-${version}.zip"; then
             zip scanner-vulns-${version}.zip scanner-defs.zip k8s-istio.zip scanner-v4-defs-${version}.zip
@@ -129,7 +130,7 @@ upload_v4_versioned_vuln() {
         else
             echo "Failed to download scanner-v4-defs-${version}.zip, skipping..."
         fi
-    done <<< "$versions"
+    done
 }
 
 diff_dumps() {
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -112,14 +112,14 @@ upload_offline_dump() {
 }
 
 upload_v4_versioned_vuln() {
-    info "Uploading offline dump"
+    info "Uploading v4 offline dump"
     cmd=()
     if is_in_PR_context; then
         cmd+=(echo "Would do")
     fi
     cd /tmp/offline-dump
     version_file="out/RELEASE_VERSION.txt"
-    versions=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V)
+    versions=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | uniq)
     while IFS= read -r version; do
         echo "$version"
         if curl --silent --show-error --max-time 60 --retry 3 -o "scanner-v4-defs-${version}.zip" "https://storage.googleapis.com/scanner-v4-test/offline-bundles/scanner-v4-defs-${version}.zip"; then
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -107,6 +107,7 @@ upload_offline_dump() {
         echo "$file_to_check added to scanner-vuln-updates.zip"
     else
         echo "$file_to_check does not exist."
+        exit 1
     fi
     "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://scanner-support-public/offline/v1/scanner-vuln-updates.zip
 }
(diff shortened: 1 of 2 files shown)


scripts/ci/jobs/diff-dumps.sh
```diff
@@ -92,6 +92,22 @@ upload_offline_dump() {
...
+    
+    file_to_check="scanner-v4-defs-${latest_version}.zip"
+
+    if [ -f "$file_to_check" ]; then
+        # If the file exists, add it to scanner-vuln-updates.zip
+        zip scanner-vuln-updates.zip "$file_to_check"
+        echo "$file_to_check added to scanner-vuln-updates.zip"
+    else
+        echo "$file_to_check does not exist."
+    fi
```
jvdm, line 110: Shouldn't this be in `create_offline_dump()`?

daynewlee: I was thinking the same thing but I don't want `create_offline_dump()` to be failed if one of the curl fails.

dcaravel (commented):
Need to be careful here to not 'break' the current latest bundle if this job runs and there is a bug.

I'd expect the output of this job to include updates to all the versioned v4 offline bundles and latest offline bundles, would expect the following to be uploaded:
```
scanner-support-public/offline/v1/scanner-vuln-updates.zip
scanner-support-public/offline/v1/4.4/scanner-vulns-4.4.zip
```
As of this initial release would expect `scanner-vuln-updates.zip` and `4.4/scanner-vulns-4.4.zip` to contain the same data

daynewlee:
> Need to be careful here to not 'break' the current latest bundle if this job runs and there is a bug.
> 
> 
> 
> I'd expect the output of this job to include updates to all the versioned v4 offline bundles and latest offline bundles, would expect the following to be uploaded:
> 
> ```
> 
> scanner-support-public/offline/v1/scanner-vuln-updates.zip
> 
> scanner-support-public/offline/v1/4.4/scanner-vulns-4.4.zip
> 
> ```
> 
> As of this initial release would expect `scanner-vuln-updates.zip` and `4.4/scanner-vulns-4.4.zip` to contain the same data
> 
> 
> 
> 

I haven't decided how to generate scanner-vuln-<version>.zip because I am hesitating to use loop to iterate all scanner-defs-<version>.zip in this script at this moment.

scripts/ci/jobs/diff-dumps.sh
jvdm: I don't fully understand your concern, what will be the impact create_offline_dump() failing if we cannot download and bundle in the scanner-v4 vulns?

daynewlee: if upload_offline_dump fails when downloading downloading scanner-v4-defs, at lease `"${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://sr-roxc/scanner/scanner-vuln-updates.zip` in `create_offline_dump()` can be executed so we won't lost the original scanner-vuln-updates.zip. 

I feel that's slightly better than if there's a failure in `create_offline_dump()`

dcaravel:
> I haven't decided how to generate scanner-vuln-[version].zip because I am hesitating to use loop to iterate all scanner-defs-[version].zip in this script at this moment as you said that might introduce bugs to this workflow. Since scanner-vuln-update.zip only contains latest v4 defs, so it makes sense to change this bash function. I'll make another PR for the v4 offline bundles.

OK, we may be introducing more variance between latest and the versioned bundles by separating these workflows (since the versioned bundles need the Scanner V2 data produced by this job). It is possible that (for example) the `scanner-defs-4.4.zip` file being added `scanner-vuln-updates.zip` will now being different from the one in `4.4/scanner-vulns-4.4.zip`.  One may have newer/older data then the other, but be expected to be the same.

daynewlee:
> > I haven't decided how to generate scanner-vuln-[version].zip because I am hesitating to use loop to iterate all scanner-defs-[version].zip in this script at this moment as you said that might introduce bugs to this workflow. Since scanner-vuln-update.zip only contains latest v4 defs, so it makes sense to change this bash function. I'll make another PR for the v4 offline bundles.
> 
> OK, we may be introducing more variance between latest and the versioned bundles by separating these workflows (since the versioned bundles need the Scanner V2 data produced by this job). If the jobs are separated it will be possible that (for example) `scanner-defs-4.4.zip` added to `scanner-vuln-updates.zip` will be different from `scanner-defs-4.4.zip` in `4.4/scanner-vulns-4.4.zip` depending on when jobs complete. One may have newer/older data then the other, but be expected to be the same.
> 
> FWIW, there are ways to have functions / commands in bash fail without breaking the whole script (if you change your mind).

Yeah, that's true. I will take that into consideration but possibly build on top of this PR. Do you think this PR/change make sense?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -92,9 +92,45 @@ upload_offline_dump() {
...
         cmd+=(echo "Would do")
     fi
+    curl --silent --show-error --max-time 60 --retry 3 --create-dirs -o out/RELEASE_VERSION.txt https://raw.githubusercontent.com/stackrox/stackrox/master/scanner/updater/version/RELEASE_VERSION
+    version_file="out/RELEASE_VERSION.txt"
+    # Use grep to extract X.Y versions, sort them, and get the last one as the latest
+    latest_version=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | tail -n 1)
+
+    curl --silent --show-error --max-time 60 --retry 3 -o scanner-v4-defs-${latest_version}.zip https://storage.googleapis.com/scanner-v4-test/offline-bundles/scanner-v4-defs-${latest_version}.zip
+    
+    file_to_check="scanner-v4-defs-${latest_version}.zip"
```
dcaravel, line 102: nit:
```suggestion
    file_to_check="scanner-v4-defs-${latest_version}.zip"
    
    curl --silent --show-error --max-time 60 --retry 3 -o scanner-v4-defs-${latest_version}.zip https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check
```

dcaravel (approved):
One more nit.

LGTM, approving because believe it will achieve the desired outcome, however because it will modify `scanner-vuln-updates.zip` which is a 'live' file, `HOLD OFF` on the merge until others approve and are aware that the change is being made. 

Also recommend before merging to test uploading this 'combo' file to older supported versions of Central and ensure nothing breaks (I don't expect things to break, but let's be sure - ie: because we're adding another 300MiB it's possible the default timeouts may need to be adjusted, and we need to be extra sure that older Centrals will ignore the extra V4 files in the bundle)

jvdm (commented):
Can you provide your testing steps?

1. Scanner V4 works.
2. Scanner V2 works.

daynewlee:
> Can you provide your testing steps?
> 
>     1. Scanner V4 works.
> 
>     2. Scanner V2 works.

That makes sense. I will add steps for testing new scanner-vuln-update.zip working with Scanner V2 but probably I can't test the bundles working with Scanner V4 since https://github.com/stackrox/stackrox/pull/9409 and [ROX-21981](https://issues.redhat.com/browse/ROX-21981) are not done yet.  @jvdm

RTann:
@daynewlee can I download a sample ZIP to look at?

daynewlee:
> @daynewlee can I download a sample ZIP to look at?

I believe https://storage.googleapis.com/scanner-support-public/offline/v1/scanner-vuln-updates.zip is overwritten by CI now. 
But v4 vulns are available
https://storage.googleapis.com/scanner-support-public/offline/v1/4.2/scanner-vulns-4.2.zip
https://storage.googleapis.com/scanner-support-public/offline/v1/4.3/scanner-vulns-4.3.zip
@RTann

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -92,9 +92,45 @@ upload_offline_dump() {
...
+    
+    curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check
+
+    if [ -f "$file_to_check" ]; then
+        # If the file exists, add it to scanner-vuln-updates.zip
+        zip scanner-vuln-updates.zip "$file_to_check"
+        echo "$file_to_check added to scanner-vuln-updates.zip"
+    else
+        echo "$file_to_check does not exist."
```
RTann, line 109: We should make sure we are aware when this happens. Can we fail the script and notify the team?

daynewlee: Is this what you mean?
https://github.com/stackrox/scanner/pull/1400/commits/2b3c1c6d674dc2123b912fb1db66a2da7e6caa76

RTann: I meant more like when we are unable to add the v4 vulns into this zip. This is an error scenario, as we shoould always have the vulns available, right?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -92,9 +92,46 @@ upload_offline_dump() {
...
+    version_file="out/RELEASE_VERSION.txt"
+    # Use grep to extract X.Y versions, sort them, and get the last one as the latest
+    latest_version=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | tail -n 1)
+
+    file_to_check="scanner-v4-defs-${latest_version}.zip"
+    
+    curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check
+
+    if [ -f "$file_to_check" ]; then
+        # If the file exists, add it to scanner-vuln-updates.zip
```
jvdm, line 105: What is the scenario when the file does not exist? `curl` will fail if `offline-bundle/$file_to_check` returns a 404 NotFound; Since `set -e` the script we should never reach this line, right? Can you clarify?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -92,9 +92,46 @@ upload_offline_dump() {
...
+upload_v4_versioned_vuln() {
+    info "Uploading v4 offline dump"
+    cmd=()
+    if is_in_PR_context; then
+        cmd+=(echo "Would do")
+    fi
+    cd /tmp/offline-dump
+    version_file="out/RELEASE_VERSION.txt"
+    versions=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | uniq)
+    while IFS= read -r version; do
```
jvdm, line 124:
```suggestion
    cat out/RELEASE_VERSION.txt |
        grep -oE '^[0-9]+\.[0-9]+' | 
        sort -V |
        uniq |
        while read -r version; do
```

jvdm: Maybe easier to read?

scripts/ci/jobs/diff-dumps.sh
daynewlee: You are right. I should do
`If curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://storage.googleapis.com/scanner-v4-test/offline-bundles/$file_to_check`

c-du (approved):

PR #1394, opened 2024-02-01
**DO NOT MERGE**: try taskRunSpecs

msugakov:
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1706117415639669

PR #1401, opened 2024-02-08
chore: Override `cpu` `requests` for bad days in Konflux

msugakov:
The overrides are possible after https://issues.redhat.com/browse/PLNSRVCE-1625 got addressed.

I did not go over OOTB limits, rather made requests high enough so that CPU is actually reserved for intensive steps when Konflux cluster is oversubscribed. However, we can redefine limits too if needed.

I did not override cpu for scanner-db* builds because there's no lengthy Go compilation. I hope the default 250m should be enough, or we'll see.

My idea for this PR is to have the feature enabled to assess its stability and usefulness for us should we observe the need to tweak requests.

This reinstates <https://github.com/stackrox/scanner/pull/1394>.

The commit history (at the time of publishing) is messy. It only makes sense to review the final state (i.e. not per commit).

### DIFF
--- a/.tekton/scanner-db-pull-request.yaml
+++ b/.tekton/scanner-db-pull-request.yaml
@@ -63,14 +63,6 @@ spec:
         secretName: '{{ git_auth_secret }}'
 
   taskRunSpecs:
-    - pipelineTaskName: build-container
-      stepSpecs:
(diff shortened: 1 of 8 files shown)


tommartensen (approved):
LGTM.
That will hopefully ensure that scanner builds are faster, because of the optimized scheduling.

PR #1416, opened 2024-02-24
ROX-22044: postgresql 15

RTann:
See for related changes: https://github.com/stackrox/rox-ci-image/pull/198

DB integration tests run in a GitHub Workflow, so it should be using PostgreSQL 15. If not, I also tested it locally by running `make db-integration-tests` while running a PostgreSQL 15 server in the background.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ jobs:
   pre-build-updater:
     runs-on: ubuntu-latest
     container:
-      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.61
+      image: quay.io/stackrox-io/apollo-ci:scanner-test-0.3.65
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


BradLugo:
I wouldn't normally comment on this, but since we're changing it in a few places, we don't generally favor `printf` over `echo` AFAIK. As for the readability (or even perhaps the functionality) of string interpolation, we usually use bash's parameter expansion syntax, i.e., `${MYVAR}`

RTann:
> I wouldn't normally comment on this, but since we're changing it in a few places, we don't generally favor `printf` over `echo` AFAIK. As for the readability (or even perhaps the functionality) of string interpolation, we usually use bash's parameter expansion syntax, i.e., `${MYVAR}`. That being said, I wouldn't block a PR on this nit

I took these changes directly from https://github.com/docker-library/postgres/blob/44ef8b226a40f86cf9df3f9299067db6779a3aa3/15/bookworm/docker-entrypoint.sh and I think it'd be best to stay as close to the original source as possible, so I'm going to keep it as-is

BradLugo (approved):
Barring the image tags, LGTM!

PR #1446, opened 2024-03-15
chore: add expiry label to images

RTann:
See https://github.com/stackrox/stackrox/pull/10385 for Scanner V4


dcaravel (commented):
Labels appear empty
```
docker inspect quay.io/stackrox-io/scanner-db-slim:2.33.x-13-gd0b3f1e261 | jq '.[].Config.Labels' | grep -E "version|release|quay"
  "io.buildah.version": "1.29.0",
  "quay.expires-after": "",
  "release": "",
  "version": ""
```

RTann:
CI failure is unrelated

dcaravel (approved):

Makefile
```diff
@@ -95,6 +103,9 @@ build-updater: deps
 	@echo "+ $@"
 	go build -trimpath -o ./bin/updater ./cmd/updater
 
+.PHONY: image-build-args
+image-build-args:
+	@echo $(IMAGE_BUILD_ARGS)
```
porridge, line 108: Rather than expose this here, _and_ duplicate the `docker build` command in the GHA yaml file, why not just call `make scanner-image...` in the GH action in the first place?
cc @BradLugo

porridge (approved):
This look OK, but I wonder why we're duplicating the `docker build` command in the workflow definition? 🤔

Makefile
RTann: I'm going to merge as-is, as this repo is more-or-less in maintenance mode, so I think this is ok

PR #1452, opened 2024-03-21
chore(go): go1.21.8

RTann:
Build images with Go 1.21.8. Note: we don't bother bumping the `go.mod` version, as there is nothing in 1.21 which we require


dcaravel (approved):
LGTM

PR #1454, opened 2024-03-22
chore(deps): bump stackrox/rox dep

RTann:
Bump this version in preparation for the future removal of `pkg/bolthelper`. `stackrox/rox` and `stackrox/scanner` unfortunately have a circular dependency, so I think it's best to try to catch up with the latest version of `stackrox/rox` before `stackrox/rox` catches up with `stackrox/scanner`.

This PR also bumps the minimum required go version to 1.19, as `stackrox/rox` requires this

PR #1456, opened 2024-03-26
chore(go): bump go.mod to go1.21

RTann:
This is in preparation of https://github.com/stackrox/scanner/pull/1454 which uses go1.21


cmd/clair/main.go
```diff
@@ -107,8 +106,6 @@ func waitForSignals(signals ...os.Signal) {
 
 // Boot starts a Clair instance with the provided config.
 func Boot(config *Config, slimMode bool) {
-	rand.Seed(time.Now().UnixNano())
```
RTann, line 110: This was deprecated in go1.20, and we get a build error using this with go1.21.

daynewlee (approved):

PR #1464, opened 2024-04-04
ROX-19862: Add manual entry for CVE-2023-32697

dcaravel:
CVE-2023-32697 was not being detected for `sqlite-jdbc-3.41.2.1.jar` due to an inaccurate CPE URI on the [NVD entry](https://nvd.nist.gov/vuln/detail/CVE-2023-32697).

To fix the match a manual entry was created with `cpe:2.3:a:xerial:sqlite-jdbc:*:*:*:*:*:*:*:*` added.

You'll notice some information appears missing when comparing the `updater` output to the [NVD page](https://nvd.nist.gov/vuln/detail/CVE-2023-32697) and API, such as  references and CWE.  This appears to be because the updater omits data that was not provided by NVD, such as other CNA's, in this case GitHub.  Consider this while reviewing the PR - if the additional data is used by Scanner / ACS it may be worthwhile adding it, unless there is a reason not to (such as licensing).

<details>
  <summary>JSON output from Scanner updater</summary>

Output from: `./bin/updater generate-dump...`
  ```json
    {
      "cve": {
        "affects": null,
        "CVE_data_meta": {
          "ASSIGNER": "",
          "ID": "CVE-2023-32697"
        },
        "data_format": "",
        "data_type": "",
        "data_version": "",
        "description": {
          "description_data": [
            {
              "lang": "en",
              "value": "SQLite JDBC is a library for accessing and creating SQLite database files in Java. Sqlite-jdbc addresses a remote code execution vulnerability via JDBC URL. This issue impacting versions 3.6.14.1 through 3.41.2.1 and has been fixed in version 3.41.2.2.\n"
            }
          ]
        },
        "problemtype": null,
        "references": null
      },
      "configurations": {
        "CVE_data_version": "",
        "nodes": [
          {
            "cpe_match": [
              {
                "cpe23Uri": "cpe:2.3:a:sqlite_jdbc_project:sqlite_jdbc:*:*:*:*:*:*:*:*",
                "versionEndExcluding": "3.41.2.2",
                "versionStartIncluding": "3.6.14.1",
                "vulnerable": true
              }
            ],
            "operator": "OR"
          }
        ]
      },
      "impact": {
        "baseMetricV3": {
          "cvssV3": {
            "attackComplexity": "LOW",
            "attackVector": "NETWORK",
            "availabilityImpact": "HIGH",
            "baseScore": 9.8,
            "baseSeverity": "CRITICAL",
            "confidentialityImpact": "HIGH",
[...]

PR #1467, opened 2024-04-05
chore: update get_pr_details with the stackrox version

BradLugo:
## Description

We have a bash function that is most, if not only, used in CI to get the PR details. This function wasn't updated when we moved the OSCI jobs to GitHub Actions, which is causing PRs not to get the labels properly. The newer version found in [stackrox/stackrox](https://github.com/stackrox/stackrox) also has better logging.

These changes update `get_pr_details()` with the current version in [stackrox/stackrox](https://github.com/stackrox/stackrox).

### DIFF
--- a/scripts/ci/lib.sh
+++ b/scripts/ci/lib.sh
@@ -337,6 +337,7 @@ pr_has_label() {
 
 # get_pr_details() from GitHub and display the result. Exits 1 if not run in CI in a PR context.
 _PR_DETAILS=""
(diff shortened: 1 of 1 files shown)


BradLugo:
It'd be cool if we could test these changes in a PR that's giving us trouble, e.g., https://github.com/stackrox/scanner/pull/1464

cc @dcaravel

BradLugo:
This will be merged in as part of https://github.com/stackrox/scanner/pull/1464.

PR #1472, opened 2024-04-10
chore(deps): remove github.com/mholt/archiver/v3 dependency

RTann:
https://github.com/mholt/archiver/pull/396 has yet to land so github.com/mholt/archiver/v3 is still affected by CVE-2024-0406. This repository is *not* affected by this vulnerability.

This PR removes the dependency to:

* minimize dependencies
* ensure security scanners do not claim Scanner is affected, when it is not

Implementation is completely based on https://github.com/mholt/archiver/blob/v3.5.1/zip.go#L140.

Note: I want to make sure the diff-dumps still give the same data, but that step is failing at the moment... In any case, it doesn't hurt to start the review process

### DIFF
--- a/pkg/vulndump/write.go
+++ b/pkg/vulndump/write.go
@@ -1,20 +1,24 @@
 package vulndump
 
 import (
+	"archive/zip"
 	"compress/flate"
 	"encoding/json"
+	"io"
+	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 
-	"github.com/mholt/archiver/v3"
 	"github.com/pkg/errors"
+	"github.com/stackrox/rox/pkg/utils"
 	"github.com/stackrox/scanner/database"
+	"github.com/stackrox/scanner/pkg/fsutil"
 )
 
 // WriteZip takes the given files and creates the vuln dump zip.
 func WriteZip(inputDir, outFile string, ignoreKubernetesVulns, ignoreRHELv2Vulns, ignoreIstioVulns bool) error {
-	zipArchive := archiver.NewZip()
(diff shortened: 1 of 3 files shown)


msugakov:
This PR and subsequent scanner dependency update in stackrox/stackrox should help with https://github.com/stackrox/stackrox/security/dependabot/270

pkg/vulndump/write.go
```diff
@@ -58,3 +62,166 @@ func WriteOSVulns(outputDir string, vulns []database.Vulnerability) error {
...
+		if err != nil {
+			return err
+		}
+		finfo := fileInfo{
+			FileInfo:   info,
+			customName: nameInArchive,
+		}
+
+		var file io.ReadCloser
+		if finfo.Mode().IsRegular() {
```
dcaravel, line 169: In the original source this was `info.Mode().IsRegular()`, curious why the change to `finfo`?

dcaravel (commented):
Looks like code was removed related to handling symlinks, was that because in the various workflows that use `WriteZip` we know symlinks are not being used?

(rest looks OK to me - holding approve for a successful diff-dump / test)

RTann:
> Looks like code was removed related to handling symlinks, was that because in the various workflows that use `WriteZip` we know symlinks are not being used?

Yes, `WriteZip` is only used by `generate-dump` and `diff-dumps`, and we do not use symlinks for either (that I know of).

> (rest looks OK to me - holding approve for a successful diff-dump / test)

I'm also waiting for `diff-dump`. Running into space issues in CI, though it seems unique to this PR, so I wonder if I messed something up, I'm getting unlucky, or GHA allocates different resources for PR vs push to master branch

PR #1476, opened 2024-04-15
chore(deps): migrate github.com/heroku/docker-registry-client => gith…

RTann:
…ub.com/stackrox/docker-registry-client


RTann:
huge pain not worth the effort

PR #1489, opened 2024-04-25
ROX-22889: Cherry pick new genesis dump into 2.32 (targeting ACS 4.3.7)

dcaravel:


image/scanner/dump/genesis_manifests.json
```diff
@@ -446,6 +446,16 @@
       "dumpLocationInGS": "gs://stackrox-scanner-ci-vuln-dump/genesis-20231025151131.zip",
       "timestamp": "2023-10-25T15:11:31.926496239Z",
       "uuid": "93AEC554-29EE-4E24-96D6-744092A98444"
+    },
+    {
+      "dumpLocationInGS": "gs://stackrox-scanner-ci-vuln-dump/genesis-20240228002053.zip",
+      "timestamp": "2024-02-28T00:20:53.211664833Z",
+      "uuid": "c73d1bf5-49cb-48cb-b475-46702de2be73"
+    },
```
jvdm, line 454: Curious: Was this missing by chance?

jvdm (approved):

image/scanner/dump/genesis_manifests.json
dcaravel: I wouldn't say missing, believe that was from the ACS 4.4.0 release - As far as I can tell we do not typically update older releases with new genesis files (perhaps we should given we now know it can address corrected records).  I figured add both records in case we need to do this again so we don't have merge conflicts.

dcaravel:
cherry picked a handful of commits that update tests cases to get CI to pass

RTann (approved):

PR #1497, opened 2024-05-06
ROX-23123: Verifying File Integrity while updating offline bundle 

daynewlee:


daynewlee:
Not sure if this is the most graceful way of doing things. Please let me know your suggestions.

jvdm (commented):
How did you test this?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -121,13 +121,19 @@ upload_offline_dump() {
     latest_version=$(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -V | tail -n 1)
 
     file_to_check="scanner-v4-defs-${latest_version}.zip"
+    temp_file="/tmp/$file_to_check"
 
-    if curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://definitions.stackrox.io/v4/offline-bundles/$file_to_check; then
-        # If the file exists, add it to scanner-vuln-updates.zip
-        zip scanner-vuln-updates.zip "$file_to_check"
-        echo "$file_to_check added to scanner-vuln-updates.zip"
+    if curl --silent --show-error --max-time 60 --retry 3 -o "$temp_file" "https://definitions.stackrox.io/v4/offline-bundles/$file_to_check"; then
```
jvdm, line 126: We have `-o e` enabled; why not just let `curl` fail on its own instead of wrapping these calls in `if-else`?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -121,13 +121,19 @@ upload_offline_dump() {
...
+        if file "$temp_file" | grep -q 'Zip archive data'; then
```
jvdm, line 127: Do you know why `curl` was successful, but the output file was invalid/incorrect/bogus?

daynewlee: I am guessing probably `curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check` is saving error messages as `$file_to_check`

maybe `if curl --silent --show-error --max-time 60 --retry 3 -o "$temp_file" "https://definitions.stackrox.io/v4/offline-bundles/$file_to_check"` doesn't check status code?

jvdm: Yeah, your guesses and maybes are possible, you probably want to look at `man curl`.

daynewlee: I decided to use `jq .` and it should return parse error if there's no json file exists.

scripts/ci/jobs/diff-dumps.sh
daynewlee: I added `jq .` to the if check so it should be ok now.

scripts/ci/jobs/diff-dumps.sh
jvdm: You're missing `--fail`:

```
       -f, --fail
              (HTTP) Fail fast with no output at all on server errors. This  is
              useful to enable scripts and users to better deal with failed at‐
              tempts.  In  normal  cases when an HTTP server fails to deliver a
              document, it returns an HTML document  stating  so  (which  often
              also  describes  why and more). This command line option prevents
              curl from outputting that and return error 22.
```

With that and `set -e`, non-successful HTTP requests will make `curl` return non-zero, and the shell will exit.

I still think it's worth doing the `jq` sanity check to ensure it's a valid JSON. It's a good sanity check. We should avoid ingesting invalid input into the pipeline.

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -122,12 +122,12 @@ upload_offline_dump() {
 
     file_to_check="scanner-v4-defs-${latest_version}.zip"
 
-    if curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://definitions.stackrox.io/v4/offline-bundles/$file_to_check; then
+    if curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://definitions.stackrox.io/v4/offline-bundles/$file_to_check && jq . $file_to_check; then
```
jvdm, line 125: Use `--fail` and you don't need the `if-else`.

scripts/ci/jobs/diff-dumps.sh
jvdm: Short-sight on my side: The file is a zip archive, so using `jq` for sanity check doesn't make sense.

Maybe we can use `unzip -l`. That will show us the zip file's content.

scripts/ci/jobs/diff-dumps.sh
jvdm: There is also `--fail-with-body`.

daynewlee: but  `--fail` doesn't help to exit the process with status 1

daynewlee: Never mind, I saw this command: `set -euo pipefail`

scripts/ci/jobs/diff-dumps.sh
daynewlee: Yeah, I was confused that we should check the json file.

daynewlee:
> How did you test this?

I can only run the diff-dump job to see if the offline bundle is generated. But I think since there's `--fail `flag and  `set -euo pipefail` in the script we should be fine. Let me know what you think. @jvdm

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -122,14 +122,10 @@ upload_offline_dump() {
...
-    if curl --silent --show-error --max-time 60 --retry 3 -o $file_to_check https://definitions.stackrox.io/v4/offline-bundles/$file_to_check; then
-        # If the file exists, add it to scanner-vuln-updates.zip
-        zip scanner-vuln-updates.zip "$file_to_check"
-        echo "$file_to_check added to scanner-vuln-updates.zip"
-    else
-        echo "$file_to_check does not exist."
-        exit 1
-    fi
+    curl --silent --show-error --fail --max-time 60 --retry 3 -o $file_to_check https://definitions.stackrox.io/v4/offline-bundles/$file_to_check
+    zip scanner-vuln-updates.zip "$file_to_check"
```
jvdm, line 126:
```suggestion
    unzip -l "$file_to_check"
    zip scanner-vuln-updates.zip "$file_to_check"
```

Do you think it is worth it?

jvdm (approved):

PR #1501, opened 2024-05-09
Fix RHSA-2022:7288 test case

daynewlee:


RTann (approved):

RTann:
going to just merge this now so scanner CI will be happier asap. Thanks for doing this!

PR #1502, opened 2024-05-13
ROX-23132: use make tag result in image names on Konflux

tommartensen:
According to https://github.com/stackrox/stackrox/pull/10749.


tommartensen:
Closing because I forgot to add `konflux` in the branch name to trigger Konflux CI :(

PR #1523, opened 2024-05-21
ROX-23836 vulnerabilities container

BradLugo:
WIP


dcaravel (commented):
LGTM, `quay.io/stackrox-io` does not have any active tags (they look to have been deleted) - is that a concern? expected?

BradLugo:
> `quay.io/stackrox-io/vulnerabilities` does not have any active tags (they look to have been deleted) - is that a concern? expected?

Expected. They were on a 7-day expiration so they all expired at this point.

image/vulnerabilities/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ARG LABEL_VERSION
+ARG LABEL_RELEASE
+ARG QUAY_TAG_EXPIRATION
+
+LABEL name="definitions" \
+      vendor="StackRox" \
+      maintainer="https://stackrox.io/" \
+      summary="Vulnerability definitions for the StackRox Security Platform" \
+      description="This image contains vulnerability definitions used by StackRox Scanner V4." \
```
RTann, line 15: you can just say `StackRox scanners` since it's for both

image/vulnerabilities/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+      version="${LABEL_VERSION}" \
+      release="${LABEL_RELEASE}" \
+      quay.expires-after="${QUAY_TAG_EXPIRATION}"
+
+COPY scanner-vuln-updates.zip /app/scanner-vuln-updates.zip
```
RTann, line 20: nit: Should we just have it in the root directory? Not sure if `app` makes sense. Or perhaps there's some Linux standard for putting random data files? Not sure if anyone else has opinions on this

.github/workflows/ci.yaml
```diff
@@ -489,6 +489,84 @@ jobs:
...
+          
+          architectures=("amd64" "arm64" "ppc64le" "s390x")
+          image_list=()
+          for arch in "${architectures[@]}"; do
+            arch_image_tag="${BASE_IMAGE_TAG}-${arch}"
+          
+            docker buildx build \
+              --platform "linux/${arch}" \
+              --load \
+              --tag "${arch_image_tag}" \
```
RTann, line 557: I wonder if we should always just do latest since we only publish one version of the offline data

image/vulnerabilities/Dockerfile
jvdm: > Or perhaps there's some Linux standard for putting random data files?

`/srv/` is the closest to what we are doing here. Maybe `/opt/` would also make sense. Although, in containers, FHS loses its relevance a bit so any directory the images' base OS won't bother is fine.

.github/workflows/ci.yaml
BradLugo: I changed `BASE_IMAGE_TAG` but would still like to have some record of the tag for troubleshooting purposes, so I left `LABEL_VERSION` and `LABEL_RELEASE` the same. Let me know if there's a better way to handle that use case.

image/vulnerabilities/Dockerfile
BradLugo: Good catch. That was from when I thought this was only related to Scanner V4 😅

image/vulnerabilities/Dockerfile
BradLugo: Cool. Yeah, I don't have a strong opinion

dcaravel (approved):

PR #1526, opened 2024-05-23
ROX-20232: Unify PR and push `PipelineRun`-s

msugakov:
After seeing very little difference between `<blah>-push.yaml` and `<blah>-pull-request.yaml`, I went ahead and merged the two.

`build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'` remains literally that on pushes as opposed to PRs where `{{pull_request_number}}` gets replaced with a number (like `1526` here). I think, it should be fine since it's not something that gets in a final image.

The only thing I regret is that the [activity page](https://console.redhat.com/preview/application-pipeline/workspaces/rh-acs/applications/acs/activity/pipelineruns) would not let easily distinguish pushes from pull requests.
image


tommartensen (approved):
This change LGTM.

What about the tags though? Should we do implement it the same way as in the collector repository? 

In this PR or another one, both is fine for me.

msugakov:
> What about the tags though?

I suppose you mean https://issues.redhat.com/browse/ROX-24116 . I want to do it in a separate PR because the approach may be slightly different. I'll try to get by without additional `apply-tags` task for Scanner.

PR #1527, opened 2024-05-27
ROX-24116: Fix tags for `init` and other cleanups

msugakov:
TODO


image/scanner/rhel/konflux.Dockerfile
```diff
@@ -18,8 +20,8 @@ RUN unzip -j blob-repo2cpe.zip -d image/scanner/dump/repo2cpe && \
     unzip -j blob-k8s-definitions.zip -d image/scanner/dump/k8s_definitions && \
     unzip -j blob-nvd-definitions.zip -d image/scanner/dump/nvd_definitions
 
-RUN echo -n "version: " && scripts/konflux/version.sh && \
-    go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(scripts/konflux/version.sh)" -o image/scanner/bin/scanner ./cmd/clair
+RUN echo -n "version: " && make --quiet --no-print-directory tag && \
+    make CGO_ENABLED=1 scanner-build-nodeps
```
msugakov, line 24: There will be now `2.33.x-80-gbef2d6709f-fast` in `github.com/stackrox/scanner/pkg/version.Version` and that one is used in multiple places. I'm not sure what's going to break but at least let's break it early enough and see.

tommartensen (approved):

Makefile
```diff
@@ -21,7 +21,7 @@ endif
 endif
 
 ifeq ($(TAG),)
-TAG=$(shell git describe --tags --abbrev=10 --dirty --long)
+TAG=$(shell git describe --tags --abbrev=10 --dirty --long)$(TAG_SUFFIX)
```
RTann, line 24: assuming this is only populated via Konflux pipelines, right?

RTann (approved):
Makefile change LGTM with just one verification question

PR #1531, opened 2024-05-31
Bump containers/image/v5 targeting ACS 4.4.3

dcaravel:


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/PuerkitoBio/goquery v1.9.0
 	github.com/ckaznocha/protoc-gen-lint v0.3.0
-	github.com/containers/image/v5 v5.29.2
+	github.com/containers/image/v5 v5.30.1
```
RTann, line 10: I'm thinking we should just do v5.29.3 because it's less disruptive. Doesn't matter as much here if it works and we don't plan on bumping the scanner Go dependency in the stackrox repo, but it protects us for the (off) chance we would need to bump the scanner dep in the stackrox repo for 4.3 in the future

dcaravel: The advisory states anything less then <5.30.1 is affected, based on that I’m assuming we would still detect [CVE-2024-3727](https://github.com/advisories/GHSA-6wvf-f2vw-3425), even though its ‘technically’ fixed with the bump to 5.29.3 that has the backported fix

dcaravel: https://github.com/stackrox/scanner/pull/1531/commits/89683b25b0f7f49cb217a7525fe3d9c6ec364655

dcaravel: done

RTann (approved):
LGTM though the E2E tests are still not pleased :(

PR #1532, opened 2024-05-31
chore: Swap `TAG_SUFFIX` for `SCANNER_TAG`

msugakov:
When looking at collector (<https://github.com/stackrox/collector/pull/1694>), I realized I did a bit of mess in my earlier <https://github.com/stackrox/scanner/pull/1527>.

Here I get rid of `TAG_SUFFIX` variable and inject the full tag via `RELEASE_TAG` environment variable. See
https://github.com/stackrox/scanner/blob/631dea9898c3c69f18ea3c45dd66e733cd616905/Makefile#L13-L21

I also inject a value into the `version` label in all four built containers. Similar to <https://github.com/stackrox/collector/pull/1694>, if/when we'll need to re-tag, we will take care of the label then. Mentioned that in ROX-24468.

Testing:
- [x] 4 Konflux-built images still have `-fast` in the tag.
- [x] 4 Konflux-built images have `-fast` in `version` label.
- [x] GHA-built images don't have `-fast` in tag or `label`.
- [x] Konflux-built scanner and slim-scanner have `-fast` version in `github.com/stackrox/scanner/pkg/version.Version`.

For the last item, smoke-tested build, seeing proper thing in logs:
```
{"Event":"Running Scanner version: 2.33.x-80-gc60f8fc954-fast","Level":"info","Location":"main.go:293","Time":"2024-06-05 17:39:50.747575"}
```
and
```
{"Event":"Running Scanner version 2.33.x-80-gc60f8fc954-fast in Node Inventory mode","Level":"info","Location":"main.go:280","Time":"2024-06-05 17:43:45.267545"}
```
Both image and node scanning seem to work.

### DIFF
--- a/.tekton/scanner-component-pipeline.yaml
+++ b/.tekton/scanner-component-pipeline.yaml
@@ -252,7 +252,7 @@ spec:
       value: $(params.build-target-stage)
     - name: BUILD_ARGS
       value:
-      - TAG_SUFFIX=$(params.output-tag-suffix)
+      - SCANNER_TAG=$(tasks.determine-image-tag.results.image-tag)
     runAfter:
     - prefetch-dependencies
     - fetch-scanner-data
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 endif
 
 ifeq ($(TAG),)
-TAG=$(shell git describe --tags --abbrev=10 --dirty --long)$(TAG_SUFFIX)
+TAG=$(shell git describe --tags --abbrev=10 --dirty --long)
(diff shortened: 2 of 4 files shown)


RTann (approved):
Makefile change LGTM. I'll leave the rest to someone with more Konflux expertise, though

kylape (approved):
My understanding is that this PR prevents the call to `make tag` from happening in the buildah task in favor of the one already happening in the `determine-image-tag` task, which is the desired behavior and is consistent with what we are doing in the other repos.

PR #1536, opened 2024-06-04
fix(ROX-23707): scanner-db-slim must have an initdb entrypoint directory

tommartensen:


image/db/rhel/konflux.Dockerfile
```diff
@@ -33,7 +33,8 @@ RUN dnf upgrade -y --nobest && \
     dnf clean all && \
     rpm --verbose -e --nodeps $(rpm -qa curl '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
     rm -rf /var/cache/dnf /var/cache/yum && \
-    chmod +x /usr/local/bin/docker-entrypoint.sh
+    chmod +x /usr/local/bin/docker-entrypoint.sh && \
+    mkdir /docker-entrypoint-initdb.d
```
msugakov, line 37:
```suggestion
    mkdir -p /docker-entrypoint-initdb.d
```
Up to you since the dir should not exist anyway.

msugakov (approved):
Good find! My comment is optional.

PR #1542, opened 2024-06-07
ROX-20855: Reduce container user permissions over `docker-entrypoint.sh`

msugakov:
There's no need to own executable by `postgres` user, hence removed `chown`.

There's no need to add executable flag because `COPY` statement preserves it and the original file has it, hence removed `chmod`.


RTann (approved):
LGTM and normal CI tests pass, but I'm not familiar with the konflux failures

msugakov:
Konflux enterprise-contract failures are noise at the moment, it's ok to ignore them.

PR #1544, opened 2024-06-13
fix(deps): switch to github.com/anchore/archiver/v3

RTann:
github.com/mholt/archiver/v3 v3.51.1 is affected by CVE-2024-0406. There is a [PR in-progress](https://github.com/mholt/archiver/pull/396) to resolve this, but it has been rather inactive lately.

We are not affected by this vulnerability, as it affects a codepath which we do not use.

I previously decided to remove the https://github.com/stackrox/scanner/pull/1472 dependency by copying functions over to this repository. However, I ran into issues with CI I have not really been wanting to resolve in that PR, and I believe this approach is safer.

In this approach, we replace github.com/mholt/archiver/v3 v3.51.1 with github.com/anchore/archiver/v3 v3.51.2. See https://github.com/mholt/archiver/compare/v3.5.1...anchore:archiver:v3.5.2 for the exact details. It is clear the only difference is the contents of the previously mentioned PR, which does not affect our codepath.

Note 1: This uses the `replace` directive instead of actually completely replacing the dependency due to annoying circular dependency things with the `stackrox/stackrox` repo.

Note 2: This will not resolve the vulnerability match in the `stackrox/stackrox` repo, as `replace` is only used when the `go.mod` in this repo is the main module. So, we will have to `replace` in the `stackrox/stackrox` repo, too

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -188,6 +188,19 @@ replace (
 
 	github.com/heroku/docker-registry-client => github.com/stackrox/docker-registry-client v0.1.0
 
+	// The current latest version of github.com/mholt/archiver/v3 (v3.5.1) suffers from CVE-2024-0406.
+	// There is currently a PR in place to resolve it (https://github.com/mholt/archiver/pull/396),
+	// but it has not had much attention recently.
+	// Just replace our usage of github.com/mholt/archiver/v3 with github.com/anchore/archiver/v3 (v3.5.2)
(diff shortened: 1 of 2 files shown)


dcaravel (commented):
Added label `generate-dumps-on-pr` to try and generate some dumps to ensure works.  May need an empty commit to trigger?

dcaravel (approved):

PR #1568, opened 2024-07-12
fix(ci): fallback to previous release when attempting unreleased version

RTann:
When RELEASE_VERSIONS.txt is updated to a new unreleased minor version, the `diff-dumps` job fails to upload the latest offline vulnerability data. That is because (as mentioned, the latest minor version does not exist, so we we attempt to pull its related data, there is nothing there.

This PR now will check if the latest version is unreleased. If that's the case, then it falls back to the latest released version, so offline vuln updates may continue.

We determine if there is a release by checking for the existence of the tag in the stackrox/stackrox repo. If it returns a 200, it exists. If it returns a 404, then it doesn't exist. Any other status code means there is some other kind of error, so we should fail

### DIFF
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -111,22 +111,43 @@ upload_offline_dump() {
     info "Uploading offline dump"
 
     cd /tmp/offline-dump
-    cmd=()
-    if is_in_PR_context; then
-        cmd+=(echo "Would do")
-    fi
+
     curl --silent --show-error --max-time 60 --retry 3 --create-dirs -o out/RELEASE_VERSION.txt https://raw.githubusercontent.com/stackrox/stackrox/master/scanner/updater/version/RELEASE_VERSION
     version_file="out/RELEASE_VERSION.txt"
-    # Use grep to extract X.Y versions, sort them, and get the last one as the latest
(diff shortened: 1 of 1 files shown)


scripts/ci/jobs/diff-dumps.sh
```diff
@@ -111,22 +111,43 @@ upload_offline_dump() {
...
+        elif [[ "$status_code" == "200" ]]; then
+            # Release was cut, so use this one.
+            latest_version="$version"
+            break
+        else
+            # Some other error occurred.
+            >&2 echo "received status code $status_code when fetching StackRox releases"
+            exit 1
+        fi
+    done < <(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort -ruV)
```
RTann, line 134: Note I updated the flags here to `-ruV` instead of just `-V`. This way we can try the latest version first (`-r`). Also no need to try an X.Y release more than once, hence the `-u`

BradLugo: I find the long-hand name for flags more helpful when reading bash; do you mind changing them here?

BradLugo (approved):
One minor suggestion. Everything else LGTM!

scripts/ci/jobs/diff-dumps.sh
RTann: yep, done [276e3a8](https://github.com/stackrox/scanner/pull/1568/commits/276e3a8ecd39e51216d374a9cee73b1a65dbe20a)

PR #1570, opened 2024-07-17
fix: add tag resolution for jobs missing it

dcaravel:
The `upload-dumps-for-downstream` job did not upload bundles with the appropriate tag - breaking downstream builds.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -541,6 +541,8 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
 
+      - uses: ./.github/actions/handle-tagged-build
+
       - uses: actions/download-artifact@v4
         with:
           name: offline-dump
@@ -638,6 +640,8 @@ jobs:
           fetch-depth: 0
 
       - uses: ./.github/actions/job-preamble
+
+      - uses: ./.github/actions/handle-tagged-build
 
       - uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


RTann (approved):
Would we now need to cut a new release?

dcaravel:
> Would we now need to cut a new release?

That's one option, OR, rename the files previously uploaded to use the proper tag (the latter being my preference)

RTann:
> > Would we now need to cut a new release?
> 
> That's one option, OR, rename the files previously uploaded to use the proper tag (the latter being my preference)

As long as you are sure the previous files have the right data, then that's fine with me

RTann (approved):

dcaravel:
> As long as you are sure the previous files have the right data, then that's fine with me

Would be the same dump of data using in the tagged build: https://github.com/stackrox/scanner/actions/runs/9845637787

Taking a step back - makes me question the `genesis-dump` usage in `.github/workflows/ci.yaml`.

When we create a new scanner tag, we would have already created and added a genesis-dump reference to `genesis_manifests.json` - However, the CI workflow will create a new genesis-dump, and not re-use the dump referenced in `genesis_manifests.json`. 

 If that is accurate, that means the vulnerabilities embedded within an tagged image will be 'newer' then those referenced in `genesis_manifests` - which may be OK - the impact may just be the `diff-dumps` will include vulns that have already been updated.

.github/workflows/ci.yaml
```diff
@@ -541,6 +541,8 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
 
+      - uses: ./.github/actions/handle-tagged-build
+
```
BradLugo, line 545: I don't think we need this here. This job only runs on pushes to the `master` branch.

PR #1573, opened 2024-07-19
ROX-25321: convert konflux builds to OCI artifacts

Stringy:


.tekton/scanner-component-pipeline.yaml
```diff
@@ -298,14 +275,16 @@ spec:
       value: $(tasks.build-container.results.IMAGE_URL)
     - name: BASE_IMAGES
       value: $(tasks.build-container.results.BASE_IMAGES_DIGESTS)
-    runAfter:
-    - build-container
+    - name: SOURCE_ARTIFACT
+      value: $(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)
```
tommartensen, line 279: Could/should this be `tasks.fetch-scanner-data.results.SOURCE_ARTIFACT` as the last task that edited the source?
(it won't matter for what this task attempts to achieve, nor for the order of tasks, just for the next person to copy&paste the params from this task to a latter task)

tommartensen (approved):
one tiny question, but not blocking

.tekton/scanner-component-pipeline.yaml
Stringy: I've also switched `sast-snyk-check` to the `fetch-scanner-data` source artifact for consistency

PR #1574, opened 2024-07-19
ROX-20757 scanner multi arch builds

Stringy:
Builds on #1573, adding multi-arch builds (s390x, ppc64le, arm64) to scanner pipeline


.tekton/scanner-component-pipeline.yaml
```diff
@@ -269,12 +331,196 @@ spec:
...
+      value:
+        - $(tasks.build-container-amd64.results.IMAGE_URL)@$(tasks.build-container-amd64.results.IMAGE_DIGEST)
+        - $(tasks.build-container-s390x.results.IMAGE_URL)@$(tasks.build-container-s390x.results.IMAGE_DIGEST)
+        - $(tasks.build-container-ppc64le.results.IMAGE_URL)@$(tasks.build-container-ppc64le.results.IMAGE_DIGEST)
+        - $(tasks.build-container-arm64.results.IMAGE_URL)@$(tasks.build-container-arm64.results.IMAGE_DIGEST)
+    runAfter:
+      - build-container-amd64
+      - build-container-s390x
+      - build-container-ppc64le
+      - build-container-arm64
```
tommartensen, line 470:
```suggestion
```

Not required, Tekton will figure out the order of tasks from the params.

.tekton/scanner-component-pipeline.yaml
```diff
@@ -269,12 +331,196 @@ spec:
...
+      params:
+      - name: name
+        value: apply-tags
+      - name: bundle
+        value: quay.io/redhat-appstudio-tekton-catalog/task-apply-tags:0.1@sha256:29add9a49a2281a3755a9b580d2b9c5cb110231b14cccf8ade2fd7895a9b4b4a
+      - name: kind
+        value: task
+      resolver: bundles
+
+  - name: build-container
```
tommartensen, line 454:
```suggestion
  - name: build-image-manifest
```

Based on the task we use here, I suggest to rename this step to avoid confusion with the actual build container tasks.

.tekton/scanner-component-pipeline.yaml
```diff
@@ -269,12 +331,196 @@ spec:
...
+      value: $(params.output-image-repo):konflux-$(params.revision)
+    - name: COMMIT_SHA
+      value: $(tasks.clone-repository.results.commit)
+    - name: IMAGES
+      value:
+      - $(params.output-image-repo):konflux-$(params.revision)-amd64
+      - $(params.output-image-repo):konflux-$(params.revision)-s390x
+      - $(params.output-image-repo):konflux-$(params.revision)-ppc64le
+      - $(params.output-image-repo):konflux-$(params.revision)-arm64
+    runAfter:
```
tommartensen, line 498: (this `runAfter` is ok, because there is no other way for Tekton to figure out the order of tasks)

.tekton/scanner-component-pipeline.yaml
```diff
@@ -269,12 +331,196 @@ spec:
...
+        - name: kind
+          value: task
+      resolver: bundles
+    when:
+      - input: $(tasks.init.results.build)
+        operator: in
+        values:
+          - "true"
+
+  - name: build-container-konflux
```
tommartensen, line 486:
```suggestion
  - name: build-image-manifest-konflux
```

Based on the task we use here, I suggest to rename this step to avoid confusion with the actual build container tasks.

tommartensen (commented):
I see the pipeline passed, what other validation can we have? 
Run the images on IBM arch VMs?

.tekton/scanner-component-pipeline.yaml
```diff
@@ -250,15 +252,13 @@ spec:
       operator: in
       values: [ "true" ]
 
-  - name: apply-tags
+  - name: apply-tags-amd64
```
msugakov, line 255: In case of `scanner` and most other pipelines (except for `collector`), we have `apply-tags` task to satisfy the `init`'s task `image-url` parameter. (We could have skipped that altogether because `init` does not do much and does even less in our case, but we wanted to stay close to the pipeline template and remain future-proof.)

We don't need per-arch `konflux-$(params.revision)-amd64` tags. The only one we need is `konflux-$(params.revision)` which is naturally placed on the manifest image which you get in `build-image-manifest`.

Would it work if you delete `build-image-manifest-konflux`, delete `apply-tags-<arch>` tasks, and add only one `apply-tags` that works on a result of `build-image-manifest`?

.tekton/scanner-component-pipeline.yaml
```diff
@@ -269,12 +269,249 @@ spec:
         value: task
       resolver: bundles
 
+  - name: build-container-s390x
```
msugakov, line 272: `build-container-s390x`, `ppc64le` and `arm64` should be very similar with the only difference in the `PLATFORM` parameter.

Similarly to GHA, Tekton supports [Matrix](https://tekton.dev/docs/pipelines/matrix/). Could you please try applying that to reduce the amount of copy-paste in these tasks that use `task-buildah-remote-oci-ta`?

.tekton/scanner-component-pipeline.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   - name: show-sbom
     params:
     - name: IMAGE_URL
-      value: $(tasks.build-container.results.IMAGE_URL)
+      value: $(tasks.build-image-manifest.results.IMAGE_URL)
```
msugakov, line 12: Folks here use amd64 image. Not sure if we should too
https://github.com/konflux-ci/build-trusted-artifacts/blob/11bc05e3a30b3870f15b881240515a652cb1e18a/.tekton/build-pipeline.yaml#L10

.tekton/scanner-component-pipeline.yaml
```diff
@@ -236,6 +236,8 @@ spec:
       value: $(tasks.fetch-scanner-data.results.SOURCE_ARTIFACT)
     - name: CACHI2_ARTIFACT
       value: $(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)
+    - name: PLATFORM
+      value: linux/amd64
```
msugakov, line 240: [nit] Since we use non-remote buildah task for amd64, it ignores the `PLATFORM` parameter.
https://github.com/konflux-ci/build-definitions/tree/main/task/buildah-oci-ta/0.2

I suggest to omit it.

```suggestion
```

.tekton/scanner-db-build.yaml
```diff
@@ -68,5 +68,9 @@ spec:
         requests:
           cpu: 1
 
+  # With multi-arch the pipeline regularly takes >1h to finish
+  timeouts:
+    pipeline: 2h30m0s
```
msugakov, line 73: Oh, even DBs need such a large timeout? Asking because we don't have any compilation in dockerfiles and so the time probably goes somewhere else (waiting for a VM?)

Stringy: We've experienced some odd behaviour for the remote builds where they seem to fill the time allocated to them and we think it might be a weird interaction between the taskRunSpec resource requests and the remote builds. We're aiming to simplify this and hopefully remove these extra timeouts

.tekton/scanner-component-pipeline.yaml
Stringy: I did some experiments with matrices at the start, but couldn't get it to work (though I did combine all architectures and had issues with trying to switch around the builder if on amd64) I'll try again

.tekton/scanner-component-pipeline.yaml
Molter73: I've tried this, `apply-tags` will not retag the manifest, it will only retag the image for the architecture that matches the host running the pipeline (`amd64` in this case). Alternatively, we could keep the `build-image-manifest-konflux` step, but have it use the stackrox images instead of applying the tags, WDYT?

msugakov: This is a weird limitation of `apply-tags`. Can you show me a pipelinerun please?

The only reason we have `apply-tags` in all pipelines but Collector is to satisfy the `init` task.

For satisfying the `init` task it should be sufficient to produce an image with a tag that's derived from information available at the pipeline level, for which we use git commit sha - `$(params.output-image-repo):konflux-$(params.revision)`. Since the intent of `init` is to check if the image was previously built, we could either give `amd64` one to it (i.e. keep only `apply-tags` for `amd64` and tag the image as `konflux-$(params.revision)`, no `-amd64` suffix) or create a new manifest tagged as `konflux-$(params.revision)`.

I think, that manifest creation is a cheap operation and it just feels cleaner (e.g. we could succeed building amd64 but fail P or Z) to check presence of builds for all arches. Therefore, I would suggest to try remove all `apply-tags*`, keep `build-image-manifest-konflux` but feed it from the same images as `build-image-manifest`.
Effectively, use `build-image-manifest-konflux` as a way to add tags.

Would this work?

For Collector, we not only create `konflux-$(params.revision)` tag for `init` but we also add `$(tasks.determine-image-tag.results.IMAGE_TAG)-latest` and `$(tasks.determine-image-tag.results.IMAGE_TAG)-slim` for StackRox deployment to succeed because this matches the convention. If tagging via `build-image-manifest*` works, these will be two extra tasks in the pipeline producing manifests tagged with `-slim` and `-latest`.

Molter73: > Can you show me a pipelinerun please?

This was about 20 commits ago, doubt I can find the exact run but will try to run it again.

I agree that generating a proper manifest is cleaner than simply retagging a single arch.

Molter73: @msugakov, sorry for the delay, here is a pipeline for scanner-v4-db where I used apply tags on the manifest: https://console.redhat.com/preview/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-v4-db-build-468sx

If you check in quay.io, you'll see the konflux image sha matches with the amd64 image and not the manifest. This is the same behaviour you'll get from doing `podman/docker tag <MANIFEST> <IMAGE>`, so I believe this is by design. I'll try creating a manifest from the stackrox images next, see if that yields the result we want.

Molter73: Looks like building the manifest twice does produce the manifest correctly: https://console.redhat.com/preview/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-v4-db-build-jzrb4

msugakov: There's `podman manifest inspect` and called for both in `scanner-v4-db-build-jzrb4`, I get identical results.
```bash
$ diff -s <(podman manifest inspect quay.io/rhacs-eng/scanner-v4-db:4.6.x-107-g70adab264b-fast) <(podman manifest inspect quay.io/rhacs-eng/scanner-v4-db:konflux-70adab264b664d3f78ab3a84086bb4dbabb12963)
Files /proc/self/fd/11 and /proc/self/fd/15 are identical
```

Means we can use manifest generation as retagging, right?

Stringy: Actually I've just realised I can use a matrix for that. Ignore me :smile:

msugakov: Yes, that's another matrix. The first one iterates over architectures, the second one iterates over manifest tags (yet each manifest includes image from each arch).

Molter73: > Means we can use manifest generation as retagging, right?

That's my understanding, yes, I have that setup on the [scanner-v4](https://github.com/stackrox/stackrox/pull/12055) and [basic-component](https://github.com/stackrox/stackrox/pull/12042) PRs already.

image/scanner/rhel/konflux.Dockerfile
Stringy, line 1: Changes in this file are due to limitation with tekton matrices:

- "normal" params can't reference matrix params
- matrix params can't be arrays
- BUILD_ARGS must be an array and the tag variable needs to have the arch suffix in order to differentiate between multi-arch images

As a result, I've used the TARGETARCH variable within the Dockerfile to construct the tag here instead. I've done the same for main (https://github.com/stackrox/stackrox/pull/12054) and collector (https://github.com/stackrox/collector/pull/1751)

@tommartensen @msugakov I'd be keen to hear your thoughts on this workaround

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -45,7 +45,7 @@ LABEL \
     summary="The image scanner for Red Hat Advanced Cluster Security for Kubernetes" \
     url="https://catalog.redhat.com/software/container-stacks/detail/60eefc88ee05ae7c5b8f041c" \
     # We must set version label to prevent inheriting value set in the base stage.
-    version="${SCANNER_TAG}" \
+    version="${SCANNER_TAG}-${TARGETARCH}" \
```
msugakov, line 48: I think, this won't harm but isn't necessary. `version` label is just for cosmetics. It has to be there, it has to carry some meaning for a human that's in the know, but that's it. Arch could be derived from other image attributes (there's `architecture` label that seems to be auto-populated).
```suggestion
    version="${SCANNER_TAG}" \
```

and besides `TARGETARCH` isn't declared as `ARG` that's passed to buildah.

```json
            "Labels": {
                "architecture": "x86_64",
                "build-date": "2024-07-26T09:35:20",
                "com.redhat.component": "rhacs-scanner-container",
                ...
                "version": "2.34.x-29-g982071d17e-fast-"
            }
```

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG BASE_TAG=latest
 FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.21 AS builder
 
 ARG SCANNER_TAG
-ENV RELEASE_TAG="${SCANNER_TAG}"
+ENV RELEASE_TAG="${SCANNER_TAG}-${TARGETARCH}"
```
msugakov, line 10: I'd suggest, let's keep `RELEASE_TAG` as it was. There's some elaborate handling where its value gets embeeded in Go binary and later used in some code (the fanout is a bit wide and deep so I procrastinate going there).
```suggestion
ENV RELEASE_TAG="${SCANNER_TAG}"
```

image/scanner/rhel/konflux.Dockerfile
msugakov: Commented in-place below and also I wonder why did you decide this may be needed?

.tekton/scanner-component-pipeline.yaml
```diff
@@ -250,31 +250,127 @@ spec:
       operator: in
       values: [ "true" ]
 
-  - name: apply-tags
+  - name: build-container-multi-arch
+    matrix:
```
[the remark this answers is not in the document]
Stringy: From what I've seen it's inconsistent in its reporting, and seemingly randomly picks a run from the list. I've only observed this while the task is running, however.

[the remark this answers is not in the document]
msugakov: Reported here https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1721997539610399

msugakov (approved):
EC feels unhappy but I think not this change might be causing it.
Green light from me after you address these last findings.

msugakov:
@tommartensen Could you PTAL?

tommartensen (approved):

PR #1586, opened 2024-08-05
chore: Add `^release-` branches for on-push Konflux builds

msugakov:
This should enable Konflux builds from release branches for the case when [SCANNER_VERSION](https://github.com/stackrox/stackrox/blob/master/SCANNER_VERSION) points to a tag/commit that was on a release branch, not on `master`.

See also: https://redhat-internal.slack.com/archives/C079XCJEZK6/p1721729454430659?thread_ts=1721678052.993799&cid=C079XCJEZK6

Prerequisite for <https://issues.redhat.com/browse/ROX-24468>.

Available CEL functions are here <https://github.com/google/cel-spec/blob/master/doc/langdef.md#list-of-standard-definitions>.

### DIFF
--- a/.tekton/scanner-build.yaml
+++ b/.tekton/scanner-build.yaml
@@ -9,7 +9,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
     # TODO(ROX-21073): re-enable for all PR branches
-    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && (target_branch == "master" || target_branch.startsWith("release-"))) || (event == "pull_request" && (source_branch.contains("konflux") || source_branch.contains("appstudio") || source_branch.contains("rhtap")))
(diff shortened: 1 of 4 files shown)


RTann (approved):
LGTM though I don't know why those "acs-enterprise-contract" jobs are failing. When I click on them the UI either gives me a 502 or says the logs don't exist

RTann (commented):
So this will capture all pushes to the release branches, right? Do we want that or to just capture all tag events (if that's posible)?

msugakov:
Thanks for the approval Ross

Regarding acs-enterprise-contract: it's unfortunately broken for all our Konflux builds. We look into this.
It's not stable yet and failures are more common than successes. Additionally, there are sporadic problems with Konflux Console.
It's fine :-) ...for now.

Regarding tags: it's a good question, and we can match on tags (example [here](https://pipelinesascode.com/docs/guide/authoringprs/#matching-an-event-to-a-pipelinerun)) but at this point I fully don't know what we need yet.
I think, running Konflux on pushes to release branches which are not just tags has a benefit of giving a CI signal whether the release branch builds well.

I suggest we start like this and later see how to adjust. We can configure to build on tags _in addition_ or _instead of_ building on release branches, when needed.

PR #1588, opened 2024-08-06
ROX-25620: Switch to new Konflux task repos

msugakov:
Context is in the ticket.
Figured digests similarly to https://github.com/stackrox/collector/pull/1780

`chown /etc/pki` -> `chown /etc/pki/ca-trust` change was tested in https://github.com/stackrox/stackrox/pull/12337

### DIFF
--- a/.tekton/scanner-component-pipeline.yaml
+++ b/.tekton/scanner-component-pipeline.yaml
@@ -15,7 +15,7 @@ spec:
       - name: name
         value: show-sbom
       - name: bundle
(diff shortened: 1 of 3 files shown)


msugakov:
```
chown: changing ownership of '/etc/pki/entitlement': Operation not permitted
```
issue seems persistent. I'll report.

tommartensen (approved):

msugakov:
Ok, that `chown` issue does not get reproduced on plain ubi or ubi-minimal, 8 or 9.
This is due to our stuff here:
https://github.com/stackrox/scanner/blob/499765062d7f5c1246870d88e0ffadc0f05da47f/image/scanner/rhel/konflux.Dockerfile#L74

Need to figure out why it exists.

tommartensen (approved):

PR #1595, opened 2024-08-08
**DO NOT MERGE** Try reproduce the problem for different UBIs

msugakov:
```
chown: changing ownership of '/etc/pki/entitlement': Operation not permitted
```


msugakov:
Served the purpose. Closing.

PR #1597, opened 2024-08-09
fix: validate digest prior to layer download

RTann:


go.mod
```diff
@@ -183,7 +183,7 @@ require (
 replace (
 	github.com/facebookincubator/nvdtools => github.com/stackrox/nvdtools v0.0.0-20231111002313-57e262e4797e
 
-	github.com/heroku/docker-registry-client => github.com/stackrox/docker-registry-client v0.1.0
+	github.com/heroku/docker-registry-client => github.com/stackrox/docker-registry-client v0.2.1-0.20240809225852-a122013c5989
```
RTann, line 186: https://github.com/stackrox/docker-registry-client/pull/21 should be merged and v0.2.1 should be cut prior to merging this PR

dcaravel (approved):
LGTM - assuming tests pass and the dep updated to a tagged version as mentioned.

RTann:
@dcaravel I updated the PR to add some more validation, as I fear the original was not sufficient. I believe now I cover all bases (fetching both manifests and layers. before was just layers).

pkg/scan/scan.go
```diff
@@ -78,36 +81,36 @@ func ProcessImage(storage database.Datastore, image *types.Image, registry, user
...
+	return dig, storage.AddImage(layer, image.SHA, lineage, image.TaggedName(), &database.DatastoreOptions{
 		UncertifiedRHEL: uncertifiedRHEL,
 	})
 }
 
 // process fetches and analyzes the layers for the requested image returning the image digest, the lineage of the last layer and the last layer digest
 func process(storage database.Datastore, image *types.Image, reg types.Registry, uncertifiedRHEL bool) (string, string, string, error) {
 	logrus.Debugf("Processing image %s", image)
-	digest, layers, err := FetchLayers(reg, image)
+	dig, layers, err := FetchLayers(reg, image)
```
RTann, line 99: For anyone wondering, `image` definitely has a valid digest. See https://github.com/stackrox/scanner/blob/2.34.1/pkg/clairify/types/types.go#L178. This is called prior to both calls to `process`. This function returns an image which may have a digest. If it does, it is validated prior to returning it via `reference.ParseAnyReference`.

pkg/scan/scan.go
```diff
@@ -44,6 +46,7 @@ func analyzeLayers(storage database.Datastore, registry types.Registry, image *t
 	for _, layer := range layers {
 		layerReadCloser := &LayerDownloadReadCloser{
 			Downloader: func() (io.ReadCloser, error) {
+				// It is assumed layer is a valid digest.Digest, so no need to parse and validate here.
```
jvdm, line 49: I am curious: Why you did not validate anyway to be on the safe side?

pkg/scan/scan.go
jvdm: https://github.com/stackrox/scanner/blob/2.34.1/pkg/clairify/types/types.go#L181

pkg/scan/scan.go
jvdm: Same for `handleManifest`.

jvdm (approved):
LGTM, just curious about why you expected validation outside the function calls? Not a blocker. I just think when this is the case, functions/methods/routines should still protect themselves from callers, but they don't need to support the invalid input by returning an error -- since this is a programmer's kinda error, they could just enforce and panic (like some standard library methods do for invalid input that cannot be handled in runtime).

pkg/scan/scan.go
RTann: I originally validated here, but I removed it once I (believed) it was accounted for elsewhere. But you're right. Best to be safe

pkg/scan/scan.go
```diff
@@ -78,36 +81,36 @@ func ProcessImage(storage database.Datastore, image *types.Image, registry, user
...
 }
 
 // process fetches and analyzes the layers for the requested image returning the image digest, the lineage of the last layer and the last layer digest
 func process(storage database.Datastore, image *types.Image, reg types.Registry, uncertifiedRHEL bool) (string, string, string, error) {
 	logrus.Debugf("Processing image %s", image)
-	digest, layers, err := FetchLayers(reg, image)
+	dig, layers, err := FetchLayers(reg, image)
 	if err != nil {
-		return digest, "", "", err
+		return dig, "", "", errors.Errorf("fetching layers for %s: %v", image.String(), err)
```
RTann, line 101: I forgot `errors.Wrapf` was a thing, so I'll use that in a followup commit

dcaravel (approved):

pkg/scan/scan.go
RTann: For `handleManifest`, `ref` may not be a digest, but rather a tag, so I didn't want to do the check there

PR #1609, opened 2024-08-15
fix: increase image poll timeout

RTann:
Previous releases done completely via OpenShift CI take... forever. Increase the image poll time so tests can more reliably pass. This will be backported into `release-2.34` and `release-2.33` once merged.


dcaravel (approved):
LGTM

BradLugo (approved):

PR #1643, opened 2024-09-13
ROX-25690: Use versioned offline defs in offline bundles

jvdm:
## Description

Update the offline bundle workflow to use versioned offline definitions (create based on https://github.com/stackrox/stackrox/pull/12516/files) instead of release-based defintions. Also, from 4.6.x onward, versioned offline bundles are published by Z-stream.

## Test

See tests performed at https://github.com/stackrox/stackrox/pull/12895

### DIFF
--- a/scripts/ci/jobs/diff-dumps.sh
+++ b/scripts/ci/jobs/diff-dumps.sh
@@ -77,7 +77,7 @@ upload_diff_dumps() {
     fi
 }
 
-create_offline_dump() {
+scanner_v2_create_and_upload_bundle() {
     info "Creating offline dump"
 
     mkdir -p /tmp/offline-dump
@@ -107,70 +107,89 @@ create_offline_dump() {
     "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://sr-roxc/scanner/scanner-vuln-updates.zip
 }
 
-upload_offline_dump() {
+# scanner_v4_create_and_add_bundles() creates versioned offline bundles for V4
+# and add V4 definitions to the latest offline bundle.  It determines the
+# version to use for each release based on the contents of the
+# VULNERABILITY_VERSION file present in each release tag.
+#
+scanner_v4_create_and_add_bundles() {
+    local latest_bundle="${1:-scanner-vuln-updates.zip}"
+
+    local curl="curl --silent --show-error --max-time 60 --retry 3"
+    local release_version_url="https://raw.githubusercontent.com/stackrox/stackrox/master/scanner/updater/version/RELEASE_VERSION"
+    local vuln_version_pattern="https://raw.githubusercontent.com/stackrox/stackrox/%s/scanner/VULNERABILITY_VERSION"
+
+    [ -f scanner-defs.zip ] || die "missing required bundle inclusion: scanner-defs.zip"
+    [ -f k8s-istio.zip ]    || die "missing required bundle inclusion: k8s-istio.zip"
+    [ -f "$latest_bundle" ] || die "missing latest bundle: $latest_bundle"
+
+    info "existing latest bundle:"
+    unzip -l "$latest_bundle"
+
+    info "fetching release versions at $release_version_url"
+
+    # Read the supported releases file, filter lines with release versions,
+    # parse 4.4.* and 4.5.* to print only the X.Y, otherwise print the whole
+    # X.Y.Z, finally merge everything together, and pipe one release per line.
+
+    $curl --fail "$release_version_url" \
+        | grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' \
+        | sed 's/^\(4\.4\|4\.5\)\..*/\1/' \
+        | sort -V \
+        | uniq \
+        | while read -r release _; do
+              v4_prefix=v4-definitions-
+              case "$release" in
+                  4.4)
+                      # We don't support 4.4 offline bundles.
+                      continue
+                      ;;
+                  4.5)
+                      # Backward compatibility, the "schema version" is the release itself.
+                      version="$release"
+                      v4_prefix="scanner-v4-defs-"
+                      ;;
+                  *)
+                      info "fetching schema version for release $release"
+                      tmp=$(mktemp)
+                      vuln_version_url=$(printf "$vuln_version_pattern" "$release")
+                      case $($curl --write-out "%{http_code}" -o "$tmp" "$vuln_version_url") in
+                          200)
+                              version=$(cat $tmp)
+                              ;;
+                          404)
+                              info "release tag not found, assuming the release was not cut, skipping..."
+                              continue
+                              ;;
+                          *)
+                              die "failed to fetch the v4 offline bundle version for $release: status $status"
+                              ;;
+                      esac
+                      ;;
+              esac
+
+              local v4_bundle="$v4_prefix$version.zip"
+              info "building $release using schema version $version: filename: $v4_bundle"
+
+              $curl \
+                  --fail \
+                  -o "$v4_bundle" \
+                  "https://definitions.stackrox.io/v4/offline-bundles/$v4_bundle"
+              zip "$latest_bundle" "$v4_bundle"
+              zip "scanner-vulns-$release.zip" scanner-defs.zip k8s-istio.zip "$v4_bundle"
+          done
+}
+
+upload_bundles() {
     info "Uploading offline dump"
-
-    cd /tmp/offline-dump
-
-    curl --silent --show-error --max-time 60 --retry 3 --create-dirs -o out/RELEASE_VERSION.txt https://raw.githubusercontent.com/stackrox/stackrox/master/scanner/updater/version/RELEASE_VERSION
-    version_file="out/RELEASE_VERSION.txt"
-
-    latest_version=""
-    # Get the latest X.Y version which has been released.
-    while read -r version; do
-        status_code=$(curl --silent --max-time 60 --retry 3 --write-out "%{http_code}" -o /dev/null "https://github.com/stackrox/stackrox/releases/tag/$version.0")
-        if [[ "$status_code" == "404" ]]; then
-            # Release was not cut, so try a previous release.
-            continue
-        elif [[ "$status_code" == "200" ]]; then
-            # Release was cut, so use this one.
-            latest_version="$version"
-            break
-        else
-            # Some other error occurred.
-            >&2 echo "received status code $status_code when fetching StackRox releases"
-            exit 1
-        fi
-    done < <(grep -oE '^[0-9]+\.[0-9]+' "$version_file" | sort --reverse --unique --version-sort)
-    if [[ -z "$latest_version" ]]; then
-      >&2 echo "programmer error: latest_version unset..."
-      exit 1
-    fi
-
-    file_to_check="scanner-v4-defs-${latest_version}.zip"
-
-    curl --silent --show-error --fail --max-time 60 --retry 3 -o "$file_to_check" "https://definitions.stackrox.io/v4/offline-bundles/$file_to_check"
-    unzip -l "$file_to_check"
-    zip scanner-vuln-updates.zip "$file_to_check"
-    echo "$file_to_check added to scanner-vuln-updates.zip"
-
     cmd=()
     if is_in_PR_context; then
         cmd+=(echo "Would do")
     fi
     "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://scanner-support-public/offline/v1/scanner-vuln-updates.zip
-}
-
-upload_v4_versioned_vuln() {
-    info "Uploading v4 offline dump"
-    cmd=()
-    if is_in_PR_context; then
-        cmd+=(echo "Would do")
-    fi
-    cd /tmp/offline-dump
-
-    cat out/RELEASE_VERSION.txt |
-        grep -oE '^[0-9]+\.[0-9]+' |
-        sort -V |
-        uniq |
-    while read -r version; do
-        echo "$version"
-        if curl --silent --show-error --max-time 60 --retry 3 -o "scanner-v4-defs-${version}.zip" "https://definitions.stackrox.io/v4/offline-bundles/scanner-v4-defs-${version}.zip"; then
-            zip scanner-vulns-${version}.zip scanner-defs.zip k8s-istio.zip scanner-v4-defs-${version}.zip
-            "${cmd[@]}" gsutil cp scanner-vulns-${version}.zip gs://scanner-support-public/offline/v1/${version}/scanner-vulns-${version}.zip
-        else
-            echo "Failed to download scanner-v4-defs-${version}.zip, skipping..."
-        fi
+    for f in scanner-vulns-*.zip; do
+        version=$(echo "$f" | sed 's/scanner-vulns-\(.*\)\.zip/\1/')
+        "${cmd[@]}" gsutil cp "$f" gs://scanner-support-public/offline/v1/${version}/scanner-vulns-${version}.zip
     done
 }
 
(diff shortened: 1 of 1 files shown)


scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
+
+    info "fetching release versions at $release_version_url"
+
+    # Read the supported releases file, filter lines with release versions,
+    # parse 4.4.* and 4.5.* to print only the X.Y, otherwise print the whole
+    # X.Y.Z, finally merge everything together, and pipe one release per line.
+
+    $curl --fail "$release_version_url" \
+        | grep '^[0-9]\.[0-9]\.[0-9]$' \
+        | sed -En 's/^(4\.4|4\.5)\..*/\1/p;t; p' \
```
dcaravel, line 137: nit: on mac this produces an error, can it be made more cross-platform friendly?

```
sed: 1: "s/^(4\.4|4\.5)\..*/\1/p ...": undefined label '; p'
```

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
+    unzip -l "$latest_bundle"
+
+    info "fetching release versions at $release_version_url"
+
+    # Read the supported releases file, filter lines with release versions,
+    # parse 4.4.* and 4.5.* to print only the X.Y, otherwise print the whole
+    # X.Y.Z, finally merge everything together, and pipe one release per line.
+
+    $curl --fail "$release_version_url" \
+        | grep '^[0-9]\.[0-9]\.[0-9]$' \
```
dcaravel, line 136: Any concerns about us reaching double digits eventually?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
+              $curl \
+                  --fail \
+                  -o "$v4_bundle" \
+                  "https://definitions.stackrox.io/v4/offline-bundles/$v4_bundle"
+              zip "$latest_bundle" "$v4_bundle"
+              zip "scanner-vulns-$release.zip" scanner-defs.zip k8s-istio.zip "$v4_bundle"
+          done
+}
 
+scanner_v4_upload_bundles() {
```
dcaravel, line 181: nit: 
```suggestion
scanner_upload_bundles() {
```
or revert it back to the old name, `upload_offline_dumps`, since the upload is not V4 specific.

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
         cmd+=(echo "Would do")
     fi
     "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://scanner-support-public/offline/v1/scanner-vuln-updates.zip
+    for f in scanner-vulns-*.zip; do
+        version=$(echo "$f" | sed 's/scanner-vunls-\(.*\)\.zip/\1/')
+        "${cmd[@]}" gsutil cp "$f" gs://scanner-support-public/offline/v1/${version}/scanner-vulns-${version}.zip
+    done
 }
 
 upload_v4_versioned_vuln() {
```
dcaravel, line 194: nit: remove `upload_v4_versioned_vuln`, it no longer appears used.

jvdm: 92a330e0

scripts/ci/jobs/diff-dumps.sh
jvdm: 92a330e0

scripts/ci/jobs/diff-dumps.sh
jvdm: 92a330e0

scripts/ci/jobs/diff-dumps.sh
jvdm: > on mac this produces an error, can it be made more cross-platform friendly?

This script depends on `sed` as shipped by the CI image `quay.io/stackrox-io/apollo-ci:scanner-test-0.4.3`. I mean, it's OK to work it out so you can run this on any `sed` implementation, but I don't think we should try to achieve this considering we can run/test this inside `quay.io/stackrox-io/apollo-ci:scanner-test-0.4.3`.

Saying that, in this case, there is a simpler expression that I believe is supported by multiple implementations (although, I cannot fully assure you as I don't know which one you're using, so I am hoping for the best).

92a330e0

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
-    curl --silent --show-error --fail --max-time 60 --retry 3 -o "$file_to_check" "https://definitions.stackrox.io/v4/offline-bundles/$file_to_check"
-    unzip -l "$file_to_check"
-    zip scanner-vuln-updates.zip "$file_to_check"
-    echo "$file_to_check added to scanner-vuln-updates.zip"
+# scanner_v4_build_bundles() creates versioned offline bundles for V4 and add V4
+# definitions to the latest offline bundle.  It determines the version to use
+# for each release based on the VULNERABILITY_VERSION value for each release
+# tag.
+#
+scanner_v4_create_and_add_bundles() {
```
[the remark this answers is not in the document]
jvdm: 92a330e0

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,48 +107,88 @@ create_offline_dump() {
...
+
+    $curl --fail "$release_version_url" \
+        | grep '^[0-9]\.[0-9]\.[0-9]$' \
+        | sed -En 's/^(4\.4|4\.5)\..*/\1/p;t; p' \
+        | sort -V \
+        | uniq \
+        | while read -r release _; do
+              case "$release" in
+                  4.4)
+                      # We don't support 4.4 offline bundles.
```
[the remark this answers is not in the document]
jvdm: This is as of today, you might know. I am just figuring that adding one additional offline definition zip inside the bundle does not worth it since we don't add it currently.

jvdm:
> Preliminary review:

Could you please send me the full review? I am OK waiting a bit for a full review, in case you're holding the remainder waiting for my replies on this preliminary.

dcaravel:
> Could you please send me the full review? I am OK waiting a bit for a full review, in case you're holding the remainder waiting for my replies on this preliminary.

Called it preliminary due to not being tested yet:

image

RTann:
> In progress, waiting on https://github.com/stackrox/stackrox/pull/12516/files for full end-to-end testing.

Should we wait for that PR to be merged and for you to do your testing before reviewing this?

RTann:
> > In progress, waiting on https://github.com/stackrox/stackrox/pull/12516/files for full end-to-end testing.
> 
> Should we wait for that PR to be merged and for you to do your testing before reviewing this?

I see this was already asked, and it seems like you still want a review, so I'll look

RTann:
Sorry I pressed the wrong button when submitting my comment 😬

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,70 +107,87 @@ create_offline_dump() {
...
+
+    info "fetching release versions at $release_version_url"
+
+    # Read the supported releases file, filter lines with release versions,
+    # parse 4.4.* and 4.5.* to print only the X.Y, otherwise print the whole
+    # X.Y.Z, finally merge everything together, and pipe one release per line.
+
+    $curl --fail "$release_version_url" \
+        | grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' \
+        | sed 's/^\(4\.4\|4\.5\)\..*/\1/' \
```
RTann, line 137: What does this do? I tried it out on the current file and a modified version where I added `4.5.4-rc.1` and `4.6.0`, but it all looks the same

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,70 +107,87 @@ create_offline_dump() {
...
+        | sort -V \
+        | uniq \
```
RTann, line 139: I made a file as such:

```
4.4.0
4.4.1
4.4.2
4.4.3
4.4.4
4.4.5
4.5.0
4.5.1
4.5.2
4.5.3
4.5.4-rc.1
4.6.0
```

and I ran it through all commands until after `uniq`, and it looks the same as before

scripts/ci/jobs/diff-dumps.sh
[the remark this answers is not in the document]
dcaravel: Makes sense, and pretty sure if we do add the additional file to the bundle it will break either 4.4 or 4.5 loads given how those Central versions perform the bundle version check

scripts/ci/jobs/diff-dumps.sh
dcaravel: Makes sense! Didn't realize these were run inside `quay.io/stackrox-io/apollo-ci:scanner-test-0.4.3`. In that case is/was OK.

scripts/ci/jobs/diff-dumps.sh
jvdm: This reads the supported releases file, filter lines with release versions, parse `4.4.*` and `4.5.*` to print only the `X.Y`, otherwise print the whole `X.Y.Z`, finally merge everything together, and pipe one release per line.

So `4.5.4-rc.1` should be filtered out, but `4.6.0` should be there.

scripts/ci/jobs/diff-dumps.sh
jvdm: `4.6.0` should be there.

```
⮩  cat RELEASE_VERSION
# This file serves as a list of StackRox versions.
# As Scanner V4 and StackRox versions are aligned, the vulnerability updaters are responsible for updating vulnerability data across all versions listed in this file.
# Subsequently, this updated data is uploaded to Google Cloud Storage.
# For each individual version of Scanner V4's vulnerability updater, corresponding vulnerability data is collected and then seamlessly uploaded to Google Cloud Storage.
# This process ensures that each version's vulnerability data is accurately captured and maintained in accordance with its respective version.
4.4.0
4.4.1
4.4.2
4.4.3
4.4.4
4.4.5
4.5.0
4.5.1
4.5.2
4.5.3
4.6.0
4.5.4-rc.1
⮩  cat RELEASE_VERSION | grep 4.6.0
4.6.0
⮩  cat RELEASE_VERSION | grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' | sed 's/^\(4\.4\|4\.5\)\..*/\1/'
4.4
4.4
4.4
4.4
4.4
4.4
4.5
4.5
4.5
4.5
4.6.0
⮩  cat RELEASE_VERSION | grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' | sed 's/^\(4\.4\|4\.5\)\..*/\1/'  | grep 4.6.0
4.6.0
⮩  
```

RTann:
```
$ cat scanner/updater/version/RELEASE_VERSION
# This file serves as a list of StackRox versions.
# As Scanner V4 and StackRox versions are aligned, the vulnerability updaters are responsible for updating vulnerability data across all versions listed in this file.
# Subsequently, this updated data is uploaded to Google Cloud Storage.
# For each individual version of Scanner V4's vulnerability updater, corresponding vulnerability data is collected and then seamlessly uploaded to Google Cloud Storage.
# This process ensures that each version's vulnerability data is accurately captured and maintained in accordance with its respective version.
4.4.0
4.4.1
4.4.2
4.4.3
4.4.4
4.4.5
4.5.0
4.5.1
4.5.2
4.5.3
4.6.0
4.5.4-rc.1
$ cat scanner/updater/version/RELEASE_VERSION |  grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' | sed 's/^\(4\.4\|4\.5\)\..*/\1/'  | sort -V | uniq
4.4.0
4.4.1
4.4.2
4.4.3
4.4.4
4.4.5
4.5.0
4.5.1
4.5.2
4.5.3
4.6.0
```

There must be something wrong with my `sed` (builtin `sed` on macOS)

I tried again in a UBI 8 image:

```
# cat RELEASE_VERSION
# This file serves as a list of StackRox versions.
# As Scanner V4 and StackRox versions are aligned, the vulnerability updaters are responsible for updating vulnerability data across all versions listed in this file.
# Subsequently, this updated data is uploaded to Google Cloud Storage.
# For each individual version of Scanner V4's vulnerability updater, corresponding vulnerability data is collected and then seamlessly uploaded to Google Cloud Storage.
# This process ensures that each version's vulnerability data is accurately captured and maintained in accordance with its respective version.
4.4.0
4.4.1
4.4.2
4.4.3
4.4.4
4.4.5
4.5.0
4.5.1
4.5.2
4.5.3
4.6.0
4.5.4-rc.1
# cat RELEASE_VERSION | grep '^[0-9]\+\.[0-9]\+\.[0-9]\+$' | sed 's/^\(4\.4\|4\.5\)\..*/\1/'  | sort -V | uniq
4.4
4.5
4.6.0
```

RTann: marking this as resolved since we run this on Linux which has a version of `sed` that isn't messed up

scripts/ci/jobs/diff-dumps.sh
RTann: resolving this. The version of `sed` shipped with my computer is messed up

scripts/ci/jobs/diff-dumps.sh
jvdm: If you want to run the same `sed` used on GNU/Linux you might want to start with:

```
brew install gnu-sed
```

I've been told that this will install the GNU port of `sed`; I don't think we are using any GNU extension in the sed expression above, but at least the POSIX implementation should be the same.

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,70 +107,87 @@ create_offline_dump() {
...
+        | while read -r release _; do
+              case "$release" in
+                  4.4)
+                      # We don't support 4.4 offline bundles.
+                      continue
+                      ;;
+                  4.5)
+                      # Backward compatibility, the "schema version" is the release itself.
+                      version="$release"
```
RTann, line 148: what happens in the next iteration of the loop? seems like `version` gets overwritten

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,70 +107,124 @@ create_offline_dump() {
...
-        cmd+=(echo "Would do")
-    fi
-    "${cmd[@]}" gsutil cp scanner-vuln-updates.zip gs://scanner-support-public/offline/v1/scanner-vuln-updates.zip
+# scanner_v4_create_and_add_bundles() creates versioned offline bundles for V4
+# and add V4 definitions to the latest offline bundle.  It determines the
+# version to use for each release based on the contents of the
+# VULNERABILITY_VERSION file present in each release tag.
+#
+scanner_v4_create_and_add_bundles() {
+    local latest_bundle="${1:-scanner-vuln-updates.zip}"
```
RTann, line 116: is this for testing?

scripts/ci/jobs/diff-dumps.sh
```diff
@@ -107,70 +107,124 @@ create_offline_dump() {
...
+        case "$release" in
+            4.4)
+                # We don't support 4.4 offline bundles.
+                continue
+                ;;
+            4.5)
+                # Backward compatibility, the "schema version" is the release itself.
+                version="$release"
+                ;;
+            *)
```
RTann, line 161: So we do plan on continuing to update `RELEASE_VERSION`? I thought we weren't going to do so. https://github.com/stackrox/stackrox/blob/master/.github/workflows/scanner-versioned-definitions-update.yaml#L319

RTann (commented):
The description of this PR says it's still untested. Is that still true?

scripts/ci/jobs/diff-dumps.sh
jvdm: It should, right? For each iteration we want to determine the `vesion` for the given `release`. For 4.5 they are the same (which is weird to read -- but convenient because we can treat 4.5 schema is called "4.5"). For later releases it should be v1, v2, etc.

RTann: I misread this and I thought the loop ended at the `esac` and not the `done` a few lines lower. Makes sense now

scripts/ci/jobs/diff-dumps.sh
jvdm: That's for testing, yes.

scripts/ci/jobs/diff-dumps.sh
jvdm: The file will not be used to determine versioned bundles to build anymore. But, we will continue to use it to list all release tags where V4 is offered.

I might have clarified that elsewhere, but again, the file is not strictly necessary, since we can derive it from other contexts (like the browsing the references in the git repository). But, it is convenient, and I don't want to jump into adopting other methods before I understand how the release tagging is managed (I have ROX-26053 to investigate further and continue from there). Also the change scope and potential blast radius is reduced if I use that file, for now.

jvdm:
> The description of this PR says it's still untested. Is that still true?

No, tests were made on dev builds for 4.6.x, see description.

PR #1677, opened 2024-10-17
fix: Add additional deps to the notification job

jvdm:
Add some leaf jobs in the dependency tree that were missing in the notification job dependencies, so that if these path fail, we see notifications.

image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -684,6 +684,10 @@ jobs:
   send-notification:
     needs:
       - diff-dumps
+      - upload-db-dump
+      - upload-dumps-for-downstream
+      - upload-dumps-for-embedding
+      - push-manifests
     runs-on: ubuntu-latest
     if: failure()
     steps:


dcaravel (approved):
This seems OK to me (based on the graph).

.github/workflows/ci.yaml
```diff
@@ -684,6 +684,10 @@ jobs:
   send-notification:
     needs:
       - diff-dumps
+      - upload-db-dump
+      - upload-dumps-for-downstream
+      - upload-dumps-for-embedding
+      - push-manifests
     runs-on: ubuntu-latest
     if: failure()
```
BradLugo, line 692: Do we want to be notified if these jobs fail at _any_ point, e.g., including PRs? If not, we might want to add additional checks (not 100% sure if this is correct):
```suggestion
    if: failure() && (github.ref_name == 'master' || startsWith(github.ref, 'refs/tags/'))
```

PR #1684, opened 2024-10-23
ROX-26094: variable blob location for Konflux

tommartensen:
Idea:
* SCANNER_TAG is passed from determine-image-tag
* If it is not a tagged build, download data from latest
* Else use the SCANNER_TAG

Hints
* switched to full ubi because git didn't install properly: https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-build-vn9l9/logs (`error: Error -1 running transaction`)
* switched to ubi 9 (from 8), because newer curl required for `--retry-all-errors` flag
* The logic to determine if we're on a tagged commit follows the pattern implemented in the determine-image-tag task.

Validation:
* tagged build
  * GHA (produces dumps): https://github.com/stackrox/scanner/actions/runs/11813726639
  * Konflux (from [0.0.0-test.10](https://github.com/stackrox/scanner/tree/refs/tags/0.0.0-test.10)): https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-wvgjg (logs gone)
* untagged build
  * Konflux: https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-9q5b9 (logs gone)
* abort double tagged build: https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-stz5w

## New validation
* Tagged build: 
  * [PipelineRun](https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-6lwr7)
  * [fetch-scanner-data log](https://github.com/user-attachments/files/17734703/scanner-on-push-6lwr7-fetch-scanner-data.log)

* Untagged build: 
  * [PipelineRun](https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-bklxk) 
  * [fetch-scanner-data log](https://github.com/user-attachments/files/17734459/scanner-on-push-bklxk-fetch-scanner-data.log)

TODO:
* [ ] remove 0.0.0-test.x tags


.tekton/scanner-build.yaml
```diff
@@ -10,8 +10,9 @@ metadata:
     pipelinesascode.tekton.dev/max-keep-runs: "500"
     # TODO(ROX-21073): re-enable for all PR branches
     pipelinesascode.tekton.dev/on-cel-expression: |
-      (event == "push" && target_branch.matches("^(master|release-.*)$")) ||
```
tommartensen, line 13: TODO: revert this change. It is present to trigger for tags.

.tekton/scanner-build.yaml
```diff
@@ -10,8 +10,9 @@ metadata:
...
+      (event == "push") ||
       (event == "pull_request" && (source_branch.matches("(konflux|renovate|appstudio|rhtap)") || body.pull_request.labels.exists(l, l.name == "konflux-build")))
+
```
tommartensen, line 15: TODO: revert this empty line change

scripts/konflux/fetch-scanner-data.sh
```diff
@@ -16,14 +16,28 @@ shift
...
 
-  # TODO(ROX-22130): Assign proper suffix for tagged commits instead of /latest/.
-  url="https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob}"
+  url="https://storage.googleapis.com/definitions.stackrox.io/scanner-data/${SCANNER_DATA_VERSION}/${blob}"
   dest="${TARGET_DIR}/blob-${blob}"
 
-  echo "Downloading ${url} > ${dest}"
-  curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
+  echo "Downloading ${url} > ${dest}, retrying 1000 times or until killed..."
+  curl --fail -s --show-error --retry 1000 --retry-delay 10 --retry-all-errors \
```
msugakov, line 40: Please add an explanation why we do this and why we retry on 4xx errors.

.tekton/scanner-build.yaml
```diff
@@ -72,9 +72,9 @@ spec:
 
   # Multiarch builds sometimes make the pipeline timeout after 1h.
   timeouts:
```
msugakov, line 74: It's worth a mention that the timeout accommodates maximum time to fetch blobs.

.tekton/scanner-db-build.yaml
```diff
@@ -60,9 +60,9 @@ spec:
       secretName: '{{ git_auth_secret }}'
 
   timeouts:
```
msugakov, line 62: Here too.

.tekton/fetch-scanner-data-task.yaml
```diff
@@ -56,9 +56,11 @@ spec:
     script: |
       #!/usr/bin/env bash
       set -euo pipefail
+      dnf -y upgrade --nobest
+      dnf -y install git
       scripts/konflux/fetch-scanner-data.sh "$(params.target-dir)" "$@"
-    # It should not take long to download blobs otherwise there's something odd going on.
-    timeout: 10m
+    # It should not take too long to download blobs otherwise there's something odd going on.
```
msugakov, line 62: I suggest to convert this comment into a message (e.g. printed by `fetch-scanner-data.sh`) which would say that the download timeout, if it happens, is likely due to not published blobs and would suggest where to go check publishing status.

scripts/konflux/fetch-scanner-data.sh
msugakov, line 40: Unfortunately, the pipeline runs you linked are gone as well as logs. I wonder how does the output look when curl is retrying. Would an engineer see the curl is active or would there just be silence in logs?
It may be better to remove `-s` (we're not redirecting stdout anyway) because that way `curl` shows remaining retries for me.

scripts/konflux/fetch-scanner-data.sh
```diff
@@ -16,14 +16,28 @@ shift
...
 for blob in "${blobs[@]}"; do
 
-  # TODO(ROX-22130): Assign proper suffix for tagged commits instead of /latest/.
-  url="https://storage.googleapis.com/definitions.stackrox.io/scanner-data/latest/${blob}"
+  url="https://storage.googleapis.com/definitions.stackrox.io/scanner-data/${SCANNER_DATA_VERSION}/${blob}"
   dest="${TARGET_DIR}/blob-${blob}"
 
-  echo "Downloading ${url} > ${dest}"
-  curl --fail -s --show-error --retry 4 --retry-max-time 30 --retry-connrefused \
+  echo "Downloading ${url} > ${dest}, retrying 1000 times or until killed..."
```
msugakov, line 39: Since we don't need to wait for `latest` blobs to appear, they are already there, I suggest assigning lower retry count to it. This is a small optimization but may give us quicker pipelines if things go wrong for most common `master` and PR builds.

msugakov (commented):
Cosmetic comments.

scripts/konflux/fetch-scanner-data.sh
tommartensen: Something like

```
Downloading https://some.url/file > .file, retrying 1000 times or until killed...
curl: (22) The requested URL returned error: 404
curl: (22) The requested URL returned error: 404
...
```

ie a new 404 every 10 seconds. 

Other options:
* Removing the `--silent` option would additionally show the progress bar for each retry and a warning.

```
Downloading https://some.url/file > .file, retrying 1000 times or until killed...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0  1565    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Warning: Problem (retrying all errors). Will retry in 10 seconds. 10 retries 
Warning: left.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0  1565    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Warning: Problem (retrying all errors). Will retry in 10 seconds. 10 retries 
Warning: left.
...
```

* Adding the `--no-progress-bar` option hides the progress bar, but adds 
```
Downloading https://some.url/file > .file, retrying 1000 times or until killed...
curl: (22) The requested URL returned error: 404
Warning: Problem (retrying all errors). Will retry in 10 seconds. 9 retries 
Warning: left.
curl: (22) The requested URL returned error: 404
Warning: Problem (retrying all errors). Will retry in 10 seconds. 9 retries 
Warning: left.
```

tommartensen: Out of these three options, I would prefer the first, least verbose one.

tommartensen:
* Tagged build: https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-6lwr7
* Untagged build: https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-on-push-bklxk

.tekton/fetch-scanner-data-task.yaml
tommartensen: Updated the message and added to the script.

.tekton/scanner-db-build.yaml
tommartensen: Added

scripts/konflux/fetch-scanner-data.sh
msugakov: I vote for `--no-progress-bar`. The reason is that the retires counter (which should decrease unlike in your samples above) gives an idea that (a) it's doing something (it may be hard to tell if a new line was logged if the whole screen is filled with the same), (b) for how long it was doing something.

tommartensen: Done in https://github.com/stackrox/scanner/pull/1684/commits/4f124a3cf8115204573cfaba4d36f1db1f06e8aa

(the reason the counter doesn't decrease is because I manually constructed the output, it does in fact decrease and not a curl bug).

msugakov (approved):

PR #1709, opened 2024-11-15
ROX-24283: enable strictfipsruntime in the Konflux build

BradLugo:
## Description

Enables the strictfipsruntime build flag for the Konflux build.

For more info about the strictfipsruntime flag, see [this doc](https://docs.google.com/document/d/1CTpSwITQfOgoOTlPITZNJZy0ltYz60hxkmWsLemCqw0/edit?tab=t.0) (there might be a better resource but this is the one David and I found), and for our general research regarding the linked ticket, see [this doc](https://docs.google.com/document/d/1KcjlevcMF4DLi-KROru7tuUUIb4wSk5e6bnXpU8HZxA/edit?tab=t.0).

Related stackrox PR: https://github.com/stackrox/stackrox/pull/12909

Verified the build works via Konflux CI and verified the `check-payload` results of the following images:
- [x] ` quay.io/rhacs-eng/scanner:2.35.x-13-g2c519017f0-fast`
- [x] ` quay.io/rhacs-eng/scanner-slim:2.35.x-13-g2c519017f0-fast`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,8 +71,7 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS   := -e CGO_ENABLED=$(CGO_ENABLED) -e GOOS=linux -e GOARCH=$(GOARCH)
 IMAGE_BUILD_ARGS     = --build-arg LABEL_VERSION=$(TAG) --build-arg LABEL_RELEASE=$(TAG) --build-arg QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION)
 BUILD_FLAGS         := CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH)
-GOTAGS              ?=
-BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" $(if $(GOTAGS),-tags=$(GOTAGS)) -o image/scanner/bin/scanner ./cmd/clair
+BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -tags=$(GOTAGS) -o image/scanner/bin/scanner ./cmd/clair
 NODESCAN_BUILD_CMD  := go build -trimpath -o tools/bin/local-nodescanner ./tools/local-nodescanner
 
 #####################################################################
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,8 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS   := -e CGO_ENABLED=$(CGO_ENABLED) -e GOOS=linux -e GOARCH=$(GOARCH)
 IMAGE_BUILD_ARGS     = --build-arg LABEL_VERSION=$(TAG) --build-arg LABEL_RELEASE=$(TAG) --build-arg QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION)
 BUILD_FLAGS         := CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH)
-BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -o image/scanner/bin/scanner ./cmd/clair
+GOTAGS              ?=
+BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" $(if $(GOTAGS),-tags=$(GOTAGS)) -o image/scanner/bin/scanner ./cmd/clair
 NODESCAN_BUILD_CMD  := go build -trimpath -o tools/bin/local-nodescanner ./tools/local-nodescanner
 
 #####################################################################
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS   := -e CGO_ENABLED=$(CGO_ENABLED) -e GOOS=linux -e GOARCH=$(GOARCH)
 IMAGE_BUILD_ARGS     = --build-arg LABEL_VERSION=$(TAG) --build-arg LABEL_RELEASE=$(TAG) --build-arg QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION)
 BUILD_FLAGS         := CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH)
-BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -tags=$(GOTAGS) -o image/scanner/bin/scanner ./cmd/clair
+BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -tags="$(GOTAGS)" -o image/scanner/bin/scanner ./cmd/clair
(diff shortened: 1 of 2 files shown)


BradLugo:
@davdhacs do you mind double-checking the images listed in the description?

Makefile
```diff
@@ -71,7 +71,8 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS   := -e CGO_ENABLED=$(CGO_ENABLED) -e GOOS=linux -e GOARCH=$(GOARCH)
 IMAGE_BUILD_ARGS     = --build-arg LABEL_VERSION=$(TAG) --build-arg LABEL_RELEASE=$(TAG) --build-arg QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION)
 BUILD_FLAGS         := CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH)
-BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -o image/scanner/bin/scanner ./cmd/clair
+GOTAGS              ?=
```
msugakov, line 74: Make does not require declaring variables. Consequently, Make does not fail when using undeclared variables.
```suggestion
```

Makefile
```diff
@@ -71,7 +71,8 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
...
+BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" $(if $(GOTAGS),-tags=$(GOTAGS)) -o image/scanner/bin/scanner ./cmd/clair
```
msugakov, line 75: After looking at the stackrox repo, it seems that `go` tolerates empty argument of `-tags` option: `-tags ""`.
Is `$(if` guard is really necessary, i.e. can you please show me an example of this command failing when GOTAGS variable is empty?

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -10,6 +10,8 @@ ARG SCANNER_TAG
 RUN if [[ "$SCANNER_TAG" == "" ]]; then >&2 echo "error: required SCANNER_TAG arg is unset"; exit 6; fi
 ENV RELEASE_TAG="${SCANNER_TAG}"
 
+ENV GOEXPERIMENT=strictfipsruntime
+ENV GOTAGS=strictfipsruntime
```
msugakov, line 14: I forgot: do these both `GOEXPERIMENT=strictfipsruntime` env variable and `-tags=strictfipsruntime` argument need to be provided at the same time?

Makefile
BradLugo: I was thinking it'd be nice to see all the variables used in the Makefile in one place (or as many as possible). I'll remove it

image/scanner/rhel/konflux.Dockerfile
BradLugo: I tried doing this in https://github.com/stackrox/stackrox/pull/12909, but only setting `GOEXPERIMENT` didn't move the needle (hence this commit: https://github.com/stackrox/stackrox/pull/12909/commits/0e094ea35b90da1473541c6ba205e61cfe779745)

msugakov: Weird. If `-tags=strictfipsruntime`/`GOTAGS=strictfipsruntime` has the effect, maybe we don't need `GOEXPERIMENT=strictfipsruntime`?

Makefile
msugakov: Less code branching makes maintenance easier.

image/scanner/rhel/konflux.Dockerfile
msugakov, line 14: +1 to @davdhacs suggestion to add a `# TODO(ROX-1234):` reminder for us to remove one or the other thing.

msugakov (changes requested):
Since there's one approval, I post this as "Request changes" merely to prevent merging without the `TODO`. I think, it's essential to have one as a reminder.

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -10,6 +10,9 @@ ARG SCANNER_TAG
 RUN if [[ "$SCANNER_TAG" == "" ]]; then >&2 echo "error: required SCANNER_TAG arg is unset"; exit 6; fi
 ENV RELEASE_TAG="${SCANNER_TAG}"
 
+# TODO(ROX-27054): Remove the redundant strictfipsruntime option if one is found to be so
```
BradLugo, line 13: I don't _love_ the way this comment reads, but it's the best I could come up with.

Makefile
```diff
@@ -71,7 +71,7 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
 IMAGE_BUILD_FLAGS   := -e CGO_ENABLED=$(CGO_ENABLED) -e GOOS=linux -e GOARCH=$(GOARCH)
 IMAGE_BUILD_ARGS     = --build-arg LABEL_VERSION=$(TAG) --build-arg LABEL_RELEASE=$(TAG) --build-arg QUAY_TAG_EXPIRATION=$(QUAY_TAG_EXPIRATION)
 BUILD_FLAGS         := CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH)
-BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -o image/scanner/bin/scanner ./cmd/clair
+BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -tags=$(GOTAGS) -o image/scanner/bin/scanner ./cmd/clair
```
msugakov, line 74: Sorry, I just realized it's safer for the future when `GOTAGS` are quoted as tags can be space-separated.
```suggestion
BUILD_CMD           := go build -trimpath -ldflags="-X github.com/stackrox/scanner/pkg/version.Version=$(TAG)" -tags="$(GOTAGS)" -o image/scanner/bin/scanner ./cmd/clair
```

msugakov (approved):
One last comment and I'm happy with scanner.

msugakov (approved):
I don't want to hold this for longer, but please double-check that the other two executables aren't included in the image.

Makefile
```diff
@@ -71,7 +71,7 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
...
 NODESCAN_BUILD_CMD  := go build -trimpath -o tools/bin/local-nodescanner ./tools/local-nodescanner
```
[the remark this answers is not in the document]
RTann: It's a development tool. From what I can tell, the (old) RHCOS node scanner runs as a container in the collector pod called `node-inventory` which runs the usual scanner binary `/scanner` from within the usual scanner image `._rox.image.scanner.fullRef` https://github.com/stackrox/stackrox/blob/master/image/templates/helm/stackrox-secured-cluster/templates/collector.yaml.htpl#L196

Looks like the typical scanner main function was modified to run the scanner in "node inventory mode" - https://github.com/stackrox/scanner/blob/master/cmd/clair/main.go#L279

So I'd say it's accurate to keep this command out

Makefile
```diff
@@ -71,7 +71,7 @@ GOPATH_WD_OVERRIDES := -w /src -e GOPATH=/go
...
```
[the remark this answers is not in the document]
RTann: This updater just runs in CI and isn't shipped to customers, so no need to be FIPS-built

image/scanner/rhel/konflux.Dockerfile
RTann: What's the conclusion of this? Just keeping both?

Makefile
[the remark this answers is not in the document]
Maddosaurus: Yes, this is correct to not be included in the image.
`local-nodescanner` is a tool that enables local testing and debugging of Scanner v2 Node Scanning code without the need to deploy ACS.
It is similar to the tools for collector and sensor we have over in the main stackrox repo :smile:

msugakov:
@BradLugo Looks like Konflux builds completed fine. Don't worry about `acs-enterprise-contract` failures.

PR #1742, opened 2024-12-12
ROX-27350: Use trusted tasks

msugakov:
https://issues.redhat.com/browse/ROX-27480 is to follow up about moving fetch script.


image/scanner/rhel/konflux.Dockerfile
```diff
@@ -16,19 +16,19 @@ ENV CI=1
 COPY . /src
 WORKDIR /src
 
-RUN scripts/konflux/fail-build-if-git-is-dirty.sh
+RUN .konflux/scripts/fail-build-if-git-is-dirty.sh
```
tommartensen, line 19: We have this script in three different repositories. Can we run it as a separate task before `build-container`?

tommartensen (approved):
LGTM with one thought for the future.

image/scanner/rhel/konflux.Dockerfile
msugakov: Good idea to remove it from repos. Eventually this script only needed in `determine-image-tag` task.

I created https://issues.redhat.com/browse/ROX-27488

PR #1749, opened 2024-12-19
Bump golang.org/x/net to v0.33.0 to fix CVE-2024-45338

Stringy:


RTann (approved):
LGTM, though note running StackRox Scanners are *not* affected by this, as the codepath which uses this is *not* run in live scanners. It is solely for CI purposes with predictable HTML contents

RTann:
CI failure is unrelated

PR #1773, opened 2025-01-17
ROX-27840: prefetch RPM dependencies for Konflux

tommartensen:
How to reproduce the lockfile: https://spaces.redhat.com/display/StackRox/How+to+prefetch+RPMs+for+ACS+Konflux+builds

Required changes:
* cannot blanket `dnf upgrade`, because we need to specify which RPMs to prefetch. Target RPMs can be added to the `packages` section for prefetching, see https://github.com/stackrox/scanner/pull/1793/files for an example (added to the How-to). 
* The lockfile tool cannot resolve `${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG}`, thus changing that to the full image name. We haven't used build arg overrides anyway.
* The repo definitions file ends with 2 empty lines, because we edit it with Python's configparser which intentionally adds them: https://bugs.python.org/issue32917.


.konflux/scripts/rpm-prefetching/rpms.in.yaml
```diff
@@ -0,0 +1,10 @@
+packages: [xz]
+contentOrigin:
+  repofiles: ["./ubi.repo"]
+context:
+    image: registry.access.redhat.com/ubi8-minimal:latest
```
tommartensen, line 5: Change to link to containerfile.

msugakov:
I suggest closing this PR as it was incorporated in https://github.com/stackrox/scanner/pull/1865.

PR #1776, opened 2025-01-21
chore: go1.23

RTann:
See https://github.com/stackrox/rox-ci-image/pull/220.

While here, I also updated `golangci-lint` to a version which supports go1.23 and I made some changes which were not very invasive. I considered copying over more of the rules we have set in https://github.com/stackrox/stackrox/blob/master/.golangci.yml but that'd be a lot to do for a repo we are in the process of deprecating


BradLugo (approved):
All images versions LGTM

PR #1786, opened 2025-01-27
remove aptsources namespace detector

RTann:
Clair does not check for this file these days, so why should this scanner?


dcaravel (approved):
RIP

PR #1794, opened 2025-01-29
Use tasks build with memory requests==limits

msugakov:
This is merely sanity checking for https://github.com/stackrox/konflux-tasks/pull/33

This will not be merged because Renovate will eventually do updates for us.

Checked pods on the cluster to ensure TA containers there look like this:
```yaml
    name: step-create-trusted-artifact
    resources:
      limits:
        memory: 2Gi
      requests:
        cpu: 33m
        memory: 2Gi
```


msugakov:
Served its validation purpose. Closing.

PR #1839, opened 2025-03-18
ROX-27856: Add BigQuery metrics to Konflux pipelines

msugakov:
The new task from https://github.com/stackrox/konflux-tasks/pull/41 is used.

### Testing

See https://github.com/stackrox/konflux-tasks/pull/41#issuecomment-2757538219

Also, EC does not seem to be complaining about these new tasks which is nice.


tommartensen (approved):
LGTM 🚀

msugakov:
`ecosystem-cert-preflight-checks` has problems but our task works. I'm going to merge.

PR #1844, opened 2025-03-20
fix(e2e): fix test + gRPC message size

RTann:
Account for a new vuln + increase gRPC size


BradLugo (approved):
Thanks @RTann! I was trying to backport the gRPC message size yesterday but got too fancy with the git commands and put it down 😅 
Do we need to port this to `release-2.35`?

PR #1851, opened 2025-03-28
ROX-27856: Use modified version of the metrics task

msugakov:
Includes all the changes as of https://github.com/stackrox/konflux-tasks/pull/45 merge.

Verified by looking at bigtable data.

PR #1853, opened 2025-03-31
ROX-27856: Use metrics task that does two inserts

msugakov:
From https://github.com/stackrox/konflux-tasks/pull/45


msugakov:
Will reuse https://github.com/stackrox/scanner/pull/1851 instead.

PR #1855, opened 2025-04-01
Try disable P and Z builds on a present label

msugakov:
testing


msugakov:
The skipped tasks render the pipeline as failed. https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/scanner-db-on-push-pxbzl

PR #1865, opened 2025-04-29
ROX-27638, ROX-20234: Prefetch RPMs, use hermetic builds

msugakov:
Continues from https://github.com/stackrox/scanner/pull/1773.

### Validation

Only CI.


tommartensen (approved):
LGTM 🚀

PR #1868, opened 2025-04-30
ROX-20312: Pin base image digests and extend Renovate config

msugakov:
See descriptions of individual commits.

### Validation

1. Relying on CI signal.
2. Ran renovate validator cli.
3. I don't know how renovate config will actually work w.r.t. image digests and our custom tasks. I suggest merging this PR and monitoring things after.

### DIFF
--- a/image/scanner/rhel/konflux.Dockerfile
+++ b/image/scanner/rhel/konflux.Dockerfile
@@ -1,10 +1,10 @@
 ARG BASE_REGISTRY=registry.access.redhat.com
 ARG BASE_IMAGE=ubi8-minimal
-ARG BASE_TAG=latest
+ARG BASE_TAG=latest@sha256:33161cf5ec11ea13bfe60cad64f56a3aa4d893852e8ec44b2fd2a6b40cc38539
 
 
 # Compiling scanner binaries and staging repo2cpe and genesis manifests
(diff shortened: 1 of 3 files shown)


image/scanner/rhel/konflux.Dockerfile
```diff
@@ -1,10 +1,10 @@
 ARG BASE_REGISTRY=registry.access.redhat.com
 ARG BASE_IMAGE=ubi8-minimal
-ARG BASE_TAG=latest
+ARG BASE_TAG=latest@sha256:33161cf5ec11ea13bfe60cad64f56a3aa4d893852e8ec44b2fd2a6b40cc38539
```
tommartensen, line 3: I am curious to see if this works, ie if Renovate will detect that this is an image tag.

tommartensen (approved):

image/scanner/rhel/konflux.Dockerfile
msugakov: It works: https://github.com/stackrox/scanner/pull/1873

msugakov: Yet, I had to go away from this because `rpm-lockfile-prototype` doesn't resolve the image referenced this way.

https://github.com/stackrox/scanner/pull/1876/files#diff-3e10f07501bbfaabdb0d2e308d4f481a0e3fd46ebf36957aa5430b72d2365da9

PR #1869, opened 2025-04-30
build: Try fix go.mod/go.sum in master

msugakov:
Scanner `master` builds seem to be broken, ref https://github.com/stackrox/scanner/commit/5a83cdcf0761ed25712fbf40b55cc800589b37bc. I also see that in my PR https://github.com/stackrox/scanner/actions/runs/14751966717/job/41411331357?pr=1868#step:5:325

The PR which last touched this was https://github.com/stackrox/scanner/pull/1866 but its own CI finished cleanly.
I don't know what's happening and why but the change suggested here looks reasonably safe to me. If it helps, I suggest merging.

Just ran `go mod tidy`. Also, GoLand suggests the same change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,7 +94,7 @@ require (
 	github.com/facebookincubator/flog v0.0.0-20190930132826-d2511d0ce33c // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
-	github.com/go-jose/go-jose/v4 v4.0.4 // indirect
+	github.com/go-jose/go-jose/v4 v4.0.5 // indirect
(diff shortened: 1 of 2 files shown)


RTann:
The master branch already has this version, so we can close this. Also Scanenr doesn't use go-jose, so it's not really necessary anyway

msugakov:
> The master branch already has this version, so we can close this.

I opened this PR #1869 five hours ago.
go-jose 4.0.5 came in #1862 which you merged less than ten minutes ago.
Saying that master **already** has this version is a bit of stretch.

> Also Scanenr doesn't use go-jose, so it's not really necessary anyway

If `go mod tidy` identifies changes in dependencies, if these changes lead to failing builds like [here](https://github.com/stackrox/scanner/actions/runs/14753631669/job/41416569539?pr=1868), it does not matter what we think any more. It matters what actually happens.

PR #1875, opened 2025-05-06
ROX-29108: Enable `rpm` lockfile renovation

msugakov:
According to <https://konflux-ci.dev/docs/mintmaker/default-config/#rpm-updates>, there's nothing else we should touch.

### Validation

I'll create a follow-up PR where I'd downgrade some package to let MintMaker create its update.

PR #1880, opened 2025-05-07
ROX-29108: Induce RPM lockfile renovation

msugakov:
This follows up on the promise of testing I made in https://github.com/stackrox/scanner/pull/1875

### Validation

Will check that Konflux CI feels happy.
This needs to be merged before we can see if it has an effect on MintMaker.


tommartensen (approved):
Let's test this. Giving green light despite pending Konflux pipelines.

PR #1820, opened 2025-02-28
fix(ci): increase timeout processing RH OVAL data

dcaravel:
To address increasing CI failures similar to:

```
Error: fetching RHELv2 vulns: rhelv2: updating feeds errors: [
rhelv2: unable to decode OVAL document at https://security.access.redhat.com/data/oval/v2/RHEL7/rhel-7-including-unpatched.oval.xml.bz2: read tcp 172.18.0.2:37174->23.55.176.171:443: use of closed network connection (Client.Timeout or context cancellation while reading body), 
rhelv2: unable to decode OVAL document at https://security.access.redhat.com/data/oval/v2/RHEL8/rhel-8-including-unpatched.oval.xml.bz2: read tcp 172.18.0.2:60114->23.39.174.135:443: use of closed network connection (Client.Timeout or context cancellation while reading body)
]
```
Compressed these files are not particularly large:
```
4.0M rhel-7-including-unpatched.oval.xml.bz2
3.7M rhel-8-including-unpatched.oval.xml.bz2
```
Uncompressed:
```
129M rhel-7-including-unpatched.oval.xml
142M rhel-8-including-unpatched.oval.xml
```
The timeout occurs while decompressing + parsing the response body - if these errors continue we can consider loading the compressed file fully into mem (or disk) and then closing the response body to allow decompressing + processing to occur outside of the http.Client timeout.

----

++ Updated testcase that was failing due to new description.

PR #1883, opened 2025-05-08
fix(ci): increase timeout processing RH OVAL data

BradLugo:
## Description

Increases the timeout for the HTTP client that requests Red Hat OVAL feeds. We previously did this for the v2 feed in this PR: https://github.com/stackrox/scanner/pull/1820. However, our "v1" OVAL updater also uses the v2 feed URLs: 
https://github.com/stackrox/scanner/blob/5d2c067ca37b31db830daf54c1a66f4a679db459/ext/vulnsrc/rhel/rhel.go#L44
https://github.com/stackrox/scanner/blob/5d2c067ca37b31db830daf54c1a66f4a679db459/ext/vulnsrc/rhel/rhel.go#L54-L58

Moreover, these changes will keep the timeout consistent between reading from the "v1" and v2 Red Hat OVAL feeds.

Example of the error these changes should fix: https://github.com/stackrox/scanner/actions/runs/14848990112/job/41689019123

### DIFF
--- a/ext/vulnsrc/rhel/rhel.go
+++ b/ext/vulnsrc/rhel/rhel.go
@@ -115,7 +115,7 @@ const (
 
 var (
 	client = &http.Client{
-		Timeout:   10 * time.Second,
+		Timeout:   60 * time.Second,
 		Transport: proxy.RoundTripper(),
 	}
 )


dcaravel (approved):
Nice catch, LGTM

PR #1897, opened 2025-05-14
ROX-27716: Switch to a task version that does not need `SOURCE_BRANCH` param

msugakov:
From https://github.com/stackrox/konflux-tasks/pull/55


msugakov:
Served its testing purpose and can be closed.

PR #1901, opened 2025-05-20
fix(ci): assign the http status variable

BradLugo:
## Description

If the http status code returned by the curl call is a status code we don't expect (i.e., the default case), we log that code. However, we don't assign the variable before executing the switch-case. Here's the error you might see:
```
./scripts/ci/jobs/diff-dumps.sh: line 165: status: unbound variable
```

Example: https://github.com/stackrox/scanner/actions/runs/15137116869/job/42551992695


dcaravel (approved):
LGTM

daynewlee (approved):

PR #1960, opened 2025-06-25
chore(deps): replace yaml lib

janisz:
`gopkg.in/yaml` is not maintained let's follow k8s recommendation and use `go.yaml.in/yaml/v3` alternatively we could use https://github.com/goccy/go-yaml but we already have indirect dependency to k8s version and it's a fork of `github.com/ghodss/yaml` so we can replace 2 deps with 1 without any code changes.

See:
- https://github.com/kubernetes/kubernetes/blob/8eda367f27b79adf5b41f019ede7708abee9602d/hack/unwanted-dependencies.json#L89C35-L90
- https://github.com/go-yaml/yaml/tree/v3?tab=readme-ov-file#this-project-is-unmaintained

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,6 +54,7 @@ require (
 	google.golang.org/grpc v1.73.0
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.5.1
 	google.golang.org/protobuf v1.36.6
+	sigs.k8s.io/yaml v1.4.0
 )
 
 require (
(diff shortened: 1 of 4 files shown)


RTann (approved):
seems fine, though I'd still recommend to do what the [README](https://github.com/kubernetes-sigs/yaml/tree/master/goyaml.v3#readme) says:

```
// Old way
import "sigs.k8s.io/yaml/goyaml.v3"

// Recommended way
import "go.yaml.in/yaml/v3"
```

PR #1963, opened 2025-07-03
go1.24

RTann:
See https://github.com/stackrox/rox-ci-image/compare/0.4.8...0.4.9 for the updated Go version


pkg/analyzer/java/java.go
```diff
@@ -36,7 +36,7 @@ var nameRegexp = regexp.MustCompile(`([[:graph:]]+)-([[:digit:]][\-.[:alnum:]]*(
 // checkName returns the extracted package name from the above regexp.
 func checkName(name string) string {
 	m := nameRegexp.FindStringSubmatch(name)
-	if m == nil || len(m) < 2 {
+	if len(m) < 2 {
```
RTann, line 39: This was called out after the linter update

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Compiling scanner binaries and staging repo2cpe and genesis manifests
-FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.23@sha256:0a070e4a8f2698b6aba3630a49eb995ff1b0a182d0c5fa264888acf9d535f384 AS builder
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_1.24@sha256:021ab8e6e4af3d1683b52ec985996f9421edb2111dc697b5781dee3c7595ae00 AS builder
```
msugakov, line 2: I added `konflux-build` label because the builds don't trigger in PRs without it (we can revisit).
Please push any/empty commit to this PR to trigger Konflux CI.
Let's make sure `openshift-golang-builder` is also happy with the change.

RTann: Thanks! I forgot the Konflux builds didn't run by default

RTann: It looks like they changed the name of the tag, so I had to update that, too. I chose the option I figured would be best

janisz (approved):

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/stackrox/scanner
 
-go 1.23.3
+go 1.24
 
-toolchain go1.23.6
+toolchain go1.24.3
```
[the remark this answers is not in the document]
msugakov: Simply, `openshift-golang-builder` has no ability to download any toolchains because that feature is disabled. The only one it can use is what's provided. If the image comes with 1.24.3, the highest toolchain we can have (when building with `openshift-golang-builder`) is 1.24.3.

tommartensen (changes requested):
A production-ready Go 1.24 image is available, please update the reference before merging.

RTann:
The E2E test failures are ok for this PR

PR #1967, opened 2025-07-10
ROX-30075: Test git tag sanitization

msugakov:
For https://github.com/stackrox/konflux-tasks/pull/69


msugakov:
Served its purpose closing because we'll get the task update via MintMaker.

PR #1999, opened 2025-07-24
e2e: fix tests with latest vuln data

RTann:


e2etests/vuln_test.go
```diff
@@ -142,9 +142,11 @@ func TestStackroxVulnImages(t *testing.T) {
 			image: "quay.io/rhacs-eng/qa:rails-cve-2016-2098",
 			expectedFeatures: []feature{
 				{"rails", "4.2.5.1", []expectedVuln{
-					{name: "CVE-2016-2098"},
-					{name: "CVE-2016-6316"},
-					{name: "CVE-2016-6317"},
+					// These three vulns should exist, but NVD set these to deferred.
+					// Commenting out until they are no longer deferred.
```
BradLugo, line 146: How will we know when they're no longer deferred?

jvdm (approved):

e2etests/vuln_test.go
RTann: https://youtu.be/se17_0zbZds?si=SHZtYEjhrSZ8xkPU&t=10

I considered just deleting these (open to that), but I also wanted to keep track of them in case we want to remember them in the future

BradLugo: 😂 

My initial thought was to file a Jira issue, but after thinking about it a bit, would these tests fail when the CVEs are no longer deferred? Since the expected results would be different

PR #2015, opened 2025-07-31
ROX-28353: use NVD 2.0 JSON feeds

RTann:
NVD will remove the 1.1 JSON feeds August 20, 2025, so this PR switches our feed loader to use the new 2.0 feeds. The schema is the same as the API, so this was pretty straightforward.

The NVD 2.0 API and feeds includes vulnerabilities with more NVD statuses than the 1.1 JSON feeds did (https://nvd.nist.gov/vuln/vulnerability-status). For example, "Deferred" vulnerabilities are included. As long as the vulnerability is not rejected and has all the fields we need, we do not care if NVD opted to change its status.

We should backport this to currently supported releases.


pkg/vulnloader/nvdloader/convert.go
```diff
@@ -152,14 +152,18 @@ func toBaseMetricV2(metrics []*apischema.CVEAPIJSON20CVSSV2) *jsonschema.NVDCVEF
...
+	if baseMetric != nil {
+		return baseMetric
+	}
+
+	baseMetric = toBaseMetricV30(metrics30)
+	if baseMetric != nil {
+		return baseMetric
 	}
+
+	return nil
```
dcaravel, line 166: Didn't like the len checks?

e2etests/vuln_test.go
```diff
@@ -142,23 +142,16 @@ func TestStackroxVulnImages(t *testing.T) {
 			image: "quay.io/rhacs-eng/qa:rails-cve-2016-2098",
 			expectedFeatures: []feature{
 				{"rails", "4.2.5.1", []expectedVuln{
+					{name: "CVE-2016-2098"},
+					{name: "CVE-2016-6316"},
+					{name: "CVE-2016-6317"},
```
dcaravel, line 147: 🤔  - these are still marked as deferred in NVD, were these moved because the 2.0 feed includes them again?

pkg/vulnloader/nvdloader/convert.go
RTann: didn't seem very necessary, and would have caused for more if/else, which is nice to avoid for readability :)

pkg/vulnloader/nvdloader/convert.go
```diff
@@ -31,6 +31,11 @@ func toJSON(vulns []*apischema.CVEAPIJSON20DefCVEItem) ([]*jsonschema.NVDCVEFeed
 			continue
 		}
 
+		// Ignore rejected vulnerabilities.
+		if strings.EqualFold(vuln.CVE.VulnStatus, "Rejected") {
```
RTann, line 35: @dcaravel @BradLugo I added this conditional after noticing the 2.0 feeds include rejected vulns. They will lack a lot of fields, but just to be safe, I opted to explicitly ignore it

BradLugo: I vaguely recall there being some contention with us supporting rejected vulns in the past, and ignoring them as the solution, so your change tracks 👍

RTann: I didn't add any because they didn't exist before, but that's a bad reason to not add tests. Just added one which covers this rejected case and some basic stuff

BradLugo (approved):
Might be helpful to have a brief explanation of the difference between the feed and API, but I'm not sure where that should live. Maybe the feature flag comment? Either way, not gonna block on that. LGTM. Thanks for taking care of this, @RTann.

PR #2092, opened 2025-08-26
fix(ci): add specific psql version to fix s390x

BradLugo:


daynewlee:
I am expert of this, do we have to worry about the Konflux build failures?

daynewlee (commented):
Left a minor question

dcaravel:
> I am not an expert of this, do we have to worry about the Konflux build failures?

Good Q - my 2 cents - since upstream images and CI do not use anything from Konflux - this at least fixes upstream.

Looks like both downstream and konflux use `registry.redhat.io/rhel8/postgresql-15` as a base, which appears to not depend on the `download.sh` script for downloading postgres.

PR #2189, opened 2025-09-15
Try drop `init` task

msugakov:


msugakov:
Tried. Got info - Conforma wants `init` task. Closing.

PR #2195, opened 2025-09-15
ROX-30898: Assign index image tag differently

msugakov:
See the related ticket.

### Validation

See my comments below.


msugakov:
Now, the interesting thing is how the `init` task would behave if the image already exists. It shouldn't even call to `skopeo` but let's make it run to be sure.
https://github.com/konflux-ci/build-definitions/blob/8ddc2b52768ed7b8c0794b59ad5b3d944354b15e/task/init/0.2/init.yaml#L52

PR #2294, opened 2025-10-07
Try set limits lower than requests

msugakov:
Testing for https://github.com/stackrox/scanner/pull/2293


msugakov:
Testing done. Closing.

PR #2562, opened 2026-01-15
chore(base-image): Migrate builds to UBI9/RHEL9

davdhacs:
## Description

Migrates scanner builds from UBI8/RHEL8 to UBI9/RHEL9 base images.

### Key Changes

**Konflux Base Images:**
- Builder: `brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_1.24`
- Scanner runtime: `registry.access.redhat.com/ubi9-minimal:latest`
- Scanner DB: `registry.redhat.io/rhel9/postgresql-15:latest`
- Image names: `rhacs-scanner-rhel9`, `rhacs-scanner-slim-rhel9`, `rhacs-scanner-db-rhel9`, `rhacs-scanner-db-slim-rhel9`

**Non-Konflux Base Images:**
- Scanner: `ubi9-minimal`
- Scanner DB: `ubi9` / `ubi9-minimal`
- Vulnerabilities: `ubi9-minimal`

**RPM/Repo Updates:**
- `rpms.lock.yaml`: xz updated from RHEL 8 (5.2.4) to RHEL 9 (5.2.5)
- `rpms.rhel.repo`: repos updated from `rhel8` to `rhel9`
- PostgreSQL download script: `pg_rhel_major=9`

**Tekton Labels:**
- CPE labels: `el8` → `el9`

**UBI9 Compatibility Fixes:**
- `update-ca-trust extract -o /etc/pki/ca-trust/extracted` for unprivileged containers ([RHBZ#2241240](https://bugzilla.redhat.com/show_bug.cgi?id=2241240))
- `cp --recursive --no-dereference --no-clobber` in restore-all-dir-contents
- `microdnf install -y xz` (explicit `-y` flag)

## Checklist
- [x] Investigated and inspected CI test results

## Testing Performed

TBD

PR #3095, opened 2026-04-14
ROX-34148: Migrate Dockerfiles to ubi9-micro base images

janisz:
Refs:
- https://github.com/stackrox/stackrox/pull/19379
- https://github.com/stackrox/stackrox/pull/19378
- https://github.com/stackrox/stackrox/pull/17406
- https://github.com/stackrox/stackrox/pull/17430

Tested:
- https://github.com/stackrox/stackrox/pull/20071


image/scanner/scripts/import-additional-cas
```diff
@@ -26,9 +26,13 @@ copy_existing /etc/pki/injected-ca-trust
...
-echo "Copying StackRox root CA from '${CA_PATH}'"
-# For RHEL
-cp "${CA_PATH}" /etc/pki/ca-trust/source/anchors/root-ca.pem
+if [ -f "$CA_PATH" ]; then
+    echo "Copying StackRox root CA from '${CA_PATH}'"
+    # For RHEL
+    cp "${CA_PATH}" /etc/pki/ca-trust/source/anchors/root-ca.pem
+else
+    echo "No StackRox root CA found at '${CA_PATH}' - skipping"
+fi
```
dcaravel: What drove this change? Could this skip add difficulty troubleshooting? (will now be a log and comms broken vs. an obvious crash)

janisz: Ah, my bad. I tried to switch db images as well and this must be a left over

image/scanner/rhel/Dockerfile
```diff
@@ -25,20 +46,15 @@ LABEL name="scanner" \
...
 COPY --from=extracted_bundle /bundle/scanner ./
 
 COPY --from=extracted_bundle /bundle/THIRD_PARTY_NOTICES/ /THIRD_PARTY_NOTICES/
 
-RUN microdnf upgrade -y --nobest && \
-    microdnf install -y xz && \
-    microdnf clean -y all && \
-    # (Optional) Remove line below to keep package management utilities
-    # We don't uninstall rpm because scanner uses it to get packages installed in scanned images.
-    rpm -e --nodeps $(rpm -qa curl '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
```
dcaravel, line 39: Missed this before: curl was explicitly removed in the past but is included in the new micro image.  If not intentional should be removed in new image as well.

janisz: ubi-micro does not include curl also it doesn't include other packages (including package manager) and we only install following packages 

- findutils 
 - ca-certificates 
 - rpm 
  - xz 
 - gzip 
 - less 
 - tar

dcaravel: It shows in the image built as part of this PR:

```sh
$ docker run -it --rm --entrypoint /bin/bash quay.io/rhacs-eng/scanner:2.39.x-152-g8a7cd6e606
bash-5.1$ curl -h
Usage: curl [options...] <url>
 -d, --data <data>   HTTP POST data
 -f, --fail          Fail silently (no output at all) on HTTP errors
 -h, --help <category>  Get help for commands
 -i, --include       Include protocol response headers in the output
 -o, --output <file>  Write to file instead of stdout
 -O, --remote-name   Write output to a file named as the remote file
 -s, --silent        Silent mode
 -T, --upload-file <file>  Transfer local FILE to destination
 -u, --user <user:password>  Server user and password
 -A, --user-agent <name>  Send User-Agent <name> to server
 -v, --verbose       Make the operation more talkative
 -V, --version       Show version number and quit

This is not the full help, this menu is stripped into categories.
Use "--help category" to get an overview of all categories.
For all options use the manual or "--help all".
```

And not in the prior tagged image(s):
```
$ docker run -it --rm --entrypoint /bin/bash quay.io/rhacs-eng/scanner:2.39.8
bash-4.4$ curl -h
bash: curl: command not found
```

Also appears the removal no longer works after this PR: https://github.com/stackrox/scanner/pull/2562 (due to `curl` package name changed to `curl-minimal` breaking removal)

janisz: You are right. I thought ubi-micro does not have curl so I thought we don't need this line but curl was installed with rpm. Now it should be fine.

janisz:
@dcaravel PTAL

rpms.in.yaml
```diff
@@ -2,13 +2,31 @@
...
 contentOrigin:
   repofiles: [ "rpms.rhel.repo" ]
+
 context:
   containerfile:
     file: image/scanner/rhel/konflux.Dockerfile
-    stageName: scanner-common
+    stageName: package_installer
+    imagePattern: registry.access.redhat.com/ubi9/ubi-micro:latest
+
```
msugakov, line 22: It would be better to do
```suggestion
context:
  bare: true
```
instead.

`context.containerfile` means: "stop taking transitive dependencies if you see them already present in the stage I point you to". Since the way we install dependencies isn't quite usual, we may run into bugs and see some dependencies excluded when they should not be.
I suggest we pay extra price for over-fetching of RPMs (we do the same in all other repos) to not run into strange build failures.

rpms.in.yaml
```diff
@@ -2,13 +2,31 @@
 # See our docs here: https://spaces.redhat.com/display/StackRox/How+to+prefetch+RPMs+for+ACS+Konflux+builds
 
 packages:
+# Scanner packages (for konflux.Dockerfile stage: package_installer)
+# These packages are installed on top of ubi9-micro base image
+# Note: coreutils and bash are excluded - ubi9-micro already includes them
```
msugakov: I think, this note about coreutils and bash isn't necessary. Up to you.

```suggestion
```

rpms.in.yaml
```diff
@@ -2,13 +2,31 @@
...
+- rpm
 - xz
+- gzip
+- less
+- tar
+
+# Note: Scanner-DB images don't need RPM prefetch because:
+# - konflux.Dockerfile uses registry.redhat.io/rhel9/postgresql-15 base (already has everything)
+# - Dockerfile and Dockerfile.slim use ubi9 (full) in dependency_builder stage (can install packages via dnf normally)
+
```
msugakov: This feels redundant too, also looks like AI output.

```suggestion
```

image/vulnerabilities/Dockerfile
```diff
@@ -2,7 +2,7 @@ ARG BASE_REGISTRY=registry.access.redhat.com
 ARG BASE_IMAGE=ubi9-minimal
 ARG BASE_TAG=latest
 
-FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG}
+FROM ${BASE_REGISTRY}/ubi9-micro:${BASE_TAG}
```
msugakov: Since you touched these, let's clean up and simplify please.


```suggestion
FROM registry.access.redhat.com/ubi9-micro:latest
```

Unless override args are passed in the calling code?

Could you please get rid of `BASE_REGISTRY`, `BASE_IMAGE` and `BASE_TAG` in other places too?

image/scanner/rhel/konflux.Dockerfile
```diff
@@ -27,8 +27,31 @@ RUN echo -n "version: " && make --quiet --no-print-directory tag && \
...
+COPY --from=ubi-micro-base / /out/
+
+RUN dnf install -y \
+    --installroot=/out/ \
+    --releasever=9 \
+    --setopt=install_weak_deps=0 \
+    --setopt=reposdir=/etc/yum.repos.d \
+    --nodocs \
+    findutils \
+    ca-certificates \
```
msugakov, line 43: I think, we also need `openssl` for FIPS.

image/scanner/rhel/Dockerfile
```diff
@@ -2,13 +2,35 @@ ARG BASE_REGISTRY=registry.access.redhat.com
...
+    --setopt=install_weak_deps=0 \
+    --nodocs \
+    findutils \
+    ca-certificates \
+    rpm \
+    xz \
+    gzip \
+    less \
+    tar && \
+    chroot /out rpm -e --nodeps curl && \
```
msugakov: I see in that thread https://github.com/stackrox/scanner/pull/3095/changes#r3242274896 why it's necessary to uninstall `curl`. That's educational.

1. Could you please add a comment telling that curl gets installed with rpm?
2. I think, it may be possible to use `rpm` without `chroot`. There's `rpm --root=/out`. Would it work?

msugakov (commented):
Looks good overall, but I've a few findings.

image/vulnerabilities/Dockerfile
janisz: Looks like this args are not used, let's remove them everywhere.

rpms.in.yaml
```diff
@@ -2,13 +2,27 @@
...
 contentOrigin:
   repofiles: [ "rpms.rhel.repo" ]
+
 context:
   containerfile:
     file: image/scanner/rhel/konflux.Dockerfile
-    stageName: scanner-common
+    stageName: package_installer
+    imagePattern: registry.access.redhat.com/ubi9/ubi-micro:latest
```
msugakov: This `context` should be deleted now. Continuing https://github.com/stackrox/scanner/pull/3095#discussion_r3324706970

image/vulnerabilities/Dockerfile
msugakov: Unresolving so not to forget remove BASE_REGISTRY, BASE_IMAGE and BASE_TAG.

image/scanner/rhel/konflux.Dockerfile
janisz: added in line 49

image/scanner/rhel/Dockerfile.slim
```diff
@@ -25,8 +21,10 @@ RUN dnf install -y \
     xz \
     gzip \
     less \
-    tar && \
-    chroot /out rpm -e --nodeps curl && \
+    tar \
+    openssl && \
```
msugakov: Here, in `image/scanner/rhel/Dockerfile.slim`, `openssl` isn't necessary because we don't distribute these images for FIPS. That'll also be consistent with how you made other upstream images, e.g. in stackrox/stackrox repo.

image/scanner/rhel/Dockerfile
```diff
@@ -25,8 +21,10 @@ RUN dnf install -y \
     xz \
     gzip \
     less \
-    tar && \
-    chroot /out rpm -e --nodeps curl && \
+    tar \
+    openssl && \
```
msugakov: Same note about `openssl` not needed in the upstream image.

.tekton/scanner-build.yaml
msugakov, line 1: Question (not about this line). 4.11 branches were already cut. Other containers are already migrated in 4.11 to ubi-micro except for Scanner V2 ones. What's your plan, do you want to backport this PR to `release-2.40` branch after merging to `master` or would you leave 4.11 as-is (on ubi9-minimal)?

msugakov (commented):
Looks good and thanks for taking care of `BASE_REGISTRY`, `BASE_IMAGE` and `BASE_TAG`.

dcaravel (approved):
LGTM, please address @msugakov's comments. Thanks.

.tekton/scanner-build.yaml
janisz: I thought about leaving it as it unless the scanner team plan to backport it

msugakov: Ok. By leaving as is you mean you'd still merge it to `master`, right?

image/scanner/rhel/Dockerfile
```diff
@@ -1,14 +1,34 @@
...
+    --setopt=install_weak_deps=0 \
+    --nodocs \
+    findutils \
+    ca-certificates \
+    rpm \
+    xz \
+    gzip \
+    less \
+    tar \
+    openssl && \
```
janisz:
```suggestion
    tar && \
```

image/scanner/rhel/Dockerfile.slim
```diff
@@ -1,14 +1,34 @@
...
+    --setopt=install_weak_deps=0 \
+    --nodocs \
+    findutils \
+    ca-certificates \
+    rpm \
+    xz \
+    gzip \
+    less \
+    tar \
+    openssl && \
```
janisz:
```suggestion
    tar && \
```

.tekton/scanner-build.yaml
janisz: Correct, merge and release with next big release of stackrox

image/scanner/rhel/Dockerfile.slim
janisz: Fixed

image/scanner/rhel/Dockerfile
janisz: Fixed

.tekton/scanner-build.yaml
```diff
@@ -58,6 +58,8 @@ spec:
     value:
     # X.Y in the cpe label must be adjusted for every version stream.
     - "cpe=cpe:/a:redhat:advanced_cluster_security:X.Y::el9"
+  - name: ACTIVATION_KEY
+    value: subscription-manager-activation-key-prod
```
msugakov: Sorry for the late comment, but `ACTIVATION_KEY` is not even needed here.

.tekton/scanner-slim-build.yaml
```diff
@@ -58,6 +58,8 @@ spec:
     value:
     # X.Y in the cpe label must be adjusted for every version stream.
     - "cpe=cpe:/a:redhat:advanced_cluster_security:X.Y::el9"
+  - name: ACTIVATION_KEY
+    value: subscription-manager-activation-key-prod
```
msugakov: Same about this one.

.tekton/scanner-build.yaml
janisz:
```suggestion
```

.tekton/scanner-slim-build.yaml
janisz:
```suggestion
```

msugakov (approved):

PR #3096, opened 2026-04-14
ROX-34058: Scanner V2 fix labels.json breakage

dcaravel:
## Description
Fixes a bug that causes Scanner to detect zero components in RHEL images and wrong OS semi-randomly.

## Conditions

Scanning an image that contain both `content_sets.json` and `labels.json` - example: `quay.io/rhacs-eng/qa:ubi9-9.7-1769417801-amd64`

## Root Cause
Scanner stored `*buildinfo*` files in a map and looped over them, when `labels.json` happened to be the first file selected no content sets are found, which cascades and breaks detection (map iterations in `Go` are non-deterministic)

### DIFF
--- a/pkg/rhelv2/rpm/rpm.go
+++ b/pkg/rhelv2/rpm/rpm.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


dcaravel:
Closing for now as scanner is being phased out.

PR #3344, opened 2026-06-30
fix(feed): Improve NVD feed download resilience (x2)

dcaravel:
Expands on https://github.com/stackrox/scanner/pull/3343 to increase the retires / timeouts - based on observed status in https://github.com/stackrox/scanner/actions/runs/28467785361/job/84372363491 riding the threshold (update: and failing)

### DIFF
--- a/pkg/vulnloader/nvdloader/loader_api.go
+++ b/pkg/vulnloader/nvdloader/loader_api.go
@@ -22,7 +22,7 @@ import (
 const urlFmt = `https://services.nvd.nist.gov/rest/json/cves/2.0?noRejected&startIndex=%d`
 
 var client = http.Client{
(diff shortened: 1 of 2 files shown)


jvdm (commented):
Concerns with Konflux timeout, do we need to look at that @BradLugo?

BradLugo:
> Concerns with Konflux timeout, do we need to look at that @BradLugo?

I think doing retries as needed will be sufficient. If yall are curious about what I was referring to, here's an example from when I bumped the konflux timeout last time:

- https://github.com/stackrox/scanner/pull/3015/changes
- https://github.com/stackrox/konflux-tasks/pull/98

BradLugo (approved):

PR #3357, opened 2026-07-01
chore(feed): cap exponential backoff

dcaravel:
The backoff introduced is a bit 'too exponential' when retries were increased, this change caps backoff to 5 mins

[ex](https://github.com/stackrox/scanner/actions/runs/28510678753/job/84510586373):
```
... retrying in 42m40s
```

```
time="2026-07-01T13:32:06Z" level=warning msg="Feed year 2024: attempt 9 failed: reading feed body (read 93318053 bytes, elapsed: 6m26.142596235s): stream error: stream ID 1; INTERNAL_ERROR; received from peer; retrying in 42m40s"
```

### DIFF
--- a/pkg/vulnloader/nvdloader/loader_feed.go
+++ b/pkg/vulnloader/nvdloader/loader_feed.go
@@ -51,6 +51,7 @@ func (l *feedLoader) downloadFeedForYear(enrichments map[string]*FileFormatWrapp
 	url := fmt.Sprintf("https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-%d.json.gz", year)
 
 	const maxRetries = 10
+	const maxBackoff = 5 * time.Minute
 	backoff := 10 * time.Second
 	var apiFeed *apischema.CVEAPIJSON20
 	for attempt := 1; ; attempt++ {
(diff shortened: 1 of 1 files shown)


kovayur (approved):

pkg/vulnloader/nvdloader/loader_feed.go
```diff
@@ -51,6 +51,7 @@ func (l *feedLoader) downloadFeedForYear(enrichments map[string]*FileFormatWrapp
 	url := fmt.Sprintf("https://nvd.nist.gov/feeds/json/cve/2.0/nvdcve-2.0-%d.json.gz", year)
 
 	const maxRetries = 10
+	const maxBackoff = 5 * time.Minute
```
jvdm: Given how sensitive NVD's availability is to request rates, and since they've explicitly acknowledged performance issues in the past, do you think it would make sense to back off for longer periods of time?

```go
backoffs := []time.Duration{
    15 * time.Second,
    30 * time.Second,
    1 * time.Minute,
    2 * time.Minute,
    4 * time.Minute,
    5 * time.Minute, // max backoff
}
```

Then:

```go
backoff := backoffs[min(attempt-1, len(backoffs)-1)]
```

For a short, fixed retry policy like this, I think the explicit table is easier to reason about than encoding the same behavior indirectly through arithmetic and a cap.

jvdm (approved):

PR #3625, opened 2026-08-14
chore(support): Do not send failure notification for PRs

dcaravel:
Updates the `CI` workflow to only send notifications to slack when a failure occurs from a scheduled or tagged CI run.

Failures from PRs or release branch pushes will no longer generate alerts.

The intent is to reduce noise and ensure alerts are meaningful.


vjwilson (approved):
/lgtm

PR #3684, opened 2026-08-24
fix(konflux, 2.39): increase timeout of scanner(-db) pipeline by 2 hours

tommartensen:
Generally, the build of the blobs roughly takes 2-3.5 hours these days. We increase the timeouts of the Konflux pipelines accordingly.

We don't need to update the *-slim pipelines because they don't wait for any blobs.

Context: https://redhat-internal.slack.com/archives/C03D78P4RFF/p1787017109044639


vjwilson (approved):
/lgtm

PR #3816, opened 2026-09-11
ROX-36649: remove empty commit from patch release

alkmim:
Remove the empty commit from patch release script. See https://redhat.atlassian.net/browse/ROX-36649 for context


scripts/patch-release.sh
```diff
@@ -335,26 +334,14 @@ main() {
...
-
-    if [[ "$head_commit_msg" =~ ^Release\ [0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-        log_warn "Tip of $release_branch is already a release commit: $head_commit_msg"
-        if ! confirm "Create another release commit on top?"; then
-            log_info "Aborted by user."
-            exit 0
-        fi
-        echo
-    fi
-
```
jvdm, line 357: Why remove this check?

scripts/patch-release.sh
```diff
@@ -443,13 +430,14 @@ main() {
         echo
     fi
 
-    # Step 8: Create release commit and tag.
-    log_info "Creating release commit and tag..."
+    # Step 8: Create release tag on the current HEAD of the release branch.
+    # Note: this intentionally does NOT create a new commit. Tagging HEAD directly
+    # avoids pushing a commit to the protected release branch, so no "bypass branch
+    # protection" permission is required (see ROX-36649).
```
jvdm: You documenting planning and historical steps, not augmenting the code to facilitate readability.

alkmim: you are right. I reverted to the previous comment style.

scripts/patch-release.sh
alkmim: We already check if the tag exists here: https://github.com/stackrox/scanner/blob/5de3e6470d246d1f9126418e3e18398aa88ce018/scripts/patch-release.sh#L310

Before, we were adding an empty commit on top of the release branch. If the tip of the  release branch was already a release commit, this check was confirming if we really wanted to add another commit. Since we no longer create the empty commit in Step 8, this check is not guarding against any action.

jvdm (approved):

